PR #22, opened 2021-08-06
NOTICK: Update state machine messages that were missed on first commit

JamesHR3:
A PR updating some flow related Avro messages got merged into the runtime-os repo just before the PR moving these messages to the API repo was merged, resulting in these getting out of sync.


mnesbit (approved):
OK

PR #135, opened 2021-11-09
NOTICK CPI/CPK signature validation POC

woggioni:
POC to validate certificate chain on CPI/CPK signatures


chrisr3 (changes requested):
The `CertPath` encodings look wrong to me.

PR #159, opened 2021-11-18
Woggioni/signature poc

woggioni:
doc [here](https://github.com/corda/corda-api/blob/woggioni/signature-poc/packaging/signature-demo/README.md)


chrisr3 (commented):
A few initial thoughts.

PR #325, opened 2022-04-13
CORE-5801 Add membership schemas for validation

charlieR3:
This PR introduces a schema for validating the a group policy is well formed. It needs to support a dynamic group, and static group and work as a group policy for the MGM CPI.

Services to expose this schema and validate against group policy samples will be created later. 

This file has been put together based on agreement in: https://github.com/corda/platform-eng-design/blob/master/core/corda-5/corda-5.0/the-host/group-policy.md
_Note: The samples in the above doc don't all compile after applying agreed updates to schema so this PR resolves those minor issues: https://github.com/corda/platform-eng-design/pull/482_


data/group-schema/src/main/resources/net/corda/schema/group/policy.1.0/corda.group.policy.json
```diff
@@ -0,0 +1,272 @@
...
+          "required": [
+            "members"
+          ],
+          "properties": {
+            "members": {
+              "description": "The member templates for the members to be added to the static network member list.",
+              "type": "array",
+              "items": {
+                "description": "A member template as a map of values for a member to be added to the static network member list.",
+                "type": "object"
```
dimosr: Should we be imposing a specific structure for each member item here?

dimosr (commented):
I had a quick look and it looks quite reasonable to me for a start (I didn't try to think through all the regexp edge cases though!) - just one question. When we later get to validation, we can spend some more time, add tests etc. but I think it shouldn't block us from progressing with other work.

data/group-schema/src/main/resources/net/corda/schema/group/policy.1.0/corda.group.policy.json
charlieR3: I did have that originally, but since we changed to the key-value approach, similar to the member info `LayeredPropertyMap`, I felt it was better to keep this open for any string-string pairs since we will basically just output all of the MGMs member info here which we may add to. Also it would be hard to validate for specific keys, specifically for arrays of data where the keys will contain indexes.

charlieR3:
> When we later get to validation, we can spend some more time, add tests etc. but I think it shouldn't block us from progressing with other work.

Absolutely. We'll include plenty of tests around this when the services consuming this are added.

data/group-schema/src/main/resources/net/corda/schema/group/policy.1.0/corda.group.policy.json
dimosr: ah ok, that makes sense - I wasn't considering the free-form key-value approach will be used here too.

charlieR3: Ah, sorry. My comment was in regards to MGM Info. I misread where in the schema this comment was. But the answer is essentially the same. We may have indexed values. We should aim to have good documentation supporting this.

driessamyn:
FYI @josephzunigadaly 

This should probably also be integrated with the CPI build tool. Not sure if that's on your backlog, @charlieR3 , or something we should look at, but worth liasing with @josephzunigadaly . Probably not a must for DP2.

charlieR3:
> FYI @josephzunigadaly
> 
> This should probably also be integrated with the CPI build tool. Not sure if that's on your backlog, @charlieR3 , or something we should look at, but worth liasing with @josephzunigadaly . Probably not a must for DP2.

There's nothing currently on our backlog for this, no. Agreed it would be useful to validate this when building the CPI also.

data/membership-schema/src/main/kotlin/net/corda/schema/membership/provider/MembershipSchemaException.kt
```diff
@@ -0,0 +1,6 @@
+package net.corda.schema.membership.provider
+
+/**
+ * Exception thrown when requested config schema is not available.
+ */
+class MembershipSchemaException(msg: String) : RuntimeException(msg)
```
nikinagy: Why not use CordaRuntimeException?

charlieR3: No good reason! Will update it.

mnesbit:
It would be good to get this updated and make the schemas available to the CPI validation work

charlieR3:
> It would be good to get this updated and make the schemas available to the CPI validation work

Will ask a team member to get this over the line in my absence.

nikinagy (dismissed):
LGTM

nikinagy (approved):
LGTM

dimosr (approved):

ronanbrowne (approved):
LGTM

PR #365, opened 2022-05-13
CORE-4596: Add topic definitions to Schema

benmcmahon100:
Add code level definitions for how we expect topics to look by default.


data/topic-schema/src/main/resources/net/corda/schema/DBSchema.yaml
```diff
@@ -0,0 +1,8 @@
+topics:
+  DBSchameTopic:
```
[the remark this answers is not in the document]
benmcmahon100: The topic level files were meant to match the first dot of the prefix, I can make them into one big one, but I thought it made it slightly easier to find - i.e: `config.X` lived in config schema.

The "Schema" part itself refers to the schemas for the topics, but I can see where the confusion might arise

data/topic-schema/src/main/resources/net/corda/schema/CPISchema.yaml
```diff
@@ -0,0 +1,17 @@
+topics:
+  CPIInfoTopic:
```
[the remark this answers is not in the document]
benmcmahon100: It's just to create a standard shape for a topic definition file. Matthew thought there might be a need for other fields, but if needs be, I can remove them later on.

data/topic-schema/src/main/resources/net/corda/schema/CPISchema.yaml
```diff
@@ -0,0 +1,17 @@
+topics:
+  CPIInfoTopic:
+    name: "cpi.info"
+    consumers:
+    producers:
+    config:
+      cleanup.policy: "compact"
```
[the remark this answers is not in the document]
benmcmahon100: These are the literal name of the settings. I don't think they're interchangeable

data/topic-schema/src/main/resources/net/corda/schema/DBSchema.yaml
```diff
@@ -0,0 +1,8 @@
+topics:
+  DBSchameTopic:
+    name: "db.entity.processor"
+    consumers:
+      - test1
+    producers:
+      - test2
```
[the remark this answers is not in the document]
benmcmahon100: Was just meant as a sample to how it might look. I've dropped them now

data/topic-schema/src/main/resources/net/corda/schema/DBSchema.yaml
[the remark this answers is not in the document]
driessamyn: I think in many ways having a single file with the topics listed alphabetical would be clearer and point out some of the inconsistencies in the topic names/"namespacing".
I don't think the current structure makes much sense and that might be a reflection of some of these inconsistencies already.

yarivus (dismissed):
LGTM - I thinks the separation to categories w/ separate yaml files is a good idea ;)

data/topic-schema/src/main/resources/net/corda/schema/CPISchema.yaml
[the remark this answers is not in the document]
davidcurrie: Given that the keys in this map aren't pre-defined, I don't think you could add anything else to the map that wasn't another topic without it becoming really confusing. I guess the question I'd ask is: are these keys actually used anywhere? If not, are they adding any value to the reader? As @driessamyn suggests elsewhere, I'd rather that they weren't a substitute for using meaningful names for the actual topics.

data/topic-schema/src/main/resources/net/corda/schema/DBSchema.yaml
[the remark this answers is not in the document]
davidcurrie: I'd be marginally in favor of a single file, particularly given the additional complexity that loading files from a package has, but I wouldn't get hung up on it.

data/topic-schema/src/main/kotlin/net/corda/schema/Schemas.kt
```diff
@@ -34,24 +34,78 @@ class Schemas {
         fun getRPCResponseTopic(topic: String) = "$topic.resp"
     }
 
+    /**
+     * Config read topic schema
+     */
+    class Config {
+        companion object {
+            const val CONFIG_TOPIC = "config.topic"
```
davidcurrie, line 42: Is there anything that ensures that topic names listed here match those defined in the YAML? Better still, could these be generated from the YAML?

driessamyn:
> LGTM - I thinks the separation to categories w/ separate yaml files is a good idea ;)

@yarivus Why?
And more importantly, please explain to me what the separation logic is.
Logically, it could be "horizontals" (i.e. processors) as by definition, the topics are the communication means between Processors. However, which direction is it in? Are we saying topics belong to the Processor that writes or reads? Logically, you'd have to say write as most commonly it's one processor writes, one or more reads. However, there are some exceptions (e.g. Flow Event). Which processor/group do they belong to in that case?
And probably worse, how do you handle topics for the RPC pattern? Does it belong to the request or response processor? And does that means you split up the request and response topic into 2 files, which would be madness, or break the rule?

So horizontals clearly don't work. So we can look at verticals. Maybe we can group things by "function". In this case, what are they? Is Vnode a function? Is packaging a function? Or are they the same group? What about foundational things like config? Are they a function by themselves? In that case, we'll typically have files with 1 or 2 topic definitions. And what for?

The only split I can see that might make sense is P2P vs. Core Corda as P2P could in theory be deployed independently.

data/topic-schema/src/main/resources/net/corda/schema/CPI.yaml
```diff
@@ -0,0 +1,22 @@
...
+    name: cpi.info
+    consumers:
+    producers:
+    config:
+      cleanup.policy: compact
+      segment.ms: 300000
+      delete.retention.ms: 300000
+      min.compaction.lag.ms: 60000
+      max.compaction.lag.ms: 300000
+      min.cleanable.dirty.ratio: 0.5
```
davidcurrie: I don't know enough about the topic usage to answer this question but, do we really want/need the ability to tweak the configuration per individual topic vs having different classes of topic? I also don't know whether it's a problem that this is obviously very Kafka specific. Having different classes of topic would make it easier to have a separation between the generic topic definitions and their specific rendering for a given message bus.

data/topic-schema/src/main/resources/net/corda/schema/CPI.yaml
```diff
@@ -0,0 +1,22 @@
+topics:
+  CPIInfoTopic:
+    name: cpi.info
+    consumers:
+    producers:
```
davidcurrie: nit: I'd have preferred to see this introduced only when we had a clear idea of what we're going to do with them

yarivus:
> 

@driessamyn I think we should avoid a single "global" file with all definitions as various teams will be conflicted on source file.
If there's a need for some "horizontal" unification standards for all the various topics - then it sounds right that we should separate those "horizontal" topic attributes to a single file to maintain a common standard.
[BTW we can leverage the object-oriented nature and have the horizontal attributes in some base-class and then have the specific vertical ones in the derived classes]

As for the "vertical" specifics of the various topics I think it's better to have those in separate yaml files, 1 per them.
You guys know better than me whether it should be at a processor level or a higher/lower resolution than that 😉

I would agree w/ @davidcurrie that it's not the most critical issue in the design and this PR specifically.
For me, the more important thing was to make the various attribute values stated in YAML files and not .kt files and have the classes load the YAML attributes.

data/topic-schema/src/main/kotlin/net/corda/schema/Schemas.kt
driessamyn: I agree, that is a must.

The easiest may be for these to loaded from the yaml at runtime. We only have these const as we refer to them in the code, but the cost of loading the yaml file at start-up into a map is probably negligible, and would be easier than code-gen.

driessamyn:
> @driessamyn I think we should avoid a single "global" file with all definitions as various teams will be conflicted on source file. 

These are edited very infrequently and based on what you're saying, it sounds like you suggest the grouping should be based on "team", which would be a strange. Again, by definition, the topics are the coms channel between 2 or more processors. 

> If there's a need for some "horizontal" unification standards for all the various topics - then it sounds right that we should separate those "horizontal" topic attributes to a single file to maintain a common standard. [BTW we can leverage the object-oriented nature and have the horizontal attributes in some base-class and then have the specific vertical ones in the derived classes]

I don't know what this means. Maybe illustrate by example.

> As for the "vertical" specifics of the various topics I think it's better to have those in separate yaml files, 1 per them. You guys know better than me whether it should be at a processor level or a higher/lower resolution than that 😉

That's my point. There isn't an easily identifiable boundary, apart from the "horizontal" ones (processors) and I layed out the problems with grouping it that way.

> I would agree w/ @davidcurrie that it's not the most critical issue in the design and this PR specifically. For me, the more important thing was to make the various attribute values stated in YAML files and not .kt files and have the classes load the YAML attributes.

You approved the PR, meaning you're happy for this to be merged.
How do you suggest this would be managed post-merge? Where will people add new topic definitions and how will the yaml and Kotlin constants be synched?

data/topic-schema/src/main/resources/net/corda/schema/CPI.yaml
mnesbit: I would like to have classes too, with the option to add the odd override. I don't think Ben was sure how best to approach that though

data/topic-schema/src/main/kotlin/net/corda/schema/Schemas.kt
mnesbit: I think the .kt file is probably required. I would love code generation, but it is a pain. We definitely could add a unit test though that reflectively scans the file and checks probably via https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.reflect.full/companion-object-instance.html

fenryka:
> > @driessamyn I think we should avoid a single "global" file with all definitions as various teams will be conflicted on source file.
> 
> These are edited very infrequently and based on what you're saying, it sounds like you suggest the grouping should be based on "team", which would be a strange. Again, by definition, the topics are the coms channel between 2 or more processors.
> 
> > If there's a need for some "horizontal" unification standards for all the various topics - then it sounds right that we should separate those "horizontal" topic attributes to a single file to maintain a common standard. [BTW we can leverage the object-oriented nature and have the horizontal attributes in some base-class and then have the specific vertical ones in the derived classes]
> 
> I don't know what this means. Maybe illustrate by example.
> 
> > As for the "vertical" specifics of the various topics I think it's better to have those in separate yaml files, 1 per them. You guys know better than me whether it should be at a processor level or a higher/lower resolution than that 😉
> 
> That's my point. There isn't an easily identifiable boundary, apart from the "horizontal" ones (processors) and I layed out the problems with grouping it that way.
> 
> > I would agree w/ @davidcurrie that it's not the most critical issue in the design and this PR specifically. For me, the more important thing was to make the various attribute values stated in YAML files and not .kt files and have the classes load the YAML attributes.
> 
> You approved the PR, meaning you're happy for this to be merged. How do you suggest this would be managed post-merge? Where will people add new topic definitions and how will the yaml and Kotlin constants be synched?

I agree with Dries, the only useful stratification we have is that of the LayerCake, it makes sense to separate P2P, Flow, and Ledger concerns since they can be deployed separately and replaced by others.

mnesbit:
I think we have to get this in largely as is for phase 1, because it is now blocking things. So lets raise JIRAs for phase 2 things that might be done by somebody who isn't @benmcmahon100 .

For now before completing this I think we need a units test(s) that reflectively checks the Kotlin Schema.kt file against the YAML resource file content and prevents PR's that add items to either kotlin, or YAML files without the matching change.

Subsequently we need phase 2 JIRAs for:
1. The Schema.kt should not be class X { companion object { } } they should just be object X {}, but that will affect the YAML checks and may tweak a load of imports in corda-runtime-os
2. We need to work out a way to do better templating of config so that there is less duplication, but without us needing to go outside Jackson features in our project dependencies
3. The details of the ACL logic need more fleshing out, especially with regard to mapping processor rights to concrete worker types (since a worker may contain several processors)
4. The precise layout of the YAML is bound to evolve as we identify better structuring of them and of the Schema.kt, so we need a JIRA to do a review and tidy up of this. However, what we have now is enough to deal with the initial kafka topic generation that is blocking things.

mnesbit:
I think we have to get this in largely as is for phase 1, because it is now blocking things. So lets raise JIRAs for phase 2 things that might be done by somebody who isn't @benmcmahon100 .

For now before completing this I think we need a units test(s) that reflectively checks the Kotlin Schema.kt file against the YAML resource file content and prevents PR's that add items to either kotlin, or YAML files without the matching change.

Subsequently we need phase 2 JIRAs for:
1. The Schema.kt should not be class X { companion object { } } they should just be object X {}, but that will affect the YAML checks and may tweak a load of imports in corda-runtime-os
2. We need to work out a way to do better templating of config so that there is less duplication, but without us needing to go outside Jackson features in our project dependencies
3. The details of the ACL logic need more fleshing out, especially with regard to mapping processor rights to concrete worker types (since a worker may contain several processors)
4. The precise layout of the YAML is bound to evolve as we identify better structuring of them and of the Schema.kt, so we need a JIRA to do a review and tidy up of this. However, what we have now is enough to deal with the initial kafka topic generation that is blocking things.

data/topic-schema/src/main/resources/net/corda/schema/CPI.yaml
benmcmahon100: Spoke to Matthew about this, big issue was that jackson didn't support anchors. so some more work will be needed to figure out how to do this without breaking anything or making the toolchain more complicated

data/topic-schema/src/main/resources/net/corda/schema/CPI.yaml
benmcmahon100: I omitted adding any values for this reason. Adding auth requires a change in the prereqs that'll break anyone who's not up to date so will need some announcing

data/topic-schema/src/main/resources/net/corda/schema/CPI.yaml
davidcurrie: I wasn't expecting anything nearly as clever as the use of anchors - just the name of a class/type here and another file that defined the properties for that class/type. It can certainly wait though.

benmcmahon100:
Follow up issues can be found as the following:
- [CORE-4919](https://r3-cev.atlassian.net/browse/CORE-4919)
- [CORE-4920](https://r3-cev.atlassian.net/browse/CORE-4920)
- [CORE-4921](https://r3-cev.atlassian.net/browse/CORE-4921)
- [CORE-4922](https://r3-cev.atlassian.net/browse/CORE-4922)

data/topic-schema/src/test/kotlin/net/corda/schema/SchemaTests.kt
```diff
@@ -0,0 +1,63 @@
...
+            .configure(KotlinFeature.NullToEmptyCollection, true)
+            .configure(KotlinFeature.NullToEmptyMap, true)
+            .configure(KotlinFeature.NullIsSameAsDefault, false)
+            .configure(KotlinFeature.SingletonSupport, false)
+            .configure(KotlinFeature.StrictNullChecks, false)
+            .build()
+    )
+
+    @Suppress("UNCHECKED_CAST")
+    private val yamlFileData by lazy {
```
mnesbit: put type on declaration to make this clearer

data/topic-schema/src/test/kotlin/net/corda/schema/SchemaTests.kt
```diff
@@ -0,0 +1,63 @@
...
+            .map { File(it.file) }
+            .filter { it.isDirectory }
+            .flatMap { it.listFiles()!!.toList() }
+            .filter { it.name.endsWith("yaml") || it.endsWith("yml") }
+            .map { mapper.readValue(it.readText()) as Map<String, *> }
+            as List<Map<String, Map<String, Map<String, *>>>>
+    }
+
+    // Scan companion objects for constant definitions
+    private val memberMap by lazy {
```
mnesbit: put type on declaration to make this clearer

data/topic-schema/src/test/kotlin/net/corda/schema/SchemaTests.kt
```diff
@@ -0,0 +1,63 @@
...
+        }.filter { it.isFinal }
+        val memberValueMap = members.map {
+            it.name to it.call(emptyList<Any>()) as String
+        }
+
+        memberValueMap.toMap()
+    }
+
+    @Test
+    fun `Validate that all yaml topics have matching code value`() {
```
mnesbit, line 72: Longer term we would want this to heck that each section has right subset of the topics

data/topic-schema/src/main/resources/net/corda/schema/Config.yaml
```diff
@@ -0,0 +1,22 @@
...
+    name: config.topic
+    consumers:
+    producers:
+    config:
+      cleanup.policy: compact
+      segment.ms: 300000
+      delete.retention.ms: 300000
+      min.compaction.lag.ms: 60000
+      max.compaction.lag.ms: 300000
+      min.cleanable.dirty.ratio: 0.5
```
mnesbit: Shouldn't the YAML files have a terminal EOL?

benmcmahon100: I can add a newline into each file

data/topic-schema/src/test/kotlin/net/corda/schema/SchemaTests.kt
benmcmahon100: NP, I should be able to parameterize the test

mnesbit (approved):
OK

ronanbrowne (approved):

PR #370, opened 2022-05-16
CORE-3965: Add the protocol to the initiating flow API and adjust session events

JamesHR3:
Adjusts the initiating/initiated flow API to remove the direct link between flow classes in favour of a string tag that the platform matches against, called the protocol. Additionally, a version parameter can be provided, although the corresponding runtime-os PR will not immediately wire this up.

https://github.com/corda/corda-runtime-os/pull/1304 is the corresponding runtime-os PR. This still needs significant work and so this PR remains a draft for now.


mnesbit (approved):
OK

lankydan (approved):

PR #374, opened 2022-05-18
CORE-4783 entity error response

barrylapthorn:
Return richer error information to the flow worker, when it makes a request via the PersistenceService to the db worker (entity processor).

Some categories of errors have been added in corda-api.

Please review those and decide whether there are enough or too many.

See also https://github.com/corda/corda-runtime-os/pull/1323

(yes I probably need to bump the version).

### DIFF
--- a/data/avro-schema/src/main/resources/avro/net/corda/data/virtualnode/EntityResponse.avsc
+++ b/data/avro-schema/src/main/resources/avro/net/corda/data/virtualnode/EntityResponse.avsc
@@ -17,12 +17,11 @@
       "doc": "id of this request for tracking"
     },
     {
-      "name": "result",
-      "doc": "Result of request",
+      "name": "responseType",
+      "doc": "Result of request, either a success or failure message",
       "type": [
-        "null",
-        "bytes",
-        "net.corda.data.ExceptionEnvelope"
+        "net.corda.data.virtualnode.EntityResponseSuccess",
+        "net.corda.data.virtualnode.EntityResponseFailure"
       ]
     }
   ]
--- a/data/avro-schema/src/main/resources/avro/net/corda/data/virtualnode/EntityResponseFailure.avsc
+++ b/data/avro-schema/src/main/resources/avro/net/corda/data/virtualnode/EntityResponseFailure.avsc
@@ -1,0 +1,27 @@
+{
+  "type": "record",
+  "name": "EntityResponseFailure",
+  "namespace": "net.corda.data.virtualnode",
+  "fields": [
+    {
+      "name": "errorType",
+      "doc": "Error type",
+      "type": {
+        "name": "Error",
+        "type": "enum",
+        "doc": "FATAL: almost certainly unrecoverable.\nNOT_READY: the component has not yet started or received all cpks.\nDATABASE: an error during database execution.\nVIRTUAL_NODE: problems retrieving vnode info\nUNKNOWN_COMMAND: an unhandled request payload type",
+        "symbols": [
+          "FATAL",
+          "NOT_READY",
+          "DATABASE",
+          "VIRTUAL_NODE"
+        ]
+      }
+    },
+    {
+      "name": "exception",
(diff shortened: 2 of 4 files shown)


data/avro-schema/src/main/resources/avro/net/corda/data/virtualnode/EntityResponse.avsc
```diff
@@ -17,12 +17,11 @@
...
-      "name": "result",
-      "doc": "Result of request",
+      "name": "responseType",
+      "doc": "Result of request, either a success or failure message",
       "type": [
-        "null",
-        "bytes",
-        "net.corda.data.ExceptionEnvelope"
+        "net.corda.data.virtualnode.EntityResponseSuccess",
+        "net.corda.data.virtualnode.EntityResponseFailure"
```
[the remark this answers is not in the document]
JamesHR3: This works for me. We can always adjust as we go if things don't quite pan out.

I actually prefer this approach anyway as we don't need to put any logic in to figure out which property we should be inspecting - we just need to check what's in here to figure out what to do. But it's not a big deal either way.

gradle.properties
```diff
@@ -7,7 +7,7 @@ cordaProductVersion = 5.0.0
 # NOTE: update this each time this module contains a breaking change
 ## NOTE: currently this is a top level revision, so all API versions will line up, but this could be moved to
 ##   a per module property in which case module versions can change independently.
-cordaApiRevision = 100
+cordaApiRevision = 102
```
JamesHR3, line 10: Have you skipped a version here or is this a case of needing to update the PR?

data/avro-schema/src/main/resources/avro/net/corda/data/virtualnode/EntityResponse.avsc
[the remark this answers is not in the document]
barrylapthorn: I think this is fine as the user would write:

```
when(response.responseType) {
  is EntityResponseSuccess -> process((response.responseType) as EntityResponseSuccess).bytes)
  is EntityResponseFailure -> handleFailure(response.responseType as EntityResponseFailure)
  else -> throw...
}
```

data/avro-schema/src/main/resources/avro/net/corda/data/virtualnode/EntityResponseFailure.avsc
```diff
@@ -0,0 +1,28 @@
...
+        "name": "Error",
+        "type": "enum",
+        "doc": "FATAL: almost certainly unrecoverable.\nNOT_READY: the component has not yet started or received all cpks.\nDATABASE: an error during database execution.\nVIRTUAL_NODE: problems retrieving vnode info\nUNKNOWN_COMMAND: an unhandled request payload type",
+        "symbols": [
+          "FATAL",
+          "NOT_READY",
+          "DATABASE",
+          "VIRTUAL_NODE",
+          "UNKNOWN_COMMAND"
+        ]
```
JamesHR3: I'm fine with starting with this. We might find we can't do very much with the distinctions between the non-fatal errors, but again, this feels like something we can adjust later.

gradle.properties
barrylapthorn: I haven't rebased (due to the version in between making my code uncompiled due to the usual api/runtime mismatch).

https://github.com/corda/corda-api/blob/release/os/5.0/gradle.properties#L10

barrylapthorn: fixed

data/avro-schema/src/main/resources/avro/net/corda/data/virtualnode/EntityResponseFailure.avsc
barrylapthorn: "worst" case with code is:

```
if (errorType == Error.FATAL)  abortFlowWorker() else retry()
```

driessamyn (approved):
LGTM

PR #377, opened 2022-05-20
CORE-4620: Import certificates API

yift-r3:
Add support for the import certificates APIs.

The runtime PR is in https://github.com/corda/corda-runtime-os/pull/1388

This will add messages (and topics) to import certificates and database infrastructure to save the certificates at the cluster level and in the node level.


data/avro-schema/src/main/resources/avro/net/corda/data/membership/rpc/certificates/request/CertificateRpcRequest.avsc
```diff
@@ -0,0 +1,20 @@
+{
+  "type": "record",
+  "name": "CertificateRpcRequest",
+  "namespace": "net.corda.data.membership.rpc.certificates.request",
```
charlieR3: Maybe there is a better package for this since it isn't strictly membership related. i.e. cluster level certificates are a level below membership.

charlieR3: (same comment on all avro schemas here)

data/avro-schema/src/main/resources/avro/net/corda/data/membership/rpc/certificates/response/CertificateImportedRpcResponse.avsc
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "record",
+  "name": "CertificateImportedRpcResponse",
```
charlieR3: If this was `ImportCertificateRpcResponse` the body could have a flag to indicate whether the operation was successful or not. What happens when persistence fails in the currrent case?

data/db-schema/src/main/resources/net/corda/db/schema/cluster-certificates/migration/certificates-creation-v1.0.xml
```diff
@@ -0,0 +1,28 @@
...
+    </column>
+    <column name="raw_certificate" type="TEXT">
+        <constraints nullable="false"/>
+    </column>
+</createTable>
+
+<addPrimaryKey tableName="cluster_certificates" columnNames="tenant_id, alias" constraintName="certificates_pk"
+               schemaName="${schema.name}"/>
+</changeSet>
+</databaseChangeLog>
```
charlieR3, line 28: Looks good but file formatting is a bit mixed up (unless it's just github showing it weird).

data/topic-schema/src/main/kotlin/net/corda/schema/Schemas.kt
```diff
@@ -97,6 +97,8 @@ class Schemas {
             const val PROPOSAL_TOPIC = "membership.proposals"
             const val MEMBERSHIP_RPC_TOPIC = "membership.rpc.ops"
             const val MEMBERSHIP_RPC_RESPONSE_TOPIC = "membership.rpc.ops.resp"
+            const val MEMBERSHIP_RPC_CERTIFICATES_TOPIC = "membership.rpc.certificates.ops"
```
charlieR3: Similar comment as for the avro schemas. Maybe we shouldn't tie this to membership since it can be cluster level.

gradle.properties
```diff
@@ -7,7 +7,7 @@ cordaProductVersion = 5.0.0
 # NOTE: update this each time this module contains a breaking change
 ## NOTE: currently this is a top level revision, so all API versions will line up, but this could be moved to
 ##   a per module property in which case module versions can change independently.
-cordaApiRevision = 109
+cordaApiRevision = 110
```
charlieR3, line 10: Just curious why this was required? Is there a breaking change here?

data/avro-schema/src/main/resources/avro/net/corda/data/membership/rpc/certificates/request/CertificateRpcRequest.avsc
charlieR3: Discussed in stand up.  
I would suggest either leave it as is in `membership` package, despite the cluster level certs not being directly connected to membership, or create a `certificates` package. e.g. `net.corda.data.certificates`.
Any preference @yift-r3 ?

data/avro-schema/src/main/resources/avro/net/corda/data/membership/rpc/certificates/response/CertificateImportedRpcResponse.avsc
yift-r3: If the RPC fails the future will complete exceptionally, and the error will propagate to the APIs.

gradle.properties
yift-r3: The problem is that if we won't explicitly change the number if someone will pull the latest release branch and have an old cache the build will fail.

charlieR3 (approved):

PR #378, opened 2022-05-20
NOTICK - Switch gonepostal infix to p2p-preview

dimosr:
Switching from `gonepostal` to `p2p-preview` to prevent confusion.


ronanbrowne (approved):
LGTM

PR #379, opened 2022-05-23
CORE-4870 - Remove normalisation of CPK metadata.

driessamyn:
Remove normalisation of the CPK metadata in the DB.
Instead, store the JSON encoded AVRO container and the packaging format version.

Also add audit columns and CPK DB Changelog (liquibase) tables.

image
  

runtime-os PR: https://github.com/corda/corda-runtime-os/pull/1346
Green build with Alpha version of this branche: https://ci02.dev.r3.com/blue/organizations/jenkins/Corda5%2Fcorda-runtime-os/detail/PR-1346/11/pipeline

### DIFF
--- a/data/db-schema/src/main/resources/net/corda/db/schema/config/migration/cpx-creation-v1.0.xml
+++ b/data/db-schema/src/main/resources/net/corda/db/schema/config/migration/cpx-creation-v1.0.xml
@@ -36,6 +36,7 @@
             <column name="file_upload_request_id" type="VARCHAR(255)">
                 <constraints nullable="true"/>
             </column>
+            <!-- audit -->
             <column name="entity_version" type="INT">
                 <constraints nullable="false"/>
             </column>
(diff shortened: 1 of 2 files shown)


data/db-schema/src/main/resources/net/corda/db/schema/config/migration/cpx-creation-v1.0.xml
```diff
@@ -209,56 +156,45 @@
...
                 <constraints nullable="false"/>
             </column>
-            <column name="min_platform_version" type="INT">
+            <column name="cpk_version" type="VARCHAR(255)">
                 <constraints nullable="false"/>
             </column>
-            <column name="target_platform_version" type="INT">
+            <column name="cpk_signer_summary_hash" type="VARCHAR(255)">
                 <constraints nullable="false"/>
             </column>
```
kyriathar: I think `cpk_name`, `cpk_version` and `cpk_signer_summary_hash` should be moved in `cpk_metadata` table, as if kept here it means same CPK in different CPI (potentially with different filename) can also have different these 3 properties which should not be true?

driessamyn: Agreed. Change pushed.

kyriathar (dismissed):
LGTM

josephzunigadaly (dismissed):

josephzunigadaly (approved):

JamesHR3 (approved):

PR #380, opened 2022-05-24
CORE-4877 deleteById/findAll persistence methods

barrylapthorn:
See also: https://github.com/corda/corda-runtime-os/pull/1354

Moved Avro messages to own package.

Added DeleteById and FindAll persistence methods.


data/avro-schema/src/main/resources/avro/net/corda/data/persistence/DeleteById.avsc
```diff
@@ -0,0 +1,18 @@
...
+  "doc": "Delete an entity by its id.  See also {@link EntityRequest#request}",
+  "namespace": "net.corda.data.persistence",
+  "fields": [
+    {
+      "name": "entityClassName",
+      "type": "string",
+      "doc": "the fully qualified entity class name"
+    },
+    {
+      "name": "id",
```
driessamyn: We use `primaryKey` as a name for the `FindEntity` schema. Maybe it's worth bringing them in line?

driessamyn (commented):
Just a small comment about naming consistency.

data/avro-schema/src/main/resources/avro/net/corda/data/persistence/DeleteEntityByPrimaryKey.avsc
```diff
@@ -0,0 +1,18 @@
+{
+  "type": "record",
+  "name": "DeleteEntityByPrimaryKey",
+  "doc": "Delete an entity by its id.  See also {@link EntityRequest#request}",
```
driessamyn: I think `DeleteEntityById` was actually correct. It's the JPA object `@Id` annotation that is relevant, it doesn't actually _have_ to be a primary key in the Db.

barrylapthorn: So you want it changed back to DeleteById then?

driessamyn: I think that would make more sense in terms of an API.
Sorry, I should have been clearer in my original comment.

driessamyn (dismissed):
LGTM

driessamyn (approved):

PR #382, opened 2022-05-25
CORE-4927 Added crypto config schema

alexey-kadyrov-r3:


charlieR3 (commented):
Looks good, however all objects by default allow additional properties. Is this the desired behavior? [additionalProperties](http://json-schema.org/understanding-json-schema/reference/object.html#additional-properties)

alexey-kadyrov-r3:
> Looks good, however all objects by default allow additional properties. Is this the desired behavior? [additionalProperties](http://json-schema.org/understanding-json-schema/reference/object.html#additional-properties)

Good point, made it false so all new properties will have to be added explicitly to the schema

charlieR3 (approved):
LGTM

PR #383, opened 2022-05-27
CORE-4864 - Fix FK constraints for CPK tables

driessamyn:
- Fix FK constraints for CPK tables
- Rename cpk to cpk_file to be explicit.

Green associated runtime-os build: https://ci02.dev.r3.com/blue/organizations/jenkins/Corda5%2Fcorda-runtime-os/detail/PR-1379/5/pipeline

### DIFF
--- a/data/db-schema/src/main/resources/net/corda/db/schema/config/migration/cpx-creation-v1.0.xml
+++ b/data/db-schema/src/main/resources/net/corda/db/schema/config/migration/cpx-creation-v1.0.xml
@@ -111,7 +111,7 @@
             </column>
         </createTable>
 
-        <addPrimaryKey tableName="cpk_file" columnNames="file_checksum" constraintName="cpk_pk"
+        <addPrimaryKey tableName="cpk_file" columnNames="file_checksum" constraintName="cpk_file_pk"
                        schemaName="${schema.name}"/>
 
         <addForeignKeyConstraint baseTableName="cpk_file" baseColumnNames="file_checksum"
--- a/data/db-schema/src/main/resources/net/corda/db/schema/config/migration/cpx-creation-v1.0.xml
+++ b/data/db-schema/src/main/resources/net/corda/db/schema/config/migration/cpx-creation-v1.0.xml
@@ -55,29 +55,6 @@
         <addUniqueConstraint tableName="cpi" columnNames="file_checksum" constraintName="db_cpi_uc1"
                              schemaName="${schema.name}"/>
 
-        <createTable tableName="cpk" schemaName="${schema.name}">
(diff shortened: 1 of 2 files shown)


data/db-schema/src/main/resources/net/corda/db/schema/config/migration/cpx-creation-v1.0.xml
```diff
@@ -147,8 +147,8 @@
                        schemaName="${schema.name}"/>
 
         <addForeignKeyConstraint baseTableName="cpk_db_change_log" baseColumnNames="cpk_file_checksum"
-                                 referencedTableName="cpk" referencedColumnNames="file_checksum"
-                                 constraintName="FK_cpk_db_change_log_cpk"
+                                 referencedTableName="cpk_metadata" referencedColumnNames="file_checksum"
```
vkolomeyko, line 150: Foreign key constraint on change log table may not such a good idea.
E.g. it may prevent splitting change log table and referenced table for the purpose of archiving. Or a very old (or mistakenly uploaded) row on the reference table might need to be deleted, in the case it might make sense to keep change log row for audit purposes.

data/db-schema/src/main/resources/net/corda/db/schema/config/migration/cpx-creation-v1.0.xml
```diff
@@ -111,12 +111,12 @@
             </column>
         </createTable>
 
-        <addPrimaryKey tableName="cpk_metadata" columnNames="file_checksum" constraintName="cpk_metadata_pk"
+        <addPrimaryKey tableName="cpk_file" columnNames="file_checksum" constraintName="cpk_pk"
```
kyriathar:
```suggestion
        <addPrimaryKey tableName="cpk_file" columnNames="file_checksum" constraintName="cpk_file_pk"
```

data/db-schema/src/main/resources/net/corda/db/schema/config/migration/cpx-creation-v1.0.xml
driessamyn: Change log is the Liquibase files, so they are extracted from the CPK on installation, so go hand-in-hand with the CPK, so it's effectively more CPK metadata, but it's a one-to-many (a CPK can have multiple DB migration files).
Maybe changelog isn't the best name, but tried to keep it in sync with the terminology Liquibase uses.

kyriathar (commented):
One minor thing, lgtm otherwise.

kyriathar (approved):
LGTM

vkolomeyko (approved):

PR #385, opened 2022-05-30
CORE-5013 Changed the tenant id to be complex type

alexey-kadyrov-r3:


alexey-kadyrov-r3:
implemented differently

PR #389, opened 2022-06-01
CORE-5014 Add soft DB deletion flag at config table 

kyriathar:
- adds `is_deleted` flag at `config` table to allow soft DB deletion.

Complement corda-runtime-os PR [#1395](https://github.com/corda/corda-runtime-os/pull/1395).


driessamyn (approved):
LGTM

PR #394, opened 2022-06-02
CORE-4557: Refactor member key properties based on key design

YashNabar:
https://r3-cev.atlassian.net/browse/CORE-4557
The following key properties exposed by the `MemberInfo` interface have changed -
- `owningKey` changed to `sessionInitiationKey`
- `identityKeys` changed to `ledgerKeys`

Updates Corda API version to 111

### DIFF
--- a/membership/src/test/java/net/corda/v5/membership/MemberInfoJavaApiTest.java
+++ b/membership/src/test/java/net/corda/v5/membership/MemberInfoJavaApiTest.java
@@ -55,17 +55,17 @@ public class MemberInfoJavaApiTest {
     }
 
     @Test
-    public void getIdentityKeys() {
+    public void getLedgerKeys() {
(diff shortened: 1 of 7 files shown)


membership/src/test/java/net/corda/v5/membership/MemberInfoJavaApiTest.java
```diff
@@ -55,17 +55,17 @@ public void getName() {
     }
 
     @Test
-    public void getIdentityKeys() {
+    public void getLedgerKeys() {
```
charlieR3, line 58: A little out of the scope of this, but while you are changing this file could you add a test to this class for getting the session initiation key also since it seems to be missing.

charlieR3 (dismissed):

charlieR3 (approved):

PR #395, opened 2022-06-03
CORE-4743: Flow Startable by RPC redesign

JamesHR3:
This PR changes the `@StartableByRpc` API to instead make it an interface that needs to be implemented. This forces a `call` method that takes a string and returns a string, to better enforce this restriction. Note that the original `Flow` interface is still implemented.


application/src/main/kotlin/net/corda/v5/application/flows/RPCStartableFlow.kt
```diff
@@ -0,0 +1,39 @@
...
+     * @return A JSON encoded string to be supplied to the flow status on flow completion as the result.
+     */
+    @Suspendable
+    fun call(requestBody: String) : String
+
+    /**
+     * By default, RPC startable flows will fail if invoked as a subflow. However, it is possible to implement the
+     * normal flow call method, which allows the flow to also be invoked as a subflow.
+     */
+    @Suspendable
```
owenstanford: I'm not sure we should inherit from Flow<T>, if the user wanted a flow to be RPCstartable and a sub-flow, would they not just implement both interfaces on their flow class?
If Flow<T> is now exclusively for sub-flows should/can we rename it?
obvs same thing for ResponderFlow

lankydan:
Just a note, I think if we are addressing startable by rpc, I think we should consider an interface for initiated flows at the same time. This means we can try end up with something consistent.

application/src/main/kotlin/net/corda/v5/application/flows/ResponderFlow.kt
```diff
@@ -0,0 +1,35 @@
...
+    /**
+     * The business logic for the flow should be written here.
+     *
+     * This is equivalent to the call method for a normal flow. This version is invoked when the flow is started via an
+     * incoming session init event, via a counterparty calling `initiateFlow`.
+     *
+     * @param session The session opened by the counterparty.
+     */
+    @Suspendable
+    fun call(session: FlowSession)
```
lankydan, line 29: Although, it makes sense that there is no return type from the flow, there is an argument that it makes it harder to test because can't verify the "output" or end result of the flow.

If we think that testing via mocks and using `verify` will be sufficient, then going with no return type is fine.

If we keep the return type, then users can still define it as `Unit` if they don't want to return anything, however Java  can't really do that and has to still return `null` which is annoying.

application/src/main/kotlin/net/corda/v5/application/flows/ResponderFlow.kt
```diff
@@ -0,0 +1,35 @@
...
+
+    /**
+     * By default this flow cannot be invoked as a subflow. Implement this to override this behaviour.
+     */
+    @Suspendable
+    override fun call(): T {
```
lankydan: This is something that I was considering, but it also irks me a bit...

application/src/main/kotlin/net/corda/v5/application/flows/RPCStartableFlow.kt
lankydan: I think Owen's point is interesting, that `Flow` is basically only for subflows now. Do we create a `SubFlow` interface instead.

If we then want to still have an overarching linking between all these interfaces, we could basically make `Flow` a marker interface. Also wondering if we keep the generic type on `Flow` even if it doesn't have a method, guess its not that important but we could still use it as a way to "clarify" the return type.

application/src/main/kotlin/net/corda/v5/application/flows/ResponderFlow.kt
lankydan: If we use the code you have at the moment, you should probably remove the generic type off of `ResponderFlow` and just have `Flow<Unit>`.

application/src/main/kotlin/net/corda/v5/application/flows/ResponderFlow.kt
lankydan: The downside to this sort of method is that a user can still override it, because all we've done is provide a default implementation.

application/src/main/kotlin/net/corda/v5/application/flows/RPCStartableFlow.kt
lankydan: Note, that I don't think a RPC flow should also be allowed to run as a subflow. If we have functionality that should be reused between both, it probably best to just write a RPC flow and then call the subflow version.

JamesHR3: I've reworked this to not inherit from `Flow`. I've not changed the name of `Flow` here but we could consider it. There's also an argument for a common marker interface, as the platform previously assumed all flow classes inherited from `Flow`.

application/src/main/kotlin/net/corda/v5/application/flows/ResponderFlow.kt
JamesHR3: I'd rather have the API make sense from a user perspective and work around any testing problems than put something on the API that's misleading to assist our testing. The issue with giving a return type is the user can't do anything with it, we just swallow it.

application/src/main/kotlin/net/corda/v5/application/flows/ResponderFlow.kt
JamesHR3: I've reworked this - I agree it's unpleasant.

application/src/main/kotlin/net/corda/v5/application/flows/RPCStartableFlow.kt
```diff
@@ -0,0 +1,30 @@
...
+ * Flows initiated due to an RPC request receive a JSON encoded string in their call method, containing the body of the
+ * RPC request that initiated this flow. The JSON marshalling service may be used to convert this input string into a
+ * concrete object. The string return type is treated by the platform as a JSON encoded string to return to the RPC
+ * service, and will appear in the RPC flow status when the flow completes. Again, this can be generated by serializing
+ * via the JSON marshalling service.
+ *
+ * Flows implementing this interface must have a no-arg constructor. The flow invocation will fail if this constructor
+ * does not exist.
+ */
+interface RPCStartableFlow {
```
JamesHR3, line 17: I know @mnesbit favours a slightly different approach here where we give two functions - one to set the parameters on the flow, and the current `call` function but with no parameters. I think the argument is that it gives the user the option to throw away some bits of the input request, as otherwise it's always in the checkpoint. I think that API is less clear though. Thoughts?

application/src/main/kotlin/net/corda/v5/application/flows/RPCRequestData.kt
```diff
@@ -0,0 +1,16 @@
+package net.corda.v5.application.flows
+
+/**
+ * A wrapper around the request data for RPC started flows.
+ *
+ * An RPC started flow will receive an instance of this interface, which can be used to retrieve the request body.
+ */
+interface RPCRequestData {
```
lankydan, line 10: Would it be worth just naming this `RPCRequest`? Only reason I say that, is that it's more generic so if we want to add new things to the interface for flows to access, we can do so without breaking apis, adding new ways to do things or having a badly named interface.

While on this subject, should we also pass in the client id into the flow, as part of this interface. It might be useful to users for logging things and currently, they would have to pass it in via the request body. Or, we could provide a way to access it like we do flow id? I know I've gone off on a tangent at this point.

application/src/main/kotlin/net/corda/v5/application/flows/RPCStartableFlow.kt
mnesbit: The wrapper handles stuff for now I think

application/src/main/kotlin/net/corda/v5/application/flows/ResponderFlow.kt
```diff
@@ -0,0 +1,27 @@
...
+
+/**
+ * A flow that may be started by receiving a message from a peer.
+ *
+ * A flow implementing this will also need to be marked as @InitiatedBy to be invoked via a session message. If both
+ * these requirements are met, then the flow will be invoked via the `call` method that takes a session. This session is
+ * created by the platform and communicates with the party that initiated the session.
+ *
+ * Note that an initiated flow cannot return a value.
+ */
```
mnesbit, line 17: Must document the expectation over the constructor

application/src/main/kotlin/net/corda/v5/application/flows/RPCStartableFlow.kt
```diff
@@ -0,0 +1,30 @@
...
+/**
+ * A flow that may be started via RPC.
+ *
+ * Flows initiated due to an RPC request receive a JSON encoded string in their call method, containing the body of the
+ * RPC request that initiated this flow. The JSON marshalling service may be used to convert this input string into a
+ * concrete object. The string return type is treated by the platform as a JSON encoded string to return to the RPC
+ * service, and will appear in the RPC flow status when the flow completes. Again, this can be generated by serializing
+ * via the JSON marshalling service.
+ *
+ * Flows implementing this interface must have a no-arg constructor. The flow invocation will fail if this constructor
```
mnesbit, line 14: Must document expectations over constructors

lankydan:
`JsonConstructor` should also be deleted.

application/src/main/kotlin/net/corda/v5/application/flows/RPCRequestData.kt
JamesHR3: The `clientID` stuff I think we should look at separately. I don't want this PR to get too off topic and I think there's some overlap there with the context propagation work we need to think about.

I'd be careful about making this object more than the request body - there's a danger of having two mechanisms to do similar things with services the flow provides too. So functions for working with the request body are fine, but extra things there I'm less sure is a good idea.

application/src/main/kotlin/net/corda/v5/application/flows/ResponderFlow.kt
JamesHR3: I apparently missed this one but I already have a note over the `RPCStartableFlow`. I've added the same note here.

JamesHR3:
> `JsonConstructor` should also be deleted.

Good catch!

mnesbit (approved):
OK

application/src/main/kotlin/net/corda/v5/application/flows/RPCRequestData.kt
```diff
@@ -0,0 +1,30 @@
...
+    /**
+     * Get the request body and deserialize it into the given type, using a JSON marshalling service.
+     *
+     * @param jsonMarshallingService The JSON marshalling service to use to deserialize this request body.
+     * @return JSON representation of the input request body.
+     */
+    fun <T> getRequestBodyAs(jsonMarshallingService: JsonMarshallingService, clazz: Class<T>) : T
+}
+
+inline fun <reified T> RPCRequestData.getRequestBodyAs(jsonMarshallingService: JsonMarshallingService) : T {
```
lankydan, line 28: Nitpicky, but probably should doc this too.

rick-r3 (approved):

PR #397, opened 2022-06-06
CORE-5040 Config DB/Kafka reconciliation integration

kyriathar:
- updates `Configuration.version` type to int.
- defines JSON schema for Reconciliation and RPC config.
- inserts into `config` table empty sections for Reconciliation and RPC configurations to be defaulted by schema.

Please note: Crypto configuration is still outstanding JIRA: [CORE-5086](https://r3-cev.atlassian.net/browse/CORE-5086).

Complement corda-runtime-os PR: [#1405](https://github.com/corda/corda-runtime-os/pull/1405)

### DIFF
--- a/data/config-schema/src/main/kotlin/net/corda/schema/configuration/ConfigDefaults.kt
+++ b/data/config-schema/src/main/kotlin/net/corda/schema/configuration/ConfigDefaults.kt
@@ -13,9 +13,6 @@ package net.corda.schema.configuration
 object ConfigDefaults {
     const val JDBC_DRIVER = "org.postgresql.Driver"
     const val DB_POOL_MAX_SIZE = 10
-    const val RECONCILIATION_PERMISSION_SUMMARY_INTERVAL_MS_DEFAULT: Long = 60000L
-    const val RECONCILIATION_CPK_WRITE_INTERVAL_MS_DEFAULT: Long = 10000L
-    const val RECONCILIATION_CPI_INFO_INTERVAL_MS_DEFAULT: Long = 10000L
 
     val WORKSPACE_DIR = "${System.getProperty("java.io.tmpdir")}/corda/workspace"
     val TEMP_DIR = "${System.getProperty("java.io.tmpdir")}/corda/tmp"
--- a/data/config-schema/src/main/kotlin/net/corda/schema/configuration/ReconciliationConfig.kt
+++ b/data/config-schema/src/main/kotlin/net/corda/schema/configuration/ReconciliationConfig.kt
@@ -5,4 +5,5 @@ object ReconciliationConfig {
         const val RECONCILIATION_PERMISSION_SUMMARY_INTERVAL_MS = "permissionSummaryIntervalMs"
         const val RECONCILIATION_CPK_WRITE_INTERVAL_MS = "cpkWriteIntervalMs"
         const val RECONCILIATION_CPI_INFO_INTERVAL_MS = "cpiInfoIntervalMs"
+        const val RECONCILIATION_CONFIG_INTERVAL_MS = "configIntervalMs"
 }
--- a/data/config-schema/src/main/resources/net/corda/schema/configuration/rpc/1.0/corda.rpc.json
+++ b/data/config-schema/src/main/resources/net/corda/schema/configuration/rpc/1.0/corda.rpc.json
@@ -4,5 +4,46 @@
   "title": "Corda RPC Configuration Schema",
   "description": "Configuration schema for the RPC subsection.",
   "type": "object",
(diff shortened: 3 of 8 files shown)


data/config-schema/src/main/kotlin/net/corda/schema/configuration/ReconciliationConfig.kt
```diff
@@ -5,4 +5,5 @@ object ReconciliationConfig {
         const val RECONCILIATION_PERMISSION_SUMMARY_INTERVAL_MS = "permissionSummaryIntervalMs"
         const val RECONCILIATION_CPK_WRITE_INTERVAL_MS = "cpkWriteIntervalMs"
         const val RECONCILIATION_CPI_INFO_INTERVAL_MS = "cpiInfoIntervalMs"
+        const val RECONCILIATION_CONFIG_INTERVAL_MS = "configIntervalMs"
```
driessamyn, line 8: I think this, along with the the other `RECONCILIATION_...` should not be part of the boot config. (I think the same for `crypto` and `rpc` but not `db`).

But what's missing from this PR is that I think we need to "prime" the DB with every config key that is dynamic.
We can do this in using Liquibase 'INSERT' (https://docs.liquibase.com/change-types/insert.html)

This would mean that when the config reconciler first runs it will pick up these empty configs and publish defaults for that key to Kafka.

The only thing we need to ensure is that regardless of the `RECONCILIATION_CONFIG_INTERVAL_MS` value, it always has to run at least once.

driessamyn (changes requested):

data/db-schema/src/main/resources/net/corda/db/schema/config/migration/config-creation-v1.0.xml
```diff
@@ -40,6 +42,21 @@
         <addPrimaryKey columnNames="section" constraintName="config_pk" tableName="config"
                        schemaName="${schema.name}"/>
 
+        <insert tableName="config" schemaName="${schema.name}">
+            <column name="section" value="corda.reconciliation"/>
+            <column name="config" value="
+                permissionSummaryIntervalMs=1000,
+                cpkWriteIntervalMs=1000,
+                cpiInfoIntervalMs=1000,
+                configIntervalMs=1000"/>
```
driessamyn: These should be empty as the defaults will be applied by the default logic on first config reconciliation.

data/config-schema/src/main/kotlin/net/corda/schema/configuration/ConfigDefaults.kt
```diff
@@ -13,9 +13,6 @@ package net.corda.schema.configuration
 object ConfigDefaults {
     const val JDBC_DRIVER = "org.postgresql.Driver"
     const val DB_POOL_MAX_SIZE = 10
-    const val RECONCILIATION_PERMISSION_SUMMARY_INTERVAL_MS_DEFAULT: Long = 60000L
-    const val RECONCILIATION_CPK_WRITE_INTERVAL_MS_DEFAULT: Long = 10000L
-    const val RECONCILIATION_CPI_INFO_INTERVAL_MS_DEFAULT: Long = 10000L
```
driessamyn, line 18: I think these defaults need to be in the json schema.

kyriathar: Good catch, yes indeed, will create the reconciliation schema and add the defaults. For now making sure things work when porting boot config to the database to be reconciled.

data/db-schema/src/main/resources/net/corda/db/schema/config/migration/config-creation-v1.0.xml
kyriathar: Done

data/config-schema/src/main/kotlin/net/corda/schema/configuration/ConfigDefaults.kt
kyriathar: Done

data/config-schema/src/main/resources/net/corda/schema/configuration/rpc/1.0/corda.rpc.json
```diff
@@ -4,5 +4,46 @@
   "title": "Corda RPC Configuration Schema",
```
driessamyn, line 4: @vkolomeyko FYI

driessamyn (dismissed):
LGTM

driessamyn (dismissed):
LGTM

driessamyn (approved):
LGTM

PR #401, opened 2022-06-08
CORE-4868 - add entityVersion to CpkIdentifier.

conal-smith:
As part of https://github.com/corda/corda-runtime-os/pull/1385 we will add entityVersion to CpkIdentifier avro object so this field can be propagated across kafka and used in CpkIdentifier when accessing maps in the cache, etc.

### DIFF
--- a/data/avro-schema/src/main/resources/avro/net/corda/data/packaging/CpkIdentifier.avsc
+++ b/data/avro-schema/src/main/resources/avro/net/corda/data/packaging/CpkIdentifier.avsc
@@ -14,6 +14,10 @@
     {
       "name": "signerSummaryHash",
       "type": ["null", "net.corda.data.crypto.SecureHash"]
+    },
+    {
+      "name": "entityVersion",
+      "type": ["null", "int"]
     }
   ]
 }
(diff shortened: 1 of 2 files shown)


data/avro-schema/src/main/resources/avro/net/corda/data/packaging/CpkIdentifier.avsc
```diff
@@ -14,6 +14,10 @@
     {
       "name": "signerSummaryHash",
       "type": ["null", "net.corda.data.crypto.SecureHash"]
+    },
+    {
+      "name": "entityVersion",
+      "type": ["null", "int"]
```
sprintengine: doesn't need to be nullable

vkolomeyko (approved):
The change looks fine to me.

Given alpha version of `corda-api` has already been produced, before merging this change, I would recommend preparing `corda-runtime-os` PR to accommodate this particular change.

That way we will minimise delay for other people who might be needing to change `corda-api` after this change is merged.

conal-smith:
Corresponding runtime-os change pointed to this alpha build here - https://github.com/corda/corda-runtime-os/pull/1423

driessamyn (approved):
LGTM

FYI @josephzunigadaly - we may need to look whether this has any implication on the serialization code, for example.

PR #404, opened 2022-06-09
CORE-4998 Flow mapper cleanup

lankydan:
Add `hasScheduledCleanup` to `SessionState`.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -7,7 +7,7 @@ cordaProductVersion = 5.0.0
 # NOTE: update this each time this module contains a breaking change
 ## NOTE: currently this is a top level revision, so all API versions will line up, but this could be moved to
 ##   a per module property in which case module versions can change independently.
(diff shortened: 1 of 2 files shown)


gradle.properties
```diff
@@ -7,7 +7,7 @@ cordaProductVersion = 5.0.0
 # NOTE: update this each time this module contains a breaking change
 ## NOTE: currently this is a top level revision, so all API versions will line up, but this could be moved to
 ##   a per module property in which case module versions can change independently.
-cordaApiRevision = 111
```
JamesHR3: I think the current version is 113 so you'll need an update here (and in runtime-os presumably).

JamesHR3: I think the current version is 113 so you'll need an update here (and in runtime-os presumably).

JamesHR3 (dismissed):

JamesHR3 (approved):

PR #405, opened 2022-06-09
CORE-648 fixes based on experience implementing an HSM

alexey-kadyrov-r3:
* CryptoService should report key schemes and supported signature specs for each key scheme as some of the EC curves are linked to which digest can be used by signing (eg. spec256r1 should use SHA256withECDSA)
* Supporting the throttling on the platform level
* Instead of using optional interfaces for additional functionality (like delete) it should be single interface coupled with reporting of capabilities (due decorators to handle common fault conditions and throttling)
* Signature spec made polymorphic to simply use in the application layer


cipher-suite/src/main/kotlin/net/corda/v5/cipher/suite/CryptoService.kt
```diff
@@ -10,41 +10,38 @@ import net.corda.v5.crypto.exceptions.CryptoServiceException
  * Note about key aliases. Corda always uses single alias to identify a key pair however some HSMs need separate
  * aliases for public and private keys, in such cases their names have to be derived from the single key pair alias.
  * It could be suffixes or whatever internal naming scheme is used.
- * Also note that's not required to keep a public
- * key in the HSM as that will be kept by the upstream Crypto Services.
+ *
+ * Also note that's not required to keep a public key in the HSM as that will be kept by the upstream Crypto Services.
```
rick-r3: that's not -> that it is not

cipher-suite/src/main/kotlin/net/corda/v5/cipher/suite/CryptoService.kt
```diff
@@ -10,41 +10,38 @@ import net.corda.v5.crypto.exceptions.CryptoServiceException
...
+ *
+ * Exception handling.
+ *
+ * As the service instances are decorated with decorators taking care about the timeout, throttling handling,
```
rick-r3: care about -> care of

cipher-suite/src/main/kotlin/net/corda/v5/cipher/suite/CryptoService.kt
```diff
@@ -10,41 +10,38 @@ import net.corda.v5.crypto.exceptions.CryptoServiceException
...
+ * exception normalisation the exceptions listed for each method are as expected be thrown by the decorators.
```
rick-r3: exception normalisation -> and exception normalisation
expected be thrown -> expected to be thrown

cipher-suite/src/main/kotlin/net/corda/v5/cipher/suite/CryptoService.kt
```diff
@@ -10,41 +10,38 @@ import net.corda.v5.crypto.exceptions.CryptoServiceException
...
+ *
+ * The service should throw the most appropriate exception as it sees it, e.g. if the input parameters are wrong it
+ * should be [IllegalArgumentException] or if the internal state is wrong for an operation then the most appropriate
+ * exception would be [IllegalStateException].
+ *
+ * If service encountered throttling situation then it should throw one
+ * of concrete implementations of [CSLThrottlingException] (such as [CSLExponentialThrottlingException]) so the
+ * library can handle it appropriately.
+ *
+ * The next exceptions are retried - [TimeoutException], [CryptoServiceLibraryException] (with the isRecoverable flage
```
rick-r3: The next exceptions -> The following exceptions

rick-r3 (dismissed):

mnesbit (approved):
ok

PR #406, opened 2022-06-10
CORE-5081: Upgrade to Kotlin 1.7.0.

chrisr3:
[Kotlin 1.7.0 has been released](https://blog.jetbrains.com/kotlin/2022/06/kotlin-1-7-0-released).

### DIFF
--- a/corda-api/build.gradle
+++ b/corda-api/build.gradle
@@ -21,13 +21,41 @@ dependencies {
         api project(':persistence')
         api project(':serialization')
 
-        api "javax.persistence:javax.persistence-api:$javaxPersistenceApiVersion"
-        api "net.corda.kotlin:kotlin-stdlib-jdk8-osgi:$kotlinVersion"
-        api "net.corda.kotlin:kotlin-stdlib-jdk7-osgi:$kotlinVersion"
-        api "org.osgi:osgi.annotation:$osgiVersion"
-        api "org.osgi:osgi.core:$osgiVersion"
-        api "org.osgi:org.osgi.service.component.annotations:$osgiScrAnnotationVersion"
-        api "org.slf4j:slf4j-api:$slf4jVersion"
+        api('javax.persistence:javax.persistence-api') {
+            version {
+                require javaxPersistenceApiVersion
+            }
+        }
+        api('net.corda.kotlin:kotlin-stdlib-jdk8-osgi') {
+            version {
+                require kotlinVersion
+            }
+        }
+        api('net.corda.kotlin:kotlin-stdlib-jdk7-osgi') {
+            version {
+                require kotlinVersion
+            }
+        }
+        api('org.osgi:osgi.annotation') {
+            version {
+                require osgiVersion
+            }
+        }
+        api('org.osgi:osgi.core') {
+            version {
+                require osgiVersion
+            }
+        }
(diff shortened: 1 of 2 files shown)


vkolomeyko:
We probably need to perform this sort of update of the following projects as well:
https://github.com/corda/corda-api
https://github.com/corda/corda-gradle-plugins
https://github.com/corda/corda-cli-plugin-host

chrisr3:
> We probably need to perform this sort of update of the following projects as well.

I am already updating `corda-api` and `corda-runtime-os`. However, any Corda Gradle plugin written in Kotlin will use Gradle's own version of Kotlin and not Corda's.

vkolomeyko:
Thank you @chrisr3 
As for upgrading https://github.com/corda/corda-cli-plugin-host project - @pmck91 could you raise a Jira for that so that we could prioritize?

vkolomeyko (dismissed):

gradle.properties
```diff
@@ -7,10 +7,10 @@ cordaProductVersion = 5.0.0
 # NOTE: update this each time this module contains a breaking change
 ## NOTE: currently this is a top level revision, so all API versions will line up, but this could be moved to
 ##   a per module property in which case module versions can change independently.
-cordaApiRevision = 112
+cordaApiRevision = KOTLIN17
```
vkolomeyko: Surely this cannot be right...

corda-api/build.gradle
```diff
@@ -21,13 +21,41 @@ dependencies {
...
-        api "javax.persistence:javax.persistence-api:$javaxPersistenceApiVersion"
-        api "net.corda.kotlin:kotlin-stdlib-jdk8-osgi:$kotlinVersion"
-        api "net.corda.kotlin:kotlin-stdlib-jdk7-osgi:$kotlinVersion"
-        api "org.osgi:osgi.annotation:$osgiVersion"
-        api "org.osgi:osgi.core:$osgiVersion"
-        api "org.osgi:org.osgi.service.component.annotations:$osgiScrAnnotationVersion"
-        api "org.slf4j:slf4j-api:$slf4jVersion"
+        api('javax.persistence:javax.persistence-api') {
+            version {
+                require javaxPersistenceApiVersion
```
vkolomeyko, line 26: I am assuming without this instruction, Kotlin 1.7 also causes JPA version to upgrade?

gradle.properties
chrisr3: It's WIP, so that `corda-runtime-os` can find artifacts in Artifactory to compile against :wink:.

corda-api/build.gradle
chrisr3: I am just making the version constraints _explicit_, so that we know precisely how Gradle will interpret them. There should be no functional change, although I was concerned that Gradle was using `strict` here instead of `require`.

gradle.properties
chrisr3: Now updated to 113 :grinning:

chrisr3:
This PR is intended to be a precursor to [corda-runtime-os#1426](https://github.com/corda/corda-runtime-os/pull/1426).

driessamyn (approved):
LGTM

mnesbit (approved):
OK

PR #411, opened 2022-06-13
CORE-3770 flow db integration

LWogan:
Add a query object to the checkpoint to represent the most recent sent db query to the db worker.
Add flowId to the entity request to allow it to be returned on the right key.
Add cats and dogs schemas to vnode schema to simplify testing
Add config for db manager to flow config
Expose persistence module as an api dependency of  application due to CordaPersistenceException being an error that can be thrown from the persistence api

runtime of pr: https://github.com/corda/corda-runtime-os/pull/1432


data/avro-schema/src/main/resources/avro/net/corda/data/flow/state/db/Query.avsc
```diff
@@ -0,0 +1,39 @@
+{
+  "type": "record",
+  "name": "Query",
+  "namespace": "net.corda.data.flow.state.db",
```
lankydan: Should we make this `PersistenceState` or `DatabaseState`? To keep it consistent with `SessionState`.

data/avro-schema/src/main/resources/avro/net/corda/data/persistence/EntityRequest.avsc
```diff
@@ -12,6 +12,11 @@
       },
       "doc": "Time ({@link java.time.Instant}) in milliseconds when the record was created."
     },
+    {
+      "name": "flowId",
+      "type": "string",
+      "doc": "The unique ID for tracking the flow. Responses will be returned on this key."
+    },
     {
```
lankydan, line 20: If we have the `flowId` here then I doubt we also need the `FlowKey`.

data/config-schema/src/main/resources/net/corda/schema/configuration/flow/1.0/corda.flow.json
```diff
@@ -53,6 +53,25 @@
           "default": 60000
         }
       }
+    },
+    "db": {
```
lankydan: @JamesHR3 @rick-r3 @driessamyn This is a wider question. 

But are we going to stick to `persistence` or `db` for naming conventions? `persistence` is what we use in the API. `db` is used internally, e.g. `DBWorker`. Having some consistency around these things will probably make things easier, for example, when users are uploading configuration they shouldn't have to think about whether it comes under `db` or `persistence`, it should be obvious because we've chosen to stick to a single wording.

data/avro-schema/src/main/resources/avro/net/corda/data/persistence/EntityRequest.avsc
LWogan: We need holding id to get the right sandbox. But we don't need the client request id so FlowKey might be overboard.

lankydan: Ah I see, ye probably just the holding identity would suffice as you said.

data/avro-schema/src/main/resources/avro/net/corda/data/flow/state/db/Query.avsc
LWogan: After standup discussion i'll rename query to persistenceState

data/config-schema/src/main/resources/net/corda/schema/configuration/flow/1.0/corda.flow.json
LWogan: After standup discussion i'll rename "db" config field in the flow config to "persistence"

lankydan (dismissed):

lankydan (approved):

benmcmahon100 (approved):

PR #412, opened 2022-06-13
CORE-3856 - Add configuration key and default for Max Allowed Message Size.

driessamyn:
Set this as 950Kb to leave some room for the AVRO message envelope.

Related runtime-os PR: https://github.com/corda/corda-runtime-os/pull/1437

### DIFF
--- a/data/config-schema/src/main/resources/net/corda/schema/configuration/messaging/1.0/corda.messaging.json
+++ b/data/config-schema/src/main/resources/net/corda/schema/configuration/messaging/1.0/corda.messaging.json
@@ -101,9 +101,11 @@
       "default": {}
     },
     "maxAllowedMessageSize": {
-      "description": "Maxiumum allowed size, in bytes, for publishing Messages. This configuration does _not_ change the Kafka configuration itself.",
+      "description": "Maximum allowed size, in bytes, for publishing Messages. This configuration is used by the platform as a guide and does _not_ change the underlying bus settings.",
       "type": "number",
-      "default": 972800
+      "default": 972800,
+      "minimum": 512000,
+      "maximum": 8388608
     }
   },
   "additionalProperties": false
--- a/data/config-schema/src/main/resources/net/corda/schema/configuration/messaging/1.0/corda.messaging.json
+++ b/data/config-schema/src/main/resources/net/corda/schema/configuration/messaging/1.0/corda.messaging.json
@@ -99,6 +99,11 @@
       "description": "Settings for publishers that write messages to the underlying bus",
       "$ref": "publisher.json",
       "default": {}
+    },
+    "maxAllowedMessageSize": {
+      "description": "Maxiumum allowed size, in bytes, for publishing Messages. This configuration does _not_ change the Kafka configuration itself.",
+      "type": "number",
+      "default": 972800
     }
   },
   "additionalProperties": false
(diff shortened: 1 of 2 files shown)


data/config-schema/src/main/resources/net/corda/schema/configuration/messaging/1.0/corda.messaging.json
```diff
@@ -99,6 +99,11 @@
       "description": "Settings for publishers that write messages to the underlying bus",
       "$ref": "publisher.json",
       "default": {}
+    },
+    "maxAllowedMessageSize": {
+      "description": "Maxiumum allowed size, in bytes, for publishing Messages. This configuration does _not_ change the Kafka configuration itself.",
```
JamesHR3: Strictly speaking the underlying bus might not be Kafka, so might be worth rephrasing that this is used by the platform as a guide and has no effect on underlying bus settings.

data/config-schema/src/main/resources/net/corda/schema/configuration/messaging/1.0/corda.messaging.json
```diff
@@ -99,6 +99,11 @@
...
+      "type": "number",
+      "default": 972800
```
JamesHR3: Wonder if it's worth adding some sort of minimum (although I'm not totally sure what that should be) - I expect we'd get some instability if this is set too small.

data/config-schema/src/main/resources/net/corda/schema/configuration/messaging/1.0/corda.messaging.json
driessamyn: Thanks. Text updated.

data/config-schema/src/main/resources/net/corda/schema/configuration/messaging/1.0/corda.messaging.json
driessamyn: Good point. I added both minimum and maximum.
We can review both values, but I choose 500kb as a minimum as I thought going small may compromise the performance of the platform, and I picked 8Mb as the maximum as that's what we currently have as the size of the chunks column in the DB.

JamesHR3 (dismissed):

data/config-schema/src/main/resources/net/corda/schema/configuration/messaging/1.0/corda.messaging.json
```diff
@@ -99,6 +99,13 @@
       "description": "Settings for publishers that write messages to the underlying bus",
       "$ref": "publisher.json",
       "default": {}
+    },
+    "maxAllowedMessageSize": {
```
LWogan, line 103: Does still apply to all messages from all producers? or just for messages that are being chunked. The name or description don't make that explicit if its just chunk size for large messages.

driessamyn: It's for all messages. Basically, it says never publish a message larger than ....

Currently we have 2 usecases: chunking and we check the AMQP serialsed objects before publishing them to Kafka after a DB query.

LWogan: Right, it just the minimum didn't make a lot of sense to me for the case of a normal publish of one simple avro record. Do we know a single avro record can't be below 500b? If it was, will the min just be ignored and we send anyway? So min would apply to chunks but max applies to all messages.

driessamyn: The minimum here is just the json schema validation. It's the minimum you're allowed to configure. So basically, we're saying you're not allowed to configure the maximum to be less than 500kb.

JamesHR3 (approved):

PR #408, opened 2022-06-10
CORE-4950: Move local hosted identities out of tests

yift-r3:
This PR:
1. Add RPC command to read a certificate chain.
2. Change the save certificate to accept a certificate chain.
3. Move the holding identity from the `p2p.test` to the `p2p` package.

The runtime PR is in https://github.com/corda/corda-runtime-os/pull/1427

PR #413, opened 2022-06-13
NOTICK - Add missing p2p topics and update configuration for compacted topics

dimosr:
Adding some p2p topics that were missing and adding the appropriate configuration to the topics that are supposed to be compacted.


data/topic-schema/src/main/kotlin/net/corda/schema/Schemas.kt
```diff
@@ -118,13 +118,21 @@ class Schemas {
         companion object {
             const val P2P_OUT_TOPIC = "p2p.out"
             const val P2P_OUT_MARKERS = "p2p.out.markers"
+            const val P2P_OUT_MARKERS_STATE = "p2p.out.markers.state"
```
yift-r3: I think we need to add `dlq` as well

(https://github.com/corda/corda-api/pull/408/files)

dimosr: good point - let me add that

yift-r3 (approved):
LGTM

dimosr:
@yift-r3 thanks for reviewing - I will hold off merging this one [until this one](https://github.com/corda/corda-runtime-os/pull/1433) has been approved so that they can both be merged in quick succession.

PR #416, opened 2022-06-14
CORE-2656: add replay marker class to avro

emilybowe:
This PR will add a new marker type for replays in the corda-api.
This will be used in corda-runtime-os in the Link Manager to write a marker on every replay, so that they can be tracked centrally from Kafka.


data/avro-schema/src/main/resources/avro/net/corda/data/p2p/markers/LinkManagerReplayMarker.avsc
```diff
@@ -0,0 +1,13 @@
+{
+  "type": "record",
+  "name": "LinkManagerReplayMarker",
+  "namespace": "net.corda.p2p.markers",
+  "fields": [
+    {
+      "name": "message",
```
dimosr: I think we can omit this field to save storage/bandwidth etc. We include it in the sent marker, so that we can easily replay messages, but we don't include it in the other markers as they can all be linked together using the key which is the message identifier.

williamvigorr3: I think we do need the `message`. I also think in you need to change the `MessageTracker` in the `DeliveryTracker` in `https://github.com/corda/corda-runtime-os/pull/1469` to process the a `LinkManagerReplayMarker` in the same way as a `LinkManagerSentMarker`.

The reason we need to do this, is that the markers topic is a compacted topic so we are only guaranteed to observe the last marker for a given message. Otherwise we will only be able to a replay messages once.

dimosr: The markers topic was never meant to be a compacted topic I believe. If it is now, we might have gone somewhere wrong. 

The markers topic was supposed to be a purely event stream and the corresponding markers state topic was supposed to be the compacted one.

dimosr: checking the latest setup that seems to be the case (markers topic isn't a compacted one): https://github.com/corda/corda-runtime-os/blob/p2p-preview/applications/tools/p2p-test/p2p-setup/src/main/resources/p2p-kafka-setup-example.conf#L125-L153

williamvigorr3: I think you are correct. We can probably just change the MessageTracker to update the `state` with the last `state` instead then, when a `LinkManagerReplayedMarker` is observed.

dimosr: yes, I think the message tracker can even remain agnostic of the replay markers, it only cares for sent & received/expired for the replay logic.

data/avro-schema/src/main/resources/avro/net/corda/data/p2p/markers/LinkManagerReplayMarker.avsc
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "record",
+  "name": "LinkManagerReplayMarker",
+  "namespace": "net.corda.p2p.markers",
+  "fields": [],
+  "doc": "The message was processed by the sending Link Manager."
```
williamvigorr3: "The message was replayed by the sending Link Manager."?

williamvigorr3 (commented):
Looks good.

gradle.properties
```diff
@@ -7,7 +7,9 @@ cordaProductVersion = 5.0.0
 # NOTE: update this each time this module contains a breaking change
 ## NOTE: currently this is a top level revision, so all API versions will line up, but this could be moved to
 ##   a per module property in which case module versions can change independently.
-cordaApiRevision = 128
+
```
williamvigorr3: Slightly nit picky: Can you remove the extra whitespace added on line 10 + 11. As everyone in the platform team regularly edits this line.

williamvigorr3 (commented):
I left one nit picky comment, otherwise looks good.

williamvigorr3 (approved):
LGTM!!!

dimosr (approved):

PR #418, opened 2022-06-14
CORE-5087 - added platform exception reporting in the flow checkpoint

owenstanford:
Adds a new field to the flow checkpoint to enable platform exceptions to be sent back to user code.


JamesHR3 (commented):
I'm happy to approve this when the time comes. Does this need an API bump? I'm not sure if the code will run with this change as-is - it'll depend a bit on how avro objects are being constructed through the runtime-os code. (In theory though you shouldn't need to.)

JamesHR3 (approved):

PR #419, opened 2022-06-15
CORE-5102: Introduce a common `Flow` marker interface and generic `MarshallingService`

JamesHR3:
This is a follow up to https://github.com/corda/corda-api/pull/395 that introduces a common `Flow` marker interface that all flow types inherit from. This allows for better type safety within the platform, as well as the ability to scan for all flows in the cordapp in the CPK metadata. The old `Flow` interface is renamed to `Subflow`, representing a flow that can only be started by other flows.

A generic `MarshallingService` interface is also introduced for parsing and formatting data into particular string formats. Currently the only implementation provided by the platform is for JSON. RPC request data is changed to allow marshalling to/from any supported marshalling service.


application/src/main/kotlin/net/corda/v5/application/flows/Subflow.kt
```diff
@@ -0,0 +1,24 @@
...
+ */
+interface Subflow<out T> : Flow {
+
+    /**
+     * This is where you fill out your business logic.
+     *
+     * @throws FlowException It can be thrown at any point of a [Flow] logic to bring it to a permanent end. The exception will be
+     * propagated to all counterparty flows.
+     * @throws CordaRuntimeException General type of exception thrown by most Corda APIs.
+     * @throws UnexpectedFlowEndException Thrown when a flow session ends unexpectedly.
```
lankydan: `UnexpectedFlowEndException` doesn't exist anymore

application/src/main/kotlin/net/corda/v5/application/marshalling/JsonMarshallingService.kt
```diff
@@ -0,0 +1,32 @@
...
+
+    /**
+     * Deserializes the [input] JSON into a list of instances of [T].
+     *
+     * @param input The JSON to deserialize.
+     * @param clazz The [Class] type to deserialize into.
+     *
+     * @return A new list of [T].
+     */
+    fun <T> parseList(input: String, clazz: Class<T>): List<T>
```
lankydan: Would it make sense for this to exist on the `MarshallingService` as well? I guess if there was a `CSVMarshalling` service then it wouldn't make sense, but it would for things like XML (not sure what else there is). Mainly thinking about this in terms of `RPCRequestData.getRequestBodyAs` would require them to create a class containing a list to be able to achieve the same functionality as `parseList`.

application/src/main/kotlin/net/corda/v5/application/flows/Flow.kt
```diff
@@ -1,17 +1,12 @@
...
+/**
+ * Marker interface from which all flow types inherit.
+ *
+ * This should not be implemented directly when writing a flow. Instead, one of the more specialised types of flow
+ * should be implemented:
+ * - [RPCStartableFlow] for flows that can be started via RPC
+ * - [ResponderFlow] for flows that can be started via a session
+ * - [Subflow] for flows that can be started from another flow
+ */
+interface Flow
```
mnesbit, line 15: can we put @DoNotImplement on from base?

application/src/main/kotlin/net/corda/v5/application/flows/FlowEngine.kt
```diff
@@ -32,7 +32,7 @@ interface FlowEngine {
      * [Flow]s it communicated with. The subflow can be retried by catching this exception.
      */
     @Suspendable
-    fun <R> subFlow(subLogic: Flow<R>): R
+    fun <R> subFlow(subLogic: Subflow<R>): R
```
mnesbit: Should we change to ` fun <R> subFlow(subLogic: SubFlow<R>, context: Map<String, String>): R` now too? Save ourselves breaking this API in another 2 minutes? Can throw not implemented if context is non-empty for now, but be good to settle the signature

application/src/main/kotlin/net/corda/v5/application/marshalling/JsonMarshallingService.kt
```diff
@@ -0,0 +1,32 @@
+package net.corda.v5.application.marshalling
+
+import net.corda.v5.base.annotations.DoNotImplement
+
+/**
+ * An optional service CorDapps and other services may use to marshall arbitrary content in and out of JSON format using standard/approved
+ * mappers.
+ */
+@DoNotImplement
+interface JsonMarshallingService : MarshallingService {
```
mnesbit: The real issue with this service is that it provides the user no opportunity to adjust it, or register customer serializers. I hope there is a JIRA for that and we will need to create yet another marker interface/annotation for those custom serializers in same way as we have for Kryo and AMQP

application/src/main/kotlin/net/corda/v5/application/flows/FlowEngine.kt
lankydan: Definitely worth providing both on the API if there is a high chance of passing in `emptyMap` from a users perspective.

lankydan: Can we also change `subLogic` to `subFlow` or `flow`.

application/src/main/kotlin/net/corda/v5/application/flows/Subflow.kt
```diff
@@ -0,0 +1,24 @@
...
+import net.corda.v5.base.annotations.Suspendable
+import net.corda.v5.base.exceptions.CordaRuntimeException
+
+/**
+ * A flow that should be started by another flow.
+ *
+ * When a flow implementing [Subflow] is called via the subflow API (see [FlowEngine]), the [call] method of this
+ * interface will be invoked.
+ */
+interface Subflow<out T> : Flow {
```
lankydan: Is `Subflow` or `SubFlow` the correct spelling 🤔 honest question.

ledger/src/main/kotlin/net/corda/v5/ledger/notary/NotaryService.kt
```diff
@@ -39,7 +39,7 @@ abstract class NotaryService : SingletonSerializeAsToken {
      * Produces a notary service flow which has the corresponding sends and receives as [NotaryClientFlow][net.corda.systemflows.NotaryClientFlow].
      * @param otherPartySession client [Party] making the request
      */
-    abstract fun createServiceFlow(otherPartySession: FlowSession): Flow<Void?>
+    abstract fun createServiceFlow(otherPartySession: FlowSession): ResponderFlow
```
mnesbit, line 42: This shouldn't exist as contract for notary should stop at the client side which we factory up based upon advertised notary type in MGM. We should not be requiring that anything is a flow in their protocol. Only contract should be that we can get back a signature if OK, else an error of not valid. The details of transport to and from notary should be absolutely opaque. Ultimately we need to talk to @relyafi about this

application/src/main/kotlin/net/corda/v5/application/flows/Subflow.kt
lankydan: `FlowEngine.subFlow` is spelt with a capital `F`. We should choose one of them at least, I think `SubFlow` is probably correct.

application/src/main/kotlin/net/corda/v5/application/marshalling/JsonMarshallingService.kt
JamesHR3: I left it off as it's not clear to me it will always be possible to implement it. I think it's there on the `JsonMarshallingService` because constructing the List `clazz` is non-trivial (the implementation reaches into the object mapper to do it). I think it not always being possible to implement is a bit of a weak argument though, as we could probably work around by e.g. returning a list with a single item in for cases where it doesn't make sense.

application/src/main/kotlin/net/corda/v5/application/flows/FlowEngine.kt
JamesHR3: I would much rather do this alongside other API changes in the context propagation epic rather than introducing it ad-hoc - we're going to have to revisit a lot of API when we come to that epic anyway. We're not going to have a settled API for a little while yet unfortunately.

application/src/main/kotlin/net/corda/v5/application/flows/Subflow.kt
JamesHR3: Yeah we're going to need to be consistent here! I prefer lowercase personally, but seeing as it's already uppercase everywhere I think we're sort of settled on that variant, so I'll change this.

ledger/src/main/kotlin/net/corda/v5/ledger/notary/NotaryService.kt
JamesHR3: This change is purely to keep things compiling. The ledger API needs an overhaul along the lines of the flow API (this is part of why the API is not going to be stable for some time).

lankydan (dismissed):

application/src/main/kotlin/net/corda/v5/application/marshalling/JsonMarshallingService.kt
rick-r3: I have a feeling we need this moved up to `MarshallingService`.  Otherwise I don't think we'll be able to have list types as the top level JSON type as I think you have to tell Jackson what you're expecting for security?

i.e. probably need a `getRequestBodyAsList()` that uses it.

Worth tests to see if that is so?

Begs the question, what about Map (or is that not a JSON thing)?  But okay to leave that for later, definitely.

JamesHR3: If we want a `getRequestBodyAsList()` then we probably need to move this down a level regardless. I'll have a look at that this afternoon.

For wider collections I think Jackson has the same thing, it's just we only implemented it for lists before. That would need a little more thought on the implementation side I think.

application/src/main/kotlin/net/corda/v5/application/marshalling/JsonMarshallingService.kt
JamesHR3: https://r3-cev.atlassian.net/browse/CORE-5190

owenstanford (dismissed):

application/src/main/kotlin/net/corda/v5/application/marshalling/MarshallingService.kt
```diff
@@ -0,0 +1,66 @@
...
+}
+
+/**
+ * Deserializes the [input] into a list of instances of [T].
+ *
+ * @param input The input string to deserialize.
+ *
+ * @return A new list of [T].
+ */
+inline fun <reified T> JsonMarshallingService.parseList(input: String): List<T> {
```
rick-r3: Should be `MarshallingService`

rick-r3 (changes requested):

ledger/src/main/kotlin/net/corda/v5/ledger/notary/NotaryService.kt
relyafi: Yep, I suspect this is simply down to the fact that we're still using the old C4 code in the new C5 repos - this code doesn't exist at all in our DP1 based dev branches, so should be removed altogether when we come to port it across to the new repos.

application/src/main/kotlin/net/corda/v5/application/marshalling/MarshallingService.kt
JamesHR3: 🤦

rick-r3 (approved):

PR #400, opened 2022-06-07
CORE-4868 - change PK of cpk_metadata to be the name, version and signer hash

conal-smith:
This changes the PK for CPK_FILE and CPK_METADATA.

The tables still have the file_checksum as a unique key and can be continued to be used in queries by runtime-os.

Accompanying runtime-os change is here: https://github.com/corda/corda-runtime-os/pull/1429

**Background of this change:**
While working on [CORE-4868](https://github.com/corda/corda-runtime-os/pull/1385) we decided to add entityVersion to the CPK identifier, but when storing a CPK during force upload, it would result in new row with the same `cpkName`, `cpkVersion`, `cpkSignerSummaryHash` and a different `cpkFileChecksum`, and `entityVersion` would still be 0. This meant the cache in `FlowSandboxServiceImpl` would have multiple CPKs with the same name, version, signerSummaryHash, but different fileChecksums, and an entity version of 0. This meant we couldn't identify a CPK using the CpkIdentifer so the cache would have the incorrect CPKs.

Now when a CPK is force uploaded, we'll merge the new CPK with the existing, hibernate will increment the `entity_version` on the existing CPK and the CPK_File will be overwritten with the new data, thus overwriting the existing CPK.

PR #420, opened 2022-06-15
NOTICK - Revert "CORE-4868 - change PK of cpk_metadata to be the name, version and signer hash"

driessamyn:
Reverts corda/corda-api#400 due to version number clash.


conal-smith (approved):
lgtm

PR #427, opened 2022-06-17
CORE-5102: Add base `Flow` interface and a generic marshalling service

JamesHR3:
See #419 for details. This is the same but with `getRequestBodyAsList` and a corrected API version number.


mnesbit (approved):
OK

blsemo (approved):

PR #428, opened 2022-06-17
CORE-4855 Defined new exceptions and deleted old one

alexey-kadyrov-r3:
https://r3-cev.atlassian.net/browse/CORE-4855

Minimised number of exceptions and changed the approach to use the existing unchecked exception where appropriate and use the specific Crypto Library exceptions only to convey additional context about them


cipher-suite/src/main/kotlin/net/corda/v5/cipher/suite/CryptoService.kt
```diff
@@ -16,20 +15,25 @@ import net.corda.v5.crypto.exceptions.CryptoServiceException
  * Exception handling.
  *
  * As the service instances are decorated with decorators taking care of the timeout, throttling handling,
- * and exception normalisation the exceptions listed for each method are as expected to be thrown by the decorators.
+ * and exception normalisation the exceptions listed for each method bellow are as expected to be thrown by the decorators.
```
mnesbit: below

mnesbit (dismissed):
OK

charlieR3 (approved):

JamesHR3 (approved):

PR #430, opened 2022-06-20
CORE-4423: Add avro and db property for vnode state

benmcmahon100:


data/avro-schema/src/main/resources/avro/net/corda/data/virtualnode/VirtualNodeInfo.avsc
```diff
@@ -47,6 +47,11 @@
         "logicalType": "timestamp-millis"
       },
       "doc": "Time ([Instant]) in milliseconds when the record was updated or added."
+    },
+    {
+      "name": "virtualNodeState",
+      "type": ["string"],
```
driessamyn: Should we use an enum for this?

benmcmahon100: Hey so that was something I discussed with Viktor, and I think we decided to implement the ENUM in the runtime side of things for validation but not data storage. The main reasoning there IIRC was that there are only some valid state transitions. The other idea was that as a string, you can still implement an `enum.from("string")`, but I don't mind changing the type either.

Even if it is an enum - would it not still make sense to transfer the value as a string over the wire? (Maybe that's how it sends the enum under the hood)

driessamyn: The AVRO schema doesn't define storage but communications (I appreciate, an in a compacted topic, they persist).
This defines the API, so I think it's important to signal what the accepted values are. I think we use enum elsewhere in the AVRO messages in cases like this (e.g. `FlowStatus`).

vkolomeyko: I felt that it would be safer to start with `string`, as at this time we are not even sure which states we are going to have. Also, it is quite likely that we might need to add more states in the near future.
If we go with `enum` right from the start then there is a high chance that we might need to add more values soon, which would mean that we will have to make changes to `corda-api` and release a new version, which is always a hassle.

My advise would be to start with `string` and switch to `enum` when this functionality matures. This is an internal cluster API anyway and end users should not be affected.

data/db-schema/src/main/resources/net/corda/db/schema/config/migration/config-creation-v1.0.xml
```diff
@@ -193,6 +193,9 @@
             <column name="hsm_connection_id" type="UUID">
                 <constraints nullable="true"/>
             </column>
+            <column name="virtual_node_state" type="string">
```
vkolomeyko: @benmcmahon100 and I discussed and we agreed to move this new column into `vnode_instance` table.

driessamyn: FYI, not impacting this PR, but for transparency.
The structure of the holding_identity and vnode_instance tables may change. They were designed like this as I was working under the assumption that we may have multiple versions of a vnode at once, in a rolling upgrade scenario for example.
After subsequent discussions with Matthew, it appears this was a false assumption on my side and that only 1 version of a vnode should exist at any time. This means that we can probably merge these 2 tables, however, I wanted to wait until we have a clearer design for Virtual Node upgrade before making this change to make sure we are comfortable with any new structure.

data/avro-schema/src/main/resources/avro/net/corda/data/virtualnode/VirtualNodeUpdateRequest.avsc
```diff
@@ -0,0 +1,16 @@
...
+  "fields": [
+    {
+      "name": "x500Name",
+      "type": "string"
+    },
+    {
+      "name": "updateActor",
+      "type": "string",
+      "doc": "ID of RPC user that requested the virtual node creation."
+    }
```
driessamyn, line 14: You will need version as well in the request as you need to do a version match with the existing, for concurrency control.

Have a look at Configuration (https://github.com/corda/corda-api/blob/release/os/5.0/data/avro-schema/src/main/resources/avro/net/corda/data/config/ConfigurationManagementRequest.avsc) as an example.

data/avro-schema/src/main/resources/avro/net/corda/data/virtualnode/VirtualNodeUpdateResponse.avsc
```diff
@@ -0,0 +1,101 @@
+{
```
driessamyn, line 1: I think the response for create or update should be identical.

data/avro-schema/src/main/resources/avro/net/corda/data/virtualnode/VirtualNodeUpdateRequest.avsc
```diff
@@ -0,0 +1,16 @@
+{
+  "type": "record",
+  "name": "VirtualNodeCreationRequest",
+  "namespace": "net.corda.data.virtualnode",
+  "fields": [
+    {
+      "name": "x500Name",
```
driessamyn, line 7: X500 name isn't enough to identify a VNode. I'm guessing you use the key to identify the vnode using the holding id short hash, but in that case, there's no point in having X500 here.
This would suggest that X500 can be changed, which is untrue.

data/avro-schema/src/main/resources/avro/net/corda/data/virtualnode/VirtualNodeUpdateRequest.avsc
```diff
@@ -0,0 +1,16 @@
+{
+  "type": "record",
+  "name": "VirtualNodeCreationRequest",
```
vkolomeyko, line 3:
```suggestion
  "name": "VirtualNodeUpdateRequest",
```
might be a copy/paste error.

PR #431, opened 2022-06-20
CORE-5195 add flow config default

LWogan:
Add default flow config for root property


gradle.properties
```diff
@@ -7,7 +7,7 @@ cordaProductVersion = 5.0.0
 # NOTE: update this each time this module contains a breaking change
 ## NOTE: currently this is a top level revision, so all API versions will line up, but this could be moved to
 ##   a per module property in which case module versions can change independently.
-cordaApiRevision = 121
+cordaApiRevision = 122
```
JamesHR3: Do you need this? I don't think it's likely this will break anything currently in runtime-os? It would be preferable if you didn't have to update this just to help prevent clashes (you're already behind at least one version number).

LWogan: good point, left it as 122 after the merge

JamesHR3 (approved):

PR #432, opened 2022-06-20
CORE-4996 Rename ledger folders

vlajos:
Rename consensual and common ledger folders to corda-ledger-consensual and corda-ledger-common.


ledger/common/build.gradle
```diff
@@ -26,3 +26,6 @@ dependencies {
     testImplementation "org.assertj:assertj-core:$assertjVersion"
 }
 
+jar {
+    baseName 'corda-ledger-common'
```
ronanbrowne: baseName is [Deprecated ](https://docs.gradle.org/current/dsl/org.gradle.api.tasks.bundling.Jar.html#org.gradle.api.tasks.bundling.Jar:baseName) in favor of archiveBaseName .

ledger/consensual/build.gradle
```diff
@@ -27,3 +27,6 @@ dependencies {
     testImplementation "org.assertj:assertj-core:$assertjVersion"
 }
 
+jar {
+    baseName 'corda-ledger-consensual'
```
ronanbrowne: baseName is [Deprecated ](https://docs.gradle.org/current/dsl/org.gradle.api.tasks.bundling.Jar.html#org.gradle.api.tasks.bundling.Jar:baseName)in favor of archiveBaseName .   


Also, im not close to this work so maybe this has been discussed elsewhere, but the Jar task is specifically set here [corda-api.common-library ](https://github.com/corda/corda-api/blob/8cbf833a7f2c15b9e23767fbf38d56cf510212ba/buildSrc/src/main/groovy/corda-api.common-library.gradle#L31) which this modules and the other uses, rather than fighting the expected naming convention and overriding it later on this level, couldn't this module have been named "ledger-consensual" and then this workaround not be needed?

ronanbrowne (changes requested):

ledger/common/build.gradle
chrisr3: Also, we want to configure tasks lazily, so:
```gradle
tasks.named('jar', Jar) {
    archiveBaseName = 'corda-ledger-common'`
   ...
}
```

vlajos: Thanks. I have changed them.

ledger/consensual/build.gradle
blsemo: I'm not a fan of repeating redundant information in the path, but if that makes the gradle magic simpler, I'm ok with it.

blsemo (approved):
LGTM

ronanbrowne (approved):
LGTM

PR #435, opened 2022-06-21
CORE-5260 new Merkle tree 

vlajos:
New Merkle tree API.
Implementation: https://github.com/corda/corda-runtime-os/pull/1492

### DIFF
--- a/crypto/src/main/kotlin/net/corda/v5/crypto/merkle/MerkleTree.kt
+++ b/crypto/src/main/kotlin/net/corda/v5/crypto/merkle/MerkleTree.kt
@@ -1,0 +1,10 @@
+package net.corda.v5.crypto.merkle
+
+import net.corda.v5.crypto.SecureHash
+
+interface MerkleTree {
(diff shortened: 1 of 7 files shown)


blsemo (approved):
LGTM

crypto/src/main/kotlin/net/corda/v5/crypto/merkle/MerkleTree.kt
```diff
@@ -0,0 +1,10 @@
+package net.corda.v5.crypto.merkle
+
+import net.corda.v5.crypto.SecureHash
+
+interface MerkleTree {
+    val leaves: List<ByteArray>
+    val digestProvider: MerkleTreeHashDigestProvider
```
alexey-kadyrov-r3, line 7: You are making the digest provider as part of the tree structure and mixing it with the properties related to the tree itself

alexey-kadyrov-r3 (approved):
LGTM

PR #436, opened 2022-06-21
CORE-4420 - Update group policy entry to contain the full holding identity

dimosr:
The stubs the p2p components had for the group policy provider were keyed only by group ID, but they need to be keyed by the full holding identity that performs the lookup. This change adjusts the schemas in this direction.

Corresponding corda-runtime-os PR: https://github.com/corda/corda-runtime-os/pull/1508


data/avro-schema/src/main/resources/avro/net/corda/data/p2p/LinkOutHeader.avsc
```diff
@@ -3,6 +3,11 @@
   "name": "LinkOutHeader",
   "namespace": "net.corda.p2p",
   "fields": [
+    {
+      "doc": "The source identity (x500 name + group ID)",
```
williamvigorr3: Slightly nit picky: Can we order this `destination` first then `source` to be the same as `AuthenticatedMessageHeader`.

dimosr: sure

williamvigorr3 (approved):
LGTM

PR #442, opened 2022-06-23
CORE-3085: Handle approved member request

yift-r3:
The runtime PR is in https://github.com/corda/corda-runtime-os/pull/1582

The PR contains:
1. Adding the approved and approving member to the approved message
2. Add the registration ID to the approved message
3. Adds Query members signatures  API.
4. Adds Approved request and member API.
5. Rename `QueryFailedResponse` to `PersistenceFailedResponse`
6. Add member signature table.


data/avro-schema/src/main/resources/avro/net/corda/data/membership/db/response/query/UpdateMemberAndRequestResponse.avsc
```diff
@@ -0,0 +1,13 @@
+{
+  "type": "record",
+  "name": "UpdateMemberAndRequestResponse",
```
charlieR3:
```suggestion
  "name": "UpdateMemberAndRegistrationRequestResponse",
```

data/avro-schema/src/main/resources/avro/net/corda/data/membership/db/response/query/MemberSignature.avsc
```diff
@@ -0,0 +1,18 @@
+{
+  "type": "record",
+  "name": "MemberSignature",
```
charlieR3, line 3: We may run in to issues here during reregistration where a member changes their member context and signature. They may not be approved immediately so we might need multiple records per member. Some of the details still need to be figured out.

charlieR3: One option could be to include the serial number that the MGM sets on a member and increments each time their info changes. Currently we just set it and leave it unchanged but on reregistration that will be incremented.

yift-r3: Would we ever want to use historical member signatures?

charlieR3: I'm just thinking of the scenario down the line with reregistration when alice could submit a new member context which the MGM needs to approve. But before it's approved, the member list is sent to bob. If we persist the member signature for the new member context it will not match the approved version of alice's member context. We will have a similar issue for the member info though and how to persist active and pending versions of a member info if we choose to allow both states at once. This is probably something that we will just need to think about when we get to reregistration.

charlieR3 (dismissed):
LGTM

charlieR3 (approved):

PR #445, opened 2022-06-23
CORE-5105 Crypto + Flow integration - Signing

lankydan:
Add a property to the `Checkpoint` schema to keep track of what the current crypto request is.

Temporarily add `decodePublicKey` to the public API to assist in testing until MGM code catches up.

Update `FlowOpsResponse` schema to include a schema that was not referenced but was used inside the `corda-runtime-os` code.

### DIFF
--- a/application/src/main/kotlin/net/corda/v5/application/crypto/SigningService.kt
+++ b/application/src/main/kotlin/net/corda/v5/application/crypto/SigningService.kt
@@ -5,7 +5,6 @@ import net.corda.v5.base.annotations.Suspendable
 import net.corda.v5.crypto.CompositeKey
 import net.corda.v5.crypto.DigitalSignature
 import net.corda.v5.crypto.SignatureSpec
-import net.corda.v5.crypto.exceptions.CryptoServiceLibraryException
 import java.security.KeyPair
 import java.security.PrivateKey
 import java.security.PublicKey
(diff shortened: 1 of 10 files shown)


alexey-kadyrov-r3 (dismissed):
LGTM

application/src/main/kotlin/net/corda/v5/application/crypto/SigningService.kt
```diff
@@ -35,4 +35,15 @@ interface SigningService {
      */
     @Suspendable
     fun sign(bytes: ByteArray, publicKey: PublicKey, signatureSpec: SignatureSpec): DigitalSignature.WithKey
+
+    /**
+    * TODO Remove this API from C5. Added as a temporary API for testing until MGM code is fully integrated.
```
JamesHR3, line 40: I wonder if it's worth having this API anyway to make the process of passing a public key in from RPC better? It's probably a question that needs wider discussion with @rick-r3 and @alexey-kadyrov-r3, as it depends quite a lot on what we're expecting users to do with crypto.

data/avro-schema/src/main/resources/avro/net/corda/data/flow/event/FlowEvent.avsc
```diff
@@ -14,7 +14,8 @@
         "net.corda.data.flow.event.Wakeup",
         "net.corda.data.flow.output.FlowStatus",
         "net.corda.data.flow.event.SessionEvent",
-        "net.corda.data.persistence.EntityResponse"
+        "net.corda.data.persistence.EntityResponse",
+        "net.corda.data.crypto.wire.ops.flow.FlowOpsResponse"
```
JamesHR3, line 18: I'm sure it's on your radar already but this is the sort of thing where I'm hoping the unification work can help us out, by reducing the need to add something here for every subsystem we need to call out to.

JamesHR3 (dismissed):
There's clearly a few places in this PR where we can simplify things a bit as we unify our managers for calling out to other parts of the system, but this seems reasonable for the first stab at the crypto stuff.

JamesHR3 (approved):

PR #446, opened 2022-06-24
CORE-5333 remove logic regarding excluding cpk specific modules in gradle caching

ronanbrowne:
Initially we added this check as we suspected this was causing very sporadic issues with some tests related to CPKs , this was never proved to be the case 100%, revesting this area currently and feeling is this should be reactivated and we can monitor as needed.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -206,18 +206,13 @@ subprojects {
         }
     }
 
-    // NOTE: Needed to ensure we do not normalize projects which apply CPK plugin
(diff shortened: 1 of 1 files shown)


chrisr3 (approved):
This Gradle incantation may bring about the :fire: _**APOCALYPSE**_ :fire: for all I know. However, at least you've removed an `afterEvaluate` before the end :+1:.

benmcmahon100 (approved):

PR #448, opened 2022-06-27
CORE-5247 Fixes for issues related to running combined worker

mbrkic-r3:
Fixes for issues related to running combined worker:

- Fixed boot DB parameters
- Made property "schema.name" local

### DIFF
--- a/data/config-schema/src/main/kotlin/net/corda/schema/configuration/BootConfig.kt
+++ b/data/config-schema/src/main/kotlin/net/corda/schema/configuration/BootConfig.kt
@@ -13,9 +13,9 @@ object BootConfig {
     const val BOOT_DB = "db"
 
     const val BOOT_DB_PARAMS = "$BOOT_DB.params"
-    const val BOOT_JDBC_URL = "$BOOT_DB.jdbcUrl"
-    const val BOOT_JDBC_USER = "$BOOT_DB.user"
-    const val BOOT_JDBC_PASS = "$BOOT_DB.pass"
+    const val BOOT_JDBC_URL = "$BOOT_DB_PARAMS.database.jdbc.url"
+    const val BOOT_JDBC_USER = "$BOOT_DB_PARAMS.database.user"
+    const val BOOT_JDBC_PASS = "$BOOT_DB_PARAMS.database.pass"
 
     const val BOOT_DIR = "dir"
     const val BOOT_WORKSPACE_DIR = "$BOOT_DIR.workspace"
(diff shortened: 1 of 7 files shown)


data/db-schema/src/main/resources/net/corda/db/schema/crypto/migration/crypto-creation-v1.0.xml
```diff
@@ -2,8 +2,15 @@
...
                    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
                    xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-4.3.xsd">
+    <property name="schema.name" value="CRYPTO" global="false"/>
+
     <changeSet author="R3.Corda" id="crypto-creation-v1.0">
-        <createTable tableName="crypto_wrapping_key">
+
+        <sql>
+            CREATE SCHEMA IF NOT EXISTS CRYPTO;
+        </sql>
```
driessamyn: Ideally this shouldn't be in here as the schema should be created outside of this. I thought the cluster bootstrapper already did that?
We can leave this in, it'll just mean we have to take it out later on.

data/db-schema/src/main/resources/net/corda/db/schema/crypto/migration/crypto-creation-v1.0.xml
```diff
@@ -22,9 +29,10 @@
         </createTable>
         <addPrimaryKey  columnNames="alias"
                         tableName="crypto_wrapping_key"
-                        constraintName="crypto_wrapping_key_pk"/>
+                        constraintName="crypto_wrapping_key_pk"
+                        schemaName="${schema.name}"/>
 
-        <createTable tableName="crypto_hsm_config">
+        <createTable tableName="crypto_hsm_config" schemaName="${schema.name}">
```
driessamyn: Similar to the above, schemas were left out as they should be passed into the connection string to this DB.

data/db-schema/src/main/resources/net/corda/db/schema/crypto/migration/crypto-creation-v1.0.xml
alexey-kadyrov-r3: it has to be deleted

alexey-kadyrov-r3: That script is only part of the migration, there is another script which runs for the cluster and for a vnode (vnode-crypto directory).

driessamyn (dismissed):

driessamyn (approved):
LGTM

driessamyn (approved):
LGTM

PR #451, opened 2022-06-28
CORE-5347: Split vnode schemas into envelope and requests

benmcmahon100:
corda/corda-runtime-os#1565


data/avro-schema/src/main/resources/avro/net/corda/data/virtualnode/VirtualNodeRequest.avsc
```diff
@@ -0,0 +1,24 @@
...
+        "logicalType": "timestamp-millis"
+      },
+      "doc": "Time ({@link java.time.Instant}) in milliseconds when the record was created."
+    },
+    {
+      "name": "request",
+      "doc": "The 'request' that we wish to make to the persistence API.",
+      "type": [
+        "net.corda.data.virtualnode.VirtualNodeCreate",
+        "net.corda.data.virtualnode.VirtualNodeInfo"
```
vkolomeyko: I kind of understood why we would include `VirtualNodeInfo` into the response.
But here when we are including `VirtualNodeInfo` into the request it is not immediately obvious to me what are we requesting.
I would suggest we remove `VirtualNodeInfo` from request. I think we agreed to later on add `VirtualNodeStateUpdate` into the request as a separate PR.

benmcmahon100: Yes Update was for later - VirtualNodeInfo was added to the request as something you'd like to ask for, unless I've understood that wrong?

sprintengine: Yes I'd say the wrapper should be called `VirtualNodeManagementRequest` and the inner objects `VirtualNodeCreateRequest` (and later `VirtualNodeUpdateRequest`) as we should explicitly make a separation in the names of the request and response objects.

Also `VirtualNodeManagementResponse`, `VirtualNodeCreateResponse` (and later `VirtualNodeUpdateResponse`)

sprintengine: We already have `VirtualNodeInfoProcessor` which gets vnode info from Kafka and loads into a cache, so we don't need a separate request making round trips to the database to get vnode info, afaik.

`VirtualNodeReconciler` will detect your change to the state and send update to kafka which will get picked up by VirtualNodeInfoProcessors.

vkolomeyko: I agree with @conalsmith-r3 except that I would not use `VirtualNodeUpdateRequest` as it implies that anything on vNode can be updated. For our case (in the separate PR) I would call it `VirtualNodeStateChangeRequest`.

benmcmahon100: Ah cool okay then if this is only for writeable operations then I'm totally on board with calling it `management`

sprintengine: Yeah a `VirtualNodeStateChangeRequest` is good.

vkolomeyko:
Looks like `corda-api` has already been changed, I think we need to merge changes from the `main` branch.
And since this is a breaking change we need to increment a version.

vkolomeyko (commented):
The changes look sensible to me - let's accommodate them in `corda-runtime-os`.

@benmcmahon100 - once this is done against alpha version, could you reference `corda-runtime-os` PR from this PR?

data/avro-schema/src/main/resources/avro/net/corda/data/virtualnode/VirtualNodeManagementResponseSuccess.avsc
```diff
@@ -0,0 +1,15 @@
...
+  "type": "record",
+  "name": "VirtualNodeManagementResponseSuccess",
+  "doc": "An {@link VirtualNodeManagementRequest} has been successful.  A return type of `null` can be considered success and/or no results.  See also:  {@link VirtualNodeManagementResponseFailure} and {@link VirtualNodeManagementResponseSuccess}",
+  "namespace": "net.corda.data.virtualnode",
+  "fields": [
+    {
+      "name": "result",
+      "doc": "Result of request, either null (success or no results), or {@link VirtualNodeInfo}",
+      "type": [
+        "net.corda.data.virtualnode.VirtualNodeInfo"
```
sprintengine: IMO I think we should use the original CreateVirtualNodeResponse here

vkolomeyko: As discussed on Slack, let's stick with `CreateVirtualNodeResponse`.

data/avro-schema/src/main/resources/avro/net/corda/data/virtualnode/VirtualNodeManagementResponse.avsc
```diff
@@ -0,0 +1,24 @@
...
+        "logicalType": "timestamp-millis"
+      },
+      "doc": "Time ({@link java.time.Instant}) in milliseconds when the record was created."
+    },
+    {
+      "name": "responseType",
+      "doc": "Result of request, either a success or failure message",
+      "type": [
+        "net.corda.data.virtualnode.VirtualNodeManagementResponseSuccess",
+        "net.corda.data.virtualnode.VirtualNodeManagementResponseFailure"
```
sprintengine, line 20: I think we can remove the extra wrapper `VirtualNodeManagementResponseSuccess`, and our management response just contain either a failure or the operation's response object, i.e. change this `responseType` to:

```
  {
      "name": "response",
      "type": [
        "net.corda.data.virtualnode.VirtualNodeCreateResponse",
        "net.corda.data.virtualnode.VirtualNodeManagementResponseFailure"
       ]
```
WDYT?

The RPCSender code can just check if the type of response is a failure then handle failure, otherwise it was a success. Might just simplify the model.

benmcmahon100: I could do that - but that means you wouldn't have the wrapper in place for when update becomes a thing.

IMO I like the wrapper for the sake of readability as on the sending end of the request - you can parse the response as 
```kotlin
when (result) {
  success -> ..
  fail -> ..
}
```

data/avro-schema/src/main/resources/avro/net/corda/data/virtualnode/VirtualNodeManagementResponseSuccess.avsc
```diff
@@ -0,0 +1,16 @@
+{
+  "type": "record",
+  "name": "VirtualNodeManagementResponseSuccess",
+  "doc": "An {@link VirtualNodeManagementRequest} has been successful.  A return type of `null` can be considered success and/or no results.  See also:  {@link VirtualNodeManagementResponseFailure} and {@link VirtualNodeManagementResponseSuccess}",
+  "namespace": "net.corda.data.virtualnode",
+  "fields": [
+    {
+      "name": "result",
+      "doc": "Result of request, either null (success or no results), or {@link VirtualNodeInfo}",
```
sprintengine: We won't be allowing a null response, so can update this doc.

data/avro-schema/src/main/resources/avro/net/corda/data/virtualnode/VirtualNodeManagementResponse.avsc
benmcmahon100: Talked it through IRL and have done a 180 on my opinion - happy to change around

sprintengine: I talked myself around to your idea I see its the pattern used in Entity stuff, so I don't mind either way.

edit - after https://github.com/corda/corda-api/pull/451/files#r910966824 we can replace `VirtualNodeManagementResponseFailure` here with `ExceptionEnvelope`

data/avro-schema/src/main/resources/avro/net/corda/data/virtualnode/VirtualNodeManagementResponseFailure.avsc
```diff
@@ -0,0 +1,28 @@
...
+      "doc": "Error type",
+      "type": {
+        "name": "Error",
+        "type": "enum",
+        "doc": "",
+        "symbols": [
+          "FATAL",
+          "NOT_READY",
+          "DATABASE",
+          "VIRTUAL_NODE"
```
sprintengine: I'm not really sure what the purpose of this is. If it's to drive HTTP exception mapping at the RPC sender side, then we'd need to decide what responses to return and in what scenarios. I don't think thats in the scope of CORE-5347. These are the error types used by `EntityResponseWaitingForHandler` to drive decisions about flows, but I'm not sure they apply here. I'd be inclined to get rid of it for now and just include the `ExceptionEnvelope` in the union in `VirtualNodeManagementResponse`. WDYT @vkolomeyko ?

vkolomeyko: Yes, I think it would make sense to remove this enum from `VirtualNodeManagementResponseFailure` and keep `ExceptionEnvelope` only.

benmcmahon100:
@conalsmith-r3 @vkolomeyko just reading back here as I've finally overcome all my tech difficulties. It seems we both like and dislike the fail success response. I think I should take Viktors feedback onboard instead and just drop the failure enum for now. We can reintroduce that additional envelope later on if it's appropriate

data/avro-schema/src/main/resources/avro/net/corda/data/virtualnode/VirtualNodeManagementResponseFailure.avsc
vkolomeyko: @benmcmahon100 - I though we wanted to remove `type` enum and keep `ExceptionEnvelope` only?

data/avro-schema/src/main/resources/avro/net/corda/data/virtualnode/VirtualNodeManagementResponse.avsc
```diff
@@ -0,0 +1,25 @@
...
+        "type": "long",
+        "logicalType": "timestamp-millis"
+      },
+      "doc": "Time ({@link java.time.Instant}) in milliseconds when the record was created."
+    },
+    {
+      "name": "responseType",
+      "doc": "Result of request, either a success or failure message",
+      "type": [
+        "net.corda.data.virtualnode.VirtualNodeInfo",
```
vkolomeyko: Do we need to include `VirtualNodeInfo` as a possible option for reply? I would much rather stick with `VirtualNodeCreateResponse` and `VirtualNodeManagementResponseFailure` only for now.

benmcmahon100:
https://github.com/corda/corda-runtime-os/pull/1565

vkolomeyko (dismissed):
`corda-api` changes look fine to me.
But before merging, can we please get `corda-runtime-os` PR into a green state, such that we could merge both PRs in close succession.

data/avro-schema/src/main/resources/avro/net/corda/data/virtualnode/VirtualNodeManagementResponseFailure.avsc
```diff
@@ -0,0 +1,13 @@
+{
+  "type": "record",
+  "name": "VirtualNodeManagementResponseFailure",
+  "doc": "The response when an {@link VirtualNodeManagementRequest} has failed. The failure type may be further categorised by the {@link Error} enumeration. See also: {@link VirtualNodeManagementResponseSuccess}",
```
sprintengine: Think we can change this doc as we don't have the Error enum or success response anymore

data/avro-schema/src/main/resources/avro/net/corda/data/virtualnode/VirtualNodeManagementResponse.avsc
```diff
@@ -0,0 +1,24 @@
+{
+  "type": "record",
+  "name": "VirtualNodeManagementResponse",
+  "doc": "The response to an {@link VirtualNodeRequest}.  On success the {@link VirtualNodeResponse#responseType} is {@link VirtualNodeManagementResponseSuccess}, on failure it is {@link VirtualNodeManagementResponseFailure}",
```
sprintengine: We can update this doc too

conal-smith (dismissed):
lgtm

conal-smith (approved):
lgtm

driessamyn (approved):
LGTM

PR #452, opened 2022-06-28
CORE-5145 Merkle Tree Factory API

vlajos:
Merkle Tree Factory API
Implementation: https://github.com/corda/corda-runtime-os/pull/1550


crypto/src/main/kotlin/net/corda/v5/crypto/merkle/MerkleService.kt
```diff
@@ -0,0 +1,23 @@
+package net.corda.v5.crypto.merkle
+
+import net.corda.v5.crypto.DigestAlgorithmName
+import net.corda.v5.crypto.SecureHash
+
+interface MerkleService {
```
blsemo: Maybe rather call it `MerkleTreeFactory`?

vlajos: OK. I have renamed it.

crypto/src/main/kotlin/net/corda/v5/crypto/merkle/MerkleTreeFactory.kt
```diff
@@ -0,0 +1,55 @@
...
+     * existing [MerkleTree].
+     * Use this function to recreate the [MerkleProof] when the [MerkleTree] is not available.
+     *
+     * @param treeSize Number of leaves in the whole tree
+     * @param leaves Leaf items whose inclusion is proved by the proof.
+     * @param hashes The helper hashes needed to reconstruct the whole tree.
+     *
+     * @return A new [MerkleProof] instance.
+     */
+    fun createProof(
```
blsemo: What is the use case for this function? Do we need this on the public API?

vlajos: Hmm. Good point.
I think deserialization will be the way how we will recreate proofs on the verifier sides.
I have removed this from here.

vlajos: Good point.
I think deserialization will be the way how we will recreate proofs on the verifier side.
I have removed it.

crypto/src/main/kotlin/net/corda/v5/crypto/merkle/MerkleTreeFactory.kt
```diff
@@ -0,0 +1,36 @@
...
+     * @param merkleTreeHashDigestProviderName name of the Hash Digest Provider class
+     * @param digestAlgorithmName name of the base Hash algorithm
+     * @param options Hash Digest provider specific options
+     *
+     * @return A new [MerkleTreeHashDigestProvider] instance.
+     */
+    fun createHashDigestProvider(
+        merkleTreeHashDigestProviderName: String,
+        digestAlgorithmName: DigestAlgorithmName,
+        options: HashMap<String, Any>? = null,
```
mnesbit: Should be Map not HashMap

mnesbit: and I would prefer it to just be empty for hashes that don't need it rather than null

crypto/src/main/kotlin/net/corda/v5/crypto/merkle/MerkleTreeFactory.kt
```diff
@@ -0,0 +1,36 @@
+package net.corda.v5.crypto.merkle
+
+import net.corda.v5.crypto.DigestAlgorithmName
+
+/**
+ * [MerkleTreeFactory] creates [MerkleTree]s, [MerkleProof]s and [MerkleTreeHashDigestProvider].
+ */
+interface MerkleTreeFactory {
```
mnesbit, line 61: I think we should add some constants to this for the various 'standard' digest algorithms and also for the parameter keys that they require

mnesbit: I think this will need to be injectable into flows, so we will need to put some annotations on and separately wire the implementation class into the flow inject. That may also mean we need to put 'Service' in the name for clarity, but we should see what we do in the other crypto factories for consistency

omfR3: I think this was initially `MerkleService` or something. I've added `@DoNotImplement` to the interface and `@Suspendable` to the functions -- is that what you mean?
The implementation class in corda-runtime-os is wired up with `@Component` and whatnot.

Can you elaborate about the constants?

vlajos: @mnesbit 
> I think we should add some constants to this for the various 'standard' digest algorithms and also for the parameter keys that they require

I have pushed some 1-1 changes and introduced MerkleTreeHashDigestProviderOption and MerkleTreeHashDigestProviderName
Is this what you meant?

vlajos: > That may also mean we need to put 'Service' in the name for clarity, but we should see what we do in the other crypto factories for consistency

`CipherSchemeMetadataImpl` and `CryptoFlowOpsTransformerFactoryImpl` do not have Service in their names.
I do not see a strict pattern, but happy to rename.

crypto/src/main/kotlin/net/corda/v5/crypto/merkle/MerkleTreeHashDigestProviderOption.kt
```diff
@@ -0,0 +1,37 @@
...
+class MerkleTreeHashDigestProviderOption(val name: String) {
+    init {
+        require(name.isNotBlank()) { "Hash digest provider option key name unavailable or not specified" }
+    }
+
+    companion object {
+        @JvmField
+        val LEAF_PREFIX: MerkleTreeHashDigestProviderOption = MerkleTreeHashDigestProviderOption("leafPrefix")
+
+        @JvmField
```
mnesbit: typo PREFIX

crypto/src/main/kotlin/net/corda/v5/crypto/merkle/MerkleTreeHashDigestProviderName.kt
```diff
@@ -0,0 +1,43 @@
...
+        @JvmField
+        val NONCE_SIZE_ONLY_VERIFY: MerkleTreeHashDigestProviderName = MerkleTreeHashDigestProviderName("NonceHashDigestProviderSizeOnlyVerify")
+
+        @JvmField
+        val TWEAKABLE: MerkleTreeHashDigestProviderName = MerkleTreeHashDigestProviderName("TweakableHashDigestProvider")
+    }
+
+    override fun toString(): String = name
+
+    override fun hashCode(): Int = name.uppercase().hashCode()
```
mnesbit: This hash code and the one on MerkleTreeHashDigestProviderOption are not right! If you uppercase the hash calculation you should also uppercase the equals comparison, or else don't make it case insensitive.

crypto/src/main/kotlin/net/corda/v5/crypto/merkle/MerkleTreeFactory.kt
```diff
@@ -0,0 +1,41 @@
...
+     * @param merkleTreeHashDigestProviderName name of the Hash Digest Provider class
+     * @param digestAlgorithmName name of the base Hash algorithm
+     * @param options Hash Digest provider specific options
+     *
+     * @return A new [MerkleTreeHashDigestProvider] instance.
+     */
+    @Suspendable
+    fun createHashDigestProvider(
+        merkleTreeHashDigestProviderName: MerkleTreeHashDigestProviderName,
+        digestAlgorithmName: DigestAlgorithmName,
```
mnesbit, line 88: I am not sure the DigestAlgorithmName, or MerkleTreeHashDigestProviderOption are really adding much over a string. What is really needed is an explanation of what the different default digest providers do and what parameters they require. This would be best of the platform defined constants were grouped next to each other rather than in separate files and it does need considerably more text.

mnesbit: If we do go with MerkleTreeHashDigestProviderOption as a distinct class it is of value to include a property to tell you the expected value type that would go into the options map

crypto/src/main/kotlin/net/corda/v5/crypto/merkle/MerkleTreeHashDigestProviderName.kt
```diff
@@ -0,0 +1,43 @@
...
+ */
+@CordaSerializable
+class MerkleTreeHashDigestProviderName(val name: String) {
+    init {
+        require(name.isNotBlank()) { "Hash digest provider name unavailable or not specified" }
+    }
+
+    companion object {
+        @JvmField
+        val DEFAULT: MerkleTreeHashDigestProviderName = MerkleTreeHashDigestProviderName("DefaultHashDigestProvider")
```
mnesbit: needs text to say what algorithm they are. ie.
DEFAULT - An implementation of rfc6962 compatible merkle tree
NONCE - A merkle tree digest that one-way derives a set of per-leaf nonces from a 16-byte entropy input. The nonces are pre-hashed with the leaf data before forming a standard merkle tree. This allows low entropy inputs to be used in the merkle tree.
NONCE_VERIFY - When verifying a NONCE digest based MerkleProof the source entropy is not communicated to the verifier. Instead only the per-leaf nonces are transmitted. Use the NONCE_VERIFY digest to validate the MerkleProof against the expected root hash.
NONCE_SIZE_ONLY_VERIFY - Using the NONCE digest it is possible to create a MerkleProof that doesn't reveal the leaf data, but provides supporting evidence of the number of leaves. These MerkleProofs can be validated against and expected root hash using the NONCE_SIZE_ONLY_VERIFY digest
TWEAKABLE - This provides an extension of the DEFAULT hash with configurable hashes for leaf and internal Merkle node operations. This is required when similar Merkle trees are used to generate a hash that is signed over so that replay attacks can be prevented.

crypto/src/main/kotlin/net/corda/v5/crypto/merkle/MerkleTreeHashDigestProviderName.kt
vlajos: `ignoreCase` has been turned on in the `equals` :

`return name.equals(otherName, true)`

Or do I miss something?

omfR3: Ah, I didn't know Kotlin had a different style than Java there -- I would have expected something like `name.equalsIgnoreCase(otherName)`. That's a strange design choice in Kotlin...

crypto/src/main/kotlin/net/corda/v5/crypto/merkle/MerkleTreeFactory.kt
vlajos: I have reverted these and introduced some simpler string constants instead.

crypto/src/main/kotlin/net/corda/v5/crypto/merkle/MerkleTreeHashDigestProviderName.kt
vlajos: Thanks. I have added these explanations.

crypto/src/main/kotlin/net/corda/v5/crypto/merkle/MerkleTreeFactory.kt
```diff
@@ -0,0 +1,91 @@
...
+ * Use the NONCE_VERIFY digest to validate the MerkleProof against the expected root hash.
+ */
+const val HASH_DIGEST_PROVIDER_NONCE_VERIFY_NAME = "NonceHashDigestProviderVerify"
+
+/**
+ * Using the NONCE digest it is possible to create a MerkleProof that doesn't reveal the leaf data, but
+ * provides supporting evidence of the number of leaves.
+ * These MerkleProofs can be validated against an expected root hash using the NONCE_SIZE_ONLY_VERIFY digest
+ */
+const val HASH_DIGEST_PROVIDER_NONCE_SIZE_ONLY_VERIFY_NAME = "NonceHashDigestProviderSizeOnlyVerify"
```
mnesbit, line 32: I think we can't do size only proofs via the public api as that used NonceHashDigestProvider's getSizeProof method. Need to think if we can resolve that via a HASH_DIGEST_PROVIDER_NONCE_SIZE_ONLY_NAME digest, or some other mechanism.

mnesbit: I agree we could delete this for now and revisit when we get to notary protocols rather than holding things up

mnesbit (dismissed):
OK I don't want to hold things up.

crypto/src/main/kotlin/net/corda/v5/crypto/merkle/MerkleTreeFactory.kt
vlajos: I have added a simple extension interface for now.

blsemo (approved):
LGTM

PR #455, opened 2022-06-29
NOTICK - Add some docs for Avro schemas that are part of p2p public API

dimosr:
After discussions with various people these days, I realised `docs` fields were missing from some Avro schemas that form the public API of the p2p layer. This adds those, so that people that use the p2p layer can understand the semantics of each field more easily.

### DIFF
--- a/data/avro-schema/src/main/resources/avro/net/corda/data/p2p/app/AuthenticatedMessageHeader.avsc
+++ b/data/avro-schema/src/main/resources/avro/net/corda/data/p2p/app/AuthenticatedMessageHeader.avsc
@@ -14,7 +14,7 @@
       "type": "net.corda.data.identity.HoldingIdentity"
     },
     {
-      "doc": "A (time-to-live) timestamp after which this message will be dropped from the p2p layer. If no ttl is specified, the p2p layer will continue replaying this message infinitely until it is delivered.",
+      "doc": "A (time-to-live) unix timestamp (in milliseconds) after which this message will be dropped from the p2p layer. If no ttl is specified, the p2p layer will continue replaying this message infinitely until it is delivered.",
       "name": "ttl",
       "type": ["null", "long"]
     },
--- a/data/avro-schema/src/main/resources/avro/net/corda/data/p2p/app/AuthenticatedMessageHeader.avsc
+++ b/data/avro-schema/src/main/resources/avro/net/corda/data/p2p/app/AuthenticatedMessageHeader.avsc
@@ -1,31 +1,37 @@
 {
   "type": "record",
   "name": "AuthenticatedMessageHeader",
+  "doc": "A message that will be delivered by the p2p layer over an end-to-end (authenticated and optionally encrypted) session.",
   "namespace": "net.corda.p2p.app",
   "fields": [
     {
+      "doc": "The destination identity for this message.",
       "name": "destination",
       "type": "net.corda.data.identity.HoldingIdentity"
     },
     {
+      "doc": "The source identity of this message.",
       "name": "source",
       "type": "net.corda.data.identity.HoldingIdentity"
     },
     {
+      "doc": "A (time-to-live) timestamp after which this message will be dropped from the p2p layer.",
       "name": "ttl",
       "type": ["null", "long"]
     },
     {
+      "doc": "A unique identifier for this message. This will be used by the p2p layer to track the delivery of this specific message.",
       "name": "messageId",
       "type": "string"
     },
     {
+      "doc": "A trace identifier. The semantics of this field can vary based on the use-case of the upstream user, but it can be used to trace together multiple instances of the same message (e.g. in cases where the upstream user also performs replays, it can assign the same trace ID to all the instances of the same replayed message.)",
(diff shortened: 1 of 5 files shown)


data/avro-schema/src/main/resources/avro/net/corda/data/p2p/app/AuthenticatedMessageHeader.avsc
```diff
@@ -1,31 +1,37 @@
...
       "name": "destination",
       "type": "net.corda.data.identity.HoldingIdentity"
     },
     {
+      "doc": "The source identity of this message.",
       "name": "source",
       "type": "net.corda.data.identity.HoldingIdentity"
     },
     {
+      "doc": "A (time-to-live) timestamp after which this message will be dropped from the p2p layer.",
```
yift-r3: Probably worth adding the unit and describing what null means.

(By the way, can we use [`duration`](https://avro.apache.org/docs/1.11.0/spec.html#Duration)?)

data/avro-schema/src/main/resources/avro/net/corda/data/p2p/app/AuthenticatedMessageHeader.avsc
```diff
@@ -1,31 +1,37 @@
 {
   "type": "record",
   "name": "AuthenticatedMessageHeader",
+  "doc": "A message that will be delivered by the p2p layer over an end-to-end (authenticated and optionally encrypted) session.",
```
yift-r3: We should probably do the same for `UnauthenticatedMessage`, `AuthenticatedMessage` and `AppMessage`.

dimosr: oh my bad, that was supposed to be in the corresponding messages - not the headers. I will move it.

data/avro-schema/src/main/resources/avro/net/corda/data/p2p/app/AuthenticatedMessageHeader.avsc
dimosr: good point, I'll add the unit and semantics of null.

Wrt to duration, we decided to use a UTC timestamp here on purpose to avoid skew effects from different systems having to calculate the expiry time. Note: even the timestamp doesn't completely eliminate issues due to clock skew, but it makes the API a bit easier to reason about.

yift-r3: So, maybe `timestamp-millis`?

dimosr: TTL is an industry term, so I think it's better (as long as we document unit and semantics)

emilybowe (dismissed):
LGTM

data/avro-schema/src/main/resources/avro/net/corda/data/p2p/app/AuthenticatedMessageHeader.avsc
dimosr: [discussed offline with Yiftach] apparently he meant the logical type, not changing the name.

williamvigorr3: This might be relevant: https://issues.apache.org/jira/browse/AVRO-2123

data/avro-schema/src/main/resources/avro/net/corda/data/p2p/app/AuthenticatedMessageHeader.avsc
```diff
@@ -4,28 +4,34 @@
...
+      "logicalType": "timestamp-millis",
       "type": ["null", "long"]
     },
     {
+      "doc": "A unique identifier for this message. This will be used by the p2p layer to track the delivery of this specific message.",
       "name": "messageId",
       "type": "string"
     },
     {
+      "doc": "A trace identifier. The semantics of this field can vary based on the use-case of the upstream user, but it can be used to trace together multiple instances of the same message (e.g. in cases where the upstream user also performs replays, it can assign the same trace ID to all the instances of the same replayed message.)",
```
williamvigorr3: Nit: can you close the bracket before the full stop. Or add a full stop before the opening bracket.

dimosr: yes, that was a typo actually - full stop was meant to be after the closing paren

data/avro-schema/src/main/resources/avro/net/corda/data/p2p/app/AuthenticatedMessageHeader.avsc
dimosr: at quick look, it seems they say that `timestamp-millis` are working fine, it's the other types that have the issue

yift-r3 (approved):
LGTM

PR #458, opened 2022-06-30
CORE-5171: Refactor of Avro checkpoint object

JamesHR3:
This PR splits the checkpoint object stored for flows into two parts - one for storing flow state and one for storing pipeline execution information. By doing this we can ensure the pipeline always has information to work with (e.g. to support retries) even before the flow information is populated. This is useful in cases where a failure occurs before the full information is known (e.g. when trying to identify what initiated flow to execute).


data/avro-schema/src/main/resources/avro/net/corda/data/flow/state/checkpoint/FlowState.avsc
```diff
@@ -0,0 +1,61 @@
...
+  "name": "FlowState",
+  "namespace": "net.corda.data.flow.state.checkpoint",
+  "doc": "State used by the flow engine to track the execution progress of the flow",
+  "fields": [
+    {
+      "name": "fiber",
+      "type": "bytes",
+      "doc": "Flow execution state, serialized at the point the flow was suspended."
+    },
+    {
```
owenstanford: maybe a bit pedantic, but do we want the immutable flow start context mixed with the mutable flow state? should the start context live on the root of the checkpoint?

JamesHR3: Part of the problem is that one failure mode is we couldn't make a valid flow start context. That happens if the initiated flow class doesn't exist (or indeed if RPC sends us something invalid), so I think we want to set up the flow state once we know we can create that object.

lankydan (dismissed):
Changes are good, approving so that its API version can be addressed.

lankydan (approved):

PR #459, opened 2022-07-01
NOTICK - Make trace ID an optional field

dimosr:
Making trace ID optional, since there can be use cases where it's not needed at all (e.g. if the user does not specify TTL and will not perform replays on its side).


yift-r3 (approved):

williamvigorr3 (approved):
LGTM

PR #460, opened 2022-07-01
CORE-5271 - remove cpk file checksum from CpiCpkEntity

conal-smith:


conal-smith:
I don't think we want to do this because cpk_file_name is also in this class, we would need to find a new home for it as well and these two could be separate task.

PR #464, opened 2022-07-06
CORE-4428 Define JSON schema for gateway

williamvigorr3:
Added a new top level config key corda.p2p.gateway. Added a schema for this key. I kept the new schema compatible with the original config for the gateway.

All that was need to change was the config key used by the Gateway (see https://github.com/corda/corda-runtime-os/pull/1606).


data/db-schema/src/main/resources/net/corda/db/schema/config/migration/config-creation-v1.0.xml
```diff
@@ -65,6 +65,17 @@
             <column name="update_actor" value="admin"/>
         </insert>
 
+        <insert tableName="config" schemaName="${schema.name}">
```
yift-r3, line 68: Is there a way to reduce the duplication?

data/config-schema/src/main/resources/net/corda/schema/configuration/p2p.gateway/1.0/corda.p2p.gateway.json
```diff
@@ -0,0 +1,102 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "$id": "https://corda.r3.com/net/corda/schema/configuration/p2p/1.0/corda.p2p.gateway.json",
+  "title": "Corda P2P Gateway Configuration Schema",
+  "description": "Configuration schema for the P2P gateway.",
+  "type": "object",
+  "properties": {
+    "hostAddress": {
+      "description": "The host name or IP address where the HTTP server will bind.",
+      "type": "string",
```
yift-r3, line 10: Can we use format `hostname` (https://json-schema.org/understanding-json-schema/reference/string.html)

data/config-schema/src/main/resources/net/corda/schema/configuration/p2p.gateway/1.0/corda.p2p.gateway.json
```diff
@@ -0,0 +1,102 @@
...
+        },
+        "initialReconnectionDelay": {
+          "description": "The initial time (in seconds) to wait before trying to reconnect.",
+          "type": "integer",
+          "default": 1,
+          "minimum": 1,
+          "maximum": 3000
+        },
+        "maximalReconnectionDelay": {
+          "description": "The maximal time (in seconds) to delay before trying to reconnect.",
```
yift-r3: There's a `duration` format (https://json-schema.org/understanding-json-schema/reference/string.html) but it looks really notn user-friendly.

williamvigorr3: Unfortunately `duration` is not in draft 7 so we can't use it. Updating to `2019-09` is a bad idea, I think draft 7 was chosen to be compatible with some libraries we are using.

data/config-schema/src/main/resources/net/corda/schema/configuration/p2p.gateway/1.0/corda.p2p.gateway.json
```diff
@@ -0,0 +1,102 @@
...
+          "default": {},
+          "description": "How to check for revocation of the server certificate.",
+          "properties": {
+            "mode": {
+            "oneOf": [
+              {"const": "SOFT_FAIL", "title": "The revocation check will succeed if the revocation status cannot be determined under certain circumstances. See java.security.cert.PKIXRevocationChecker.Option.SOFT_FAIL java doc for more details."},
+              {"const": "HARD_FAIL", "title": "Opposite of SOFT_FAIL - i.e. most rigorous check. Among other things, this check requires that CRL checking URL is available on every level of certificate chain."},
+              {"const": "OFF", "title": "Switch CRL check off."}
+            ],
+            "default": "OFF"
```
dimosr: I think we should switch the default to `SOFT_FAIL` (I also had a discussion with Matthew and he told me that's the case for C4). I think we had put `OFF` as the default in the beginning to ensure the default setup without CRLs would work, but this should be the case with `SOFT_FAIL`, while also using them when there.

data/config-schema/src/main/resources/net/corda/schema/configuration/p2p.gateway/1.0/corda.p2p.gateway.json
```diff
@@ -0,0 +1,102 @@
...
+           }
+          }
+        },
+        "additionalProperties": false
+      }
+    },
+    "connectionConfig": {
+      "description": "Controls the connection pool.",
```
```suggestion
      "description": "Controls the TCP connection pool to peer gateways.",
```

data/config-schema/src/main/resources/net/corda/schema/configuration/p2p.gateway/1.0/corda.p2p.gateway.json
```diff
@@ -0,0 +1,102 @@
...
+      "type": "object",
+      "default": {},
+      "properties": {
+        "maxClientConnections": {
+          "description": "Maximum size of the connection cache.",
```
```suggestion
          "description": "Maximum size of the connection pool.",
```

data/config-schema/src/main/resources/net/corda/schema/configuration/p2p.gateway/1.0/corda.p2p.gateway.json
```diff
@@ -0,0 +1,102 @@
...
+          "maximum": 10000
+        },
+        "initialReconnectionDelay": {
+          "description": "The initial time (in seconds) to wait before trying to reconnect.",
+          "type": "integer",
+          "default": 1,
+          "minimum": 1,
+          "maximum": 3000
+        },
+        "maximalReconnectionDelay": {
```
dimosr: nit: can we refactor this to `maximumReconnectionDelay` (or just `maxReconnectionDelay` for brevity) now?

data/config-schema/src/main/resources/net/corda/schema/configuration/p2p.gateway/1.0/corda.p2p.gateway.json
```suggestion
          "description": "The maximum time (in seconds) to delay before trying to reconnect.",
```

data/config-schema/src/main/resources/net/corda/schema/configuration/p2p.gateway/1.0/corda.p2p.gateway.json
williamvigorr3: I will do that and run a test with the p2p deployment.

data/db-schema/src/main/resources/net/corda/db/schema/config/migration/config-creation-v1.0.xml
williamvigorr3: I don't think so. There is no for loop in liquibase.

dimosr: not sure that's possible (https://stackoverflow.com/questions/25840467/liquibase-common-columns)

yift-r3: We can change the format to YAML and then use anchors...

dimosr (dismissed):
LGTM

data/db-schema/src/main/resources/net/corda/db/schema/config/migration/config-creation-v1.0.xml
dimosr: interesting.. I think we'd need to get some consensus on this with the wider group. You can drop a question in #corda-platform and see what people think.

Keen to get that in without waiting for this though - we can add a tech debt item to address it, if people agree.

williamvigorr3: I don't know if liquidbase is supposed to be imperative (we can have that debate on the channel).

data/config-schema/src/main/resources/net/corda/schema/configuration/p2p.gateway/1.0/corda.p2p.gateway.json
```diff
@@ -0,0 +1,103 @@
...
+    "hostAddress": {
+      "description": "The host name or IP address where the HTTP server will bind.",
+      "type": "string",
+      "format": "hostname",
+      "default": "0.0.0.0"
+    },
+    "hostPort": {
+      "description": "The HTTP port.",
+      "type": "integer",
+      "default": 0,
```
dimosr: I think it would be better to have a constant port as the default, instead of using random assignment. This would also help with our setup (or generally setups), where load balancers might need to know that value in advance. I would use `8080`, which is sort of standard after `80` which might require root privileges.

dimosr (approved):

PR #465, opened 2022-07-08
CORE-5036: Persist group policy (WIP)

yift-r3:
Runtime PR in https://github.com/corda/corda-runtime-os/pull/1617

This PR allow use to persist the group policy properties


data/avro-schema/src/main/resources/avro/net/corda/data/membership/db/request/MembershipPersistenceRequest.avsc
```diff
@@ -14,6 +14,7 @@
       "type": [
         "net.corda.data.membership.db.request.command.PersistRegistrationRequest",
         "net.corda.data.membership.db.request.command.PersistMemberInfo",
+        "net.corda.data.membership.db.request.command.PersistGroupPolicyRequest",
```
charlieR3: `PersistGroupPolicy` would be more in line with the naming of the other commands I think

data/db-schema/src/main/resources/net/corda/db/schema/vnode-vault/migration/vnode-vault-creation-v1.0.xml
```diff
@@ -72,6 +72,19 @@
         <addPrimaryKey columnNames="group_id, member_name"
                        constraintName="vnode_member_info_pkey"
                        tableName="vnode_member_info"/>
+
+        <createTable tableName="vnode_group_policy">
+            <column name="schema_version" type="INT" autoIncrement="true">
```
charlieR3: This is the version of this record, rather than the version of the schema. I'd suggest renaming this to just `version`

data/avro-schema/src/main/resources/avro/net/corda/data/membership/db/request/command/PersistGroupPolicyRequest.avsc
```diff
@@ -0,0 +1,13 @@
+{
+  "type": "record",
+  "name": "PersistGroupPolicyRequest",
```
charlieR3: Same as `PersistGroupPolicy` comment above

data/db-schema/src/main/resources/net/corda/db/schema/vnode-vault/migration/vnode-vault-creation-v1.0.xml
```diff
@@ -72,6 +72,19 @@
...
+                <constraints nullable="false" primaryKey="true"/>
+            </column>
+            <column name="effective_from" type="TIMESTAMP">
```
charlieR3: We could have many group policies effective at one time (depending on what CPI version / group policy version a member has installed). Maybe `created` would be a better name for this column?

membership/src/main/kotlin/net/corda/v5/membership/GroupPolicyProperties.kt
```diff
@@ -0,0 +1,8 @@
+package net.corda.v5.membership
+
+import net.corda.v5.base.types.LayeredPropertyMap
+
+/**
+ * The properties of the group policy
+ */
+interface GroupPolicyProperties : LayeredPropertyMap
```
charlieR3: This seems to only be used internally so could it be moved to the runtime-os? Or you think it needs to be public? We do have the `GroupPolicy` interface public already.

data/avro-schema/src/main/resources/avro/net/corda/data/membership/db/response/query/PersistGroupPolicyResponse.avsc
```diff
@@ -0,0 +1,13 @@
+{
+  "type": "record",
+  "name": "PersistGroupPolicyResponse",
+  "namespace": "net.corda.data.membership.db.response.query",
```
charlieR3: Should this be in `net.corda.data.membership.db.response.query` instead to mirror the request package?

charlieR3 (approved):
LGTM

PR #467, opened 2022-07-08
CORE-4882: remove test changelogs, should be handled instead by loadin…

dickon:
CORE-4882: remove dogs changelog, should be handled instead by loading the CPI.


driessamyn (approved):
LGTM

driessamyn:
Associated `runtime-os` change (and build): https://github.com/corda/corda-runtime-os/pull/1569

PR #468, opened 2022-07-08
CORE-5544 Expose findAll functionality to the persistence service

simon-johnson-r3:


JamesHR3 (commented):
I think you might need to bump the API version here, because any implementations of `PersistenceService` won't compile without overriding the new method.

JamesHR3 (approved):

PR #469, opened 2022-07-08
CORE-5521 fix Avro generation issue 

ronanbrowne:
Fix Avro Generation issue where the Build does not realize the avro schema has changed. 

- This also removes the need to manually call data:avro-schema:clean as a work around by introducing a new cleanAvro task which is always triggered.
- disable build caching for existing avro related tasks ( Delete tasks not catchable by default so not needed in for mentioned item)

### DIFF
--- a/data/avro-schema/build.gradle
+++ b/data/avro-schema/build.gradle
@@ -67,7 +67,6 @@ avro {
 def generateOSGiPackageInfo = tasks.register("generateOSGiPackageInfo", Task) {
     outputs.cacheIf { false }
     group = "codeGeneration"
-
     def packages = [] as Set<String>
     def classes = [] as Set<String>
 
--- a/data/avro-schema/build.gradle
+++ b/data/avro-schema/build.gradle
@@ -43,14 +43,9 @@ sourceSets {
     }
 }
 
-def cleanAvro = tasks.register("cleanAvro", Delete) {
-    group = "codeGeneration"
-    delete generatedAvroDir
-}
-
 def generateAvro = tasks.register("generateAvro", GenerateAvroJavaTask) {
-    dependsOn cleanAvro
     outputs.cacheIf { false }
+    outputs.upToDateWhen { false }
     group = "codeGeneration"
     source("src/main/resources/avro")
     outputDir = generatedAvroDir
--- a/data/avro-schema/build.gradle
+++ b/data/avro-schema/build.gradle
@@ -43,7 +43,14 @@ sourceSets {
     }
 }
 
+def cleanAvero = tasks.register("cleanAvero", Delete) {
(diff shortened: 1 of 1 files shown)


data/avro-schema/build.gradle
```diff
@@ -43,7 +43,14 @@ sourceSets {
...
 }
 
+def cleanAvro = tasks.register("cleanAvro", Delete) {
+    group = "codeGeneration"
+    delete generatedAvroDir
+}
+
 def generateAvro = tasks.register("generateAvro", GenerateAvroJavaTask) {
+    dependsOn cleanAvro
+    outputs.cacheIf { false }
```
omfR3, line 47: Didn't realise it was that easy. Nice :)

omfR3:
:+1:

yarivus (dismissed):
LGTM

chrisr3 (commented):
Is this all just another way of saying:
```gradle
def generateAvro = tasks.register('generateAvro', GenerateAvroJavaTask) {
    outputs.upToDateWhen { false }
    ...
}
```
And maybe add `outputs.cacheIf { false }` too...

benmcmahon100:
> Is this all just another way of saying:
> 
> ```groovy-gradle
> def generateAvro = tasks.register('generateAvro', GenerateAvroJavaTask) {
>     outputs.upToDateWhen { false }
>     ...
> }
> ```
> 
> And maybe add `outputs.cacheIf { false }` too...

@chrisr3 What would be the benefit of changing to use that over the current workaround? Happy to steal the branch and add that change if it makes sense

chrisr3:
> @chrisr3 What would be the benefit of changing to use that over the current workaround? Happy to steal the branch and add that change if it makes sense

I am just trying to simplify the Gradle scripts, because IMHO "Less is _Much_ More" as far as Gradle is concerned. So rather than create a new `cleanAvro` task, just declare that the `generateAvro` task's output are _always_ out of date

benmcmahon100:
> > @chrisr3 What would be the benefit of changing to use that over the current workaround? Happy to steal the branch and add that change if it makes sense
> 
> I am just trying to simplify the Gradle scripts, because IMHO "Less is _Much_ More" as far as Gradle is concerned. So rather than create a new `cleanAvro` task, just declare that the `generateAvro` task's output are _always_ out of date

Ah okay I missed that I definitely agree. @ronanbrowne I thought you were off which is why I volunteered - you happy to make the change instead?

ronanbrowne:
@chrisr3 / @benmcmahon100 sorry just catching up on this PR. Ya that's fine  cleanAvro  can go in favor of outputs.upToDateWhen use.  Thanks all.

benmcmahon100 (approved):

PR #472, opened 2022-07-12
CORE-5520 adding support for ECIES

alexey-kadyrov-r3:
Added required changes for the crypto service and avro messages as well as added notion of the key schemes usage - for signing, ECDH key agreement.


YashNabar (approved):
LGTM

charlieR3 (approved):

PR #476, opened 2022-07-13
NOTICK: Allow Gradle to run on Java 17, while still building for Java 11.

chrisr3:
We use Gradle toolchains to build and test Corda on Java 11, so allow Gradle _itself_ to run on Java 17.


mcgovc (approved):
LGTM

yarivus (commented):
LGTM

PR #479, opened 2022-07-15
CORE-3764 set flow and session related timeouts in a way that do not conflict with each other

LWogan:
set session ttl such that it is not significantly longer than the amount of time a flow  will resend a message that has not received a reply. Theres no need for p2p to keep trying multiple messages if they are duplicates.  

Set resend window closer to the flow wakeup interval as being set significantly lower than this will have no effect on the flow and may cause confusion. 

Values will likely be changed over time depending on customer needs


JamesHR3 (approved):
Let's try this out and see how well it works.

PR #480, opened 2022-07-18
CORE-5601 Unified external events

lankydan:
The sending on events to external workers is unified to simplify the handling of sending requests and receiving responses. This reduces the amount of code we have to write to add new functionality that requires communication with other workers.

As we are making a generic solution there will be limitations that each specific use-case requires, however a number of scenarios have already been identified and the solution _should_ cater for them all.

The generic code handles the sending events to other workers, receiving `ExternalEventResponse`s from them and determining whether to resume a flow or not based on the responses.

## Effects on the API

- `Checkpoint` now contains `ExternalEventState` rather than `CryptoState` and `PersistenceState`. This state will hold information about any type of external event that is to be sent from the flow worker.
- `ExternalEventContext` should be embedded into the events/payloads sent to external workers. This contains context information, which includes, a request id and flow id.
- `ExternalEventResponse`s are returned back to the flow worker after processing an external event. `ExternalEventResponse` contains an Avro serialized byte array of the underlying payload that the external processor returns to the flow worker.
- `ExternalEventResponseError` to convey error information back to the flow worker.


lankydan:
Runtime PR - https://github.com/corda/corda-runtime-os/pull/1669

data/avro-schema/src/main/resources/avro/net/corda/data/flow/state/external/ExternalEventState.avsc
```diff
@@ -1,25 +1,22 @@
 {
   "type": "record",
-  "name": "CryptoState",
-  "namespace": "net.corda.data.flow.state.crypto",
+  "name": "ExternalEventState",
+  "namespace": "net.corda.data.flow.state.external",
   "fields": [
```
owenstanford, line 6: doc?

lankydan:
Pretty sure @driessamyn and @dickon are aware of this, but linking anyway.

data/avro-schema/src/main/resources/avro/net/corda/data/flow/event/external/ExternalEventContext.avsc
```diff
@@ -0,0 +1,18 @@
...
+    {
+      "name": "requestId",
+      "type": "string",
+      "doc": "The request id of the external event. Used for deduplication when processing responses."
+    },
+    {
+      "name": "flowId",
+      "type": "string",
+      "doc": "The request id of the external event. Used for deduplication when processing responses."
+    }
```
JamesHR3: I assume one of the doc strings is wrong (or you don't need one of them!). I guess it's the flow ID, which we're using to ensure the response gets put back on the right key?

data/avro-schema/src/main/resources/avro/net/corda/data/flow/event/external/ExternalEventResponseExceptionEnvelope.avsc
```diff
@@ -0,0 +1,27 @@
+{
+  "type": "record",
+  "name": "ExternalEventResponseExceptionEnvelope",
```
JamesHR3: I don't think we should call this exception envelope, rather it's an external event error.

data/avro-schema/src/main/resources/avro/net/corda/data/flow/event/external/ExternalEventResponse.avsc
```diff
@@ -0,0 +1,62 @@
...
+      "name": "payload",
+      "type": [
+        "null",
+        "net.corda.data.persistence.EntityResponse",
+        "net.corda.data.crypto.wire.ops.flow.FlowOpsResponse"
+      ],
+      "doc": "The underlying payload that can contain domain specific information that is not serialized into the {@link ExternalEventResponse#data} property. `null` if the response represents an error and therefore does not have a payload to respond with."
+    },
+    {
+      "name": "data",
```
JamesHR3: What's the distinction between `payload` and `data`?

Should we use a union on the response side so we reply either with success or failure, rather than nulling some fields out some of the time? (I know we use that pattern elsewhere, I just don't like it very much!)

data/avro-schema/src/main/resources/avro/net/corda/data/flow/event/FlowEvent.avsc
```diff
@@ -14,8 +14,7 @@
         "net.corda.data.flow.event.Wakeup",
         "net.corda.data.flow.output.FlowStatus",
         "net.corda.data.flow.event.SessionEvent",
-        "net.corda.data.persistence.EntityResponse",
-        "net.corda.data.crypto.wire.ops.flow.FlowOpsResponse"
+        "net.corda.data.flow.event.external.ExternalEventResponse"
```
JamesHR3, line 17: Not related to your PR, but I've just noticed `FlowStatus` is in this list and it shouldn't be. Whoops!

data/avro-schema/src/main/resources/avro/net/corda/data/flow/event/external/ExternalEventContext.avsc
```diff
@@ -0,0 +1,18 @@
+{
+  "type": "record",
+  "name": "ExternalEventContext",
```
JamesHR3, line 3: I think this should be baked into the external event rather than on the individual request types. We probably shouldn't be polluting client-specific data structures with our stuff. (They will need to read it, but it's part of our API, not theirs.)

JamesHR3 (commented):
Looks pretty close. I have a couple of thoughts, but nothing major stands out.

data/avro-schema/src/main/resources/avro/net/corda/data/flow/event/FlowEvent.avsc
lankydan: I dont think that's true right? The `FlowStatus` should never be received by the flow which is what `FlowEvent` is used for.

JamesHR3: I think we're saying the same thing? But the issue is it's already in there, so we're currently saying it can be received by the flow.

lankydan: oh I read it as "should be"... 🤦

data/avro-schema/src/main/resources/avro/net/corda/data/flow/event/external/ExternalEventContext.avsc
lankydan: Needs to be on the underlying payloads so that each processor can receive whatever objects they want rather than one specified by the external event framework.

data/avro-schema/src/main/resources/avro/net/corda/data/flow/event/external/ExternalEventResponse.avsc
lankydan: Removed data.

data/avro-schema/src/main/resources/avro/net/corda/data/flow/event/external/ExternalEventResponseExceptionEnvelope.avsc
lankydan: Sure, will change.

data/avro-schema/src/main/resources/avro/net/corda/data/flow/event/external/ExternalEventContext.avsc
lankydan: Fixed.

lankydan:
__When merged used my commit description__

__When merged used my commit description__

__When merged used my commit description__

__When merged used my commit description__

__When merged used my commit description__

data/avro-schema/src/main/resources/avro/net/corda/data/flow/event/external/ExternalEvent.avsc
```diff
@@ -0,0 +1,31 @@
...
+      "doc": "The topic that the external event is sent to."
+    },
+    {
+      "name": "key",
+      "type": "bytes",
+      "doc": "Avro serialized representation of the key that the external event is partitioned by."
+    },
+    {
+      "name": "payload",
+      "type": "bytes",
```
driessamyn, line 19: I raised this JIRA a little while ago: https://r3-cev.atlassian.net/browse/CORE-5931

Given we are refactoring this, would it be worth introducing this now?
I think for now the AMQPPayload schema could probably just be a wrapper for `bytes`, but would make the API more explicit.

lankydan: In this case it would be an `AvroPayload` though rather than `AMQPPayload`, just making sure we're on the same page.

driessamyn: Ah. Sorry, misread that. I thought this was an AMQP payload (as it would be in `EntityResponse` for example. I guess more so reason to be specific ;)

But why is this bytes and not a union of allowed payloads? Doesn't that make the API more difficult to interpret?

Thinking, for example, of the usecase of parking these into a Datawarehouse, how would somebody be able to figure out how to convert this?

JamesHR3: We changed this deliberately. The problem with this being a union is that we're expecting to have to generate external events from layers of the API higher up than the flow API (ledger being the primary example). We're going to rapidly run into an extensibility problem here if we force a union where we don't control all the possible variants that could be in that union.

driessamyn: Understood.

Would it make sense to use a wrapper schema as discussed above (AvroPayload) and include the name/fingerprint of the schema in there (similar to our wrapper at the top level)? Then at least you could programmatically deserialise things outside Corda as long as you have access to the Schema registry.

data/avro-schema/src/main/resources/avro/net/corda/data/flow/event/external/ExternalEventResponse.avsc
```diff
@@ -0,0 +1,34 @@
...
+  "name": "ExternalEventResponse",
+  "namespace": "net.corda.data.flow.event.external",
+  "doc": "Events from workers that are sent back to the flow worker to be consumed by flows",
+  "fields": [
+    {
+      "name": "requestId",
+      "type": "string",
+      "doc": "The request id of the external event response. Used for deduplication when processing responses."
+    },
+    {
```
owenstanford, line 12: Should the payload be 
```"type": ["bytes", "net.corda.data.flow.event.external.ExternalEventResponseError"]```
to force the response to be one or the other?

File: data/avro-schema/src/main/resources/avro/net/corda/data/flow/event/external/ExternalEventResponseError.avsc [hunk 9]
```diff
@@ -0,0 +1,27 @@
...
+  "fields": [
+    {
+      "name": "errorType",
+      "doc": "Error type",
+      "type": {
+        "name": "ExternalEventResponseErrorType",
+        "type": "enum",
+        "doc": "RETRY: A retriable error.\nPLATFORM: An error that should be propagated to the calling flow.\nFATAL: A fatal error.",
+        "symbols": [
+          "RETRY",
```
owenstanford: Maybe TRANSIENT rather than RETRY to describe the type of error rather than the action to take?

data/avro-schema/src/main/resources/avro/net/corda/data/flow/state/external/ExternalEventState.avsc
```diff
@@ -0,0 +1,52 @@
+{
```
owenstanford, line 1: missing doc string

data/avro-schema/src/main/resources/avro/net/corda/data/flow/state/external/ExternalEventStateStatus.avsc
```diff
@@ -0,0 +1,30 @@
+{
+  "type": "record",
+  "name": "ExternalEventStateStatus",
+  "namespace": "net.corda.data.flow.state.external",
```
owenstanford, line 4: doc

data/avro-schema/src/main/resources/avro/net/corda/data/flow/state/waiting/external/ExternalEventResponse.avsc
```diff
@@ -0,0 +1,12 @@
+{
+  "type": "record",
+  "name": "ExternalEventResponse",
+  "namespace": "net.corda.data.flow.state.waiting.external",
```
owenstanford, line 4: doc

data/avro-schema/src/main/resources/avro/net/corda/data/flow/event/external/ExternalEvent.avsc
driessamyn: Was thinking about this again this morning, not sure why, but shouldn't we use [the existing `AvroEnvelope`](https://github.com/corda/corda-api/blob/release/os/5.0/data/avro-schema/src/main/resources/avro/net/corda/data/AvroEnvelope.avsc) in this case?
@lankydan , @JamesHR3 

I'm thinking again in case you want to archive Kafka messages for analysis, or even just monitoring (some orgs will want to do this). From building [KafkaSnoop](https://github.com/driessamyn/kafkasnoop), I know you need to have some indication of what the schema is in order to do anything with the binary payload. 
With AMQP it's less important as you don't need a schema (although you still will want to know it's AMQP).

lankydan: @driessamyn @JamesHR3 I'll raise a tech debt ticket for this and address it separately.

lankydan: Created - https://r3-cev.atlassian.net/browse/CORE-6401

kyriathar (dismissed):

data/avro-schema/src/main/resources/avro/net/corda/data/flow/event/external/ExternalEvent.avsc
mnesbit: The serialization library is reading the schema from the byte array which contains the fingerprint. So the ByteArray is a lot more self-describing than it seems.

data/avro-schema/src/main/resources/avro/net/corda/data/flow/state/external/ExternalEventState.avsc
```diff
@@ -0,0 +1,53 @@
...
+  "type": "record",
+  "name": "ExternalEventState",
+  "namespace": "net.corda.data.flow.state.external",
+  "doc": "ExternalEventState contains the information required to send an external event and receive a response.",
+  "fields": [
+    {
+      "name": "requestId",
+      "type": "string",
+      "doc": "Unique ID of a single crypto request. The Id remains the same for each resend."
```
blsemo: Crypto request?

mnesbit (approved):
ok

blsemo (approved):

PR #481, opened 2022-07-19
CORE-5745 - remove un-needed things from persistence API.

driessamyn:
- Remove postFilter and postProcessorName from the persitence API as this will be implemented as a ledger specific feature on the ledger API.
- Add offset and limit parameters to the query API.
- Change `Cursor` return type to `List`
- Changed API slightly after discussion.

Related `corda-runtime-os` changes: https://github.com/corda/corda-runtime-os/pull/1694

Notes:
- `NamedQueryFilter` and `GenericQueryPostProcessor` was referenced in the legacy ledger implementation, so I moved them in that module.
- Can the `Cursor` interface (and associated interfaces) be removed from `base`?


application/src/main/kotlin/net/corda/v5/application/persistence/PersistenceQueryRequest.kt
```diff
@@ -20,45 +16,33 @@ data class PersistenceQueryRequest(
...
      */
-    val postFilter: NamedQueryFilter?,
+    val offset: Int,
     /**
-     * The name of the post-processor that will process named query results.
+     * The maximum number of results to return.
+     * Default is "all records".
      */
-    val postProcessorName: String?
+    val limit: Int
```
driessamyn: The JPA API uses `maxResults` and `firstResult`, but limit and offset are more commonly used in the SQL world, which is why I aligned with that, but happy to get other opinions.

application/src/main/kotlin/net/corda/v5/application/persistence/PersistenceService.kt
```diff
@@ -116,60 +115,43 @@ interface PersistenceService {
...
      */
     @Suspendable
     fun <R> query(
         queryName: String,
         namedParameters: Map<String, Any>,
-        postFilter: NamedQueryFilter,
-        postProcessorName: String
+        offset: Int,
+        limit: Int,
     ): Cursor<R>
```
driessamyn: I didn't add one that has just offset, as I was thinking limit by itself may make sense, but maybe there aren't many use cases for using offset without limit.
Happy to add one though.

ledger-obsolete/src/main/kotlin/net/corda/v5/ledger/obsolete/services/vault/NamedQueryFilter.kt
```diff
@@ -0,0 +1,6 @@
+package net.corda.v5.ledger.obsolete.services.vault
+
+import net.corda.v5.base.annotations.CordaSerializable
+
+@CordaSerializable
+interface NamedQueryFilter
```
driessamyn, line 6: FYI @blsemo, @vlajos 
We removed post filter/post processing from the public persistence API. But as legacy ledger is still referencing this, I moved the objects here so it continues to compile.
Please let me know if this is an issue.

ledger-obsolete/src/main/kotlin/net/corda/v5/ledger/obsolete/services/vault/GenericQueryPostProcessor.kt
```diff
@@ -1,4 +1,4 @@
-package net.corda.v5.application.persistence.query
+package net.corda.v5.ledger.obsolete.services.vault
```
driessamyn, line 1: FYI @blsemo, @vlajos 
We removed post filter/post processing from the public persistence API. But as legacy ledger is still referencing this, I moved the objects here so it continues to compile.
Please let me know if this is an issue.

application/src/main/kotlin/net/corda/v5/application/persistence/PersistenceQueryRequest.kt
```diff
@@ -1,14 +1,10 @@
...
-
 /**
  * Request object for making Query requests on the Persistence Service.
  *
  * @param queryName the name of the named query registered in the persistence context.
  * @param namedParameters the named parameters to be set in the named query.
- * @param postFilter the filter to be applied after named query execution.
- * @param postProcessorName the name of the post-processor that will process named query results.
  */
 data class PersistenceQueryRequest(
```
lankydan: If anyone can think of a better name for this class then that would be great. It doesn't _feel_ right but I also can't think of what is the correct name...

application/src/main/kotlin/net/corda/v5/application/persistence/PersistenceService.kt
```diff
@@ -105,79 +103,62 @@ interface PersistenceService {
...
      * @throws CordaPersistenceException if an error happens during query operation
      */
     @Suspendable
     fun <R> query(
         queryName: String,
         namedParameters: Map<String, Any>,
-        postFilter: NamedQueryFilter,
-        postProcessorName: String
-    ): Cursor<R>
+        offset: Int,
```
lankydan: Do we think there are ever going to be other properties to add to the query similar to `offset` and `limit`? I think having them in an object for backwards compatibility in the future would make sense without the need for tons of overloads to the `PersistenceService` (although this would be somewhat moved to the request object).

If we don't see there being many extra properties to add in the future, then what we have here is fine.

application/src/main/kotlin/net/corda/v5/application/persistence/PersistenceService.kt
```diff
@@ -105,79 +103,62 @@ interface PersistenceService {
...
+     * Execute a named query in a single transaction with optional offset and limit applied to the results. Casts results to the specified type [R].
      *
      * @param persistenceQueryRequest the request containing information to execute named queries with optional filtering and post-processing
      * @param R the type of the results
-     * @return Cursor configured to poll data for this named query.
+     * @return list of entities found. Empty list if none were found.
      * @throws CordaPersistenceException if an error happens during query operation
      */
     @Suspendable
-    fun <R> query(persistenceQueryRequest: PersistenceQueryRequest): Cursor<R>
```
lankydan, line 180: I think we should drop this or keep it depending on what the answer to my comment above is. Since we added the `PersistenceQueryRequest` due to the amount of properties there were at the time (like 5?). The current max is 4 with the offset + limit so the benefit has gone down a bit. Curious what others think.

application/src/main/kotlin/net/corda/v5/application/persistence/PersistenceService.kt
```diff
@@ -105,79 +103,62 @@ interface PersistenceService {
      *
      * @param queryName the name of the named query registered in the persistence context.
      * @param namedParameters the named parameters to be set in the named query.
-     * @param R the type of the results
-     * @return Cursor configured to poll data for this named query.
+     * @param R the type of the results.
+     * @return list of entities found. Empty list if none were found.
      * @throws CordaPersistenceException if an error happens during query operation
      */
     @Suspendable
```
lankydan: I think we should just bin this one and force paging/limiting onto developers.

application/src/main/kotlin/net/corda/v5/application/persistence/PersistenceQueryRequest.kt
driessamyn: Maybe just `Query` ?
The JPA equivalent I think is just `Query`, which is returned from `createNamedQuery()` or `createdQuery()`.
Although in JPA it is both the builder and the thing that represents the query.

application/src/main/kotlin/net/corda/v5/application/persistence/PersistenceService.kt
driessamyn: The same crossed my mind and I was following the pattern that was there, but maybe we don't need the overloads.

If you think of the JPA API, it's quite different. There, you have query factories: `createNamedQuery()`, `createQuery()` and they return a `Query` object which is effectively a builder that exposes methods for building the query (`setParameter()`, `setMaxResults()` ... all return Query) and also a result functions (`getResultList()`, `getResultStream()` ...)
That's a quite JAVA-y API, but feels quite natural to me.

lankydan: I don't _mind_ that approach, however I do quite like the `Map<String, Any>` for kotlin usage, for Java the `setParameter` option probably still looks nicer even in newer versions of Java. Maybe we provide an option to pass in a map of parameters via `setParameters` and `setParameter` for per parameter usage. We can then remove the other APIs and just provide this one?

If we do this, then `Query` like you said in your previous comment sounds good to me.

Want more people than just me to agree though since it's drifting further away from where we were in DP1.

Also looping in @conalsmith-r3 since he created the original API. I don't think we considered doing it this way at the time so would be good to know what he thinks too.

application/src/main/kotlin/net/corda/v5/application/persistence/PersistenceService.kt
lankydan: Agreed, don't think it makes sense by itself,

application/src/main/kotlin/net/corda/v5/application/persistence/PersistenceService.kt
sprintengine: It would be quite nice having just one query function here returning a `Query` builder object and the user can set the parameters on it themselves, I think it would be intuitive for the user since it is similar to JPA Query object.

sprintengine: Something like this?

```
interface Query<R> {
    fun setParameter(name: String, value: Any): Query<R>
    fun setParameters(parameters: Map<String, Any>): Query<R>
    fun setLimit(limit: Int): Query<R>
    fun setOffset(offset: Int): Query<R>
    fun execute(): List<R>
}
```

```
    @Suspendable
    fun <R> query(
        queryName: String
    ): Query<R>
```

application/src/main/kotlin/net/corda/v5/application/persistence/Query.kt
```diff
@@ -0,0 +1,15 @@
...
+ *
+ * @param R the type of the results.
+ */
+interface Query<R> {
+    /**
+     * Execute the [Query]
+     *
+     * @return list of entities found. Empty list if none were found.
+     */
+    fun execute(): List<R>
```
JamesHR3, line 17: Is this the point where we'd go to the DB worker to run it? If so this needs to be marked `@Suspendable`. (We should probably do that for all APIs anyway as a hedge, even if it's not actually going to suspend, as adding the annotation later is not binary compatible with an old version without it.)

sprintengine, line 17: I think `@suspendable` will need to be removed from some of the PersistenceService functions and put here?

application/src/main/kotlin/net/corda/v5/application/persistence/PersistenceService.kt
```diff
@@ -92,73 +92,26 @@ interface PersistenceService {
...
+     * @return a [PagedQuery] that returns the list of entities found.
      * @throws CordaPersistenceException if an error happens during find operation
      */
     @Suspendable
-    fun <T : Any> findAll(entityClass: Class<T>): List<T>
+    fun <T : Any> findAll(entityClass: Class<T>): PagedQuery<T>
 
     /**
-     * Execute a named query in a single transaction. Casts results to the specified type [R].
+     * Execute a named query in a single transaction. Casts results to the specified type [T].
```
JamesHR3: Presumably this isn't executing the query any more, it's just generating it, and the user needs to fill in any parameters/offsets/limits before running it?

I wouldn't be against putting a quick usage example in here just to emphasise the point. (Our API docs are pretty bad in this regard, it's something we need to take a look at at some point soon.)

application/src/main/kotlin/net/corda/v5/application/persistence/Query.kt
lankydan, line 17: Needs `@Suspendable` on it.

JamesHR3 (commented):
I think I prefer this structure, just a few comments.

application/src/main/kotlin/net/corda/v5/application/persistence/Query.kt
sprintengine, line 17: Thoughts on having `fun getResultList(): List<R>` and `fun singleResult(): R` instead of execute()?

lankydan (commented):
Personally I'm happy with this structure, it looks similar to jpa/hibernate and gives us the flexibility we need for backwards compatibility in the future.

application/src/main/kotlin/net/corda/v5/application/persistence/Query.kt
driessamyn: I added it on `execute()` and left it on the methods in the `persistenceService`. Is that what you meant?

application/src/main/kotlin/net/corda/v5/application/persistence/PersistenceService.kt
driessamyn: I updated the kdocs and added a code example. I wasn't sure what the best way to do this is in Kdoc, I couldn't find an example, and I wasn't sure whether it should be Java or Kotlin, but have a look.
It seems like in Kdoc you can either do this by using markdown, or you can reference an actual sample, which is maybe something we should think about, but then we ought to do it for the entire API?

driessamyn: ik

driessamyn: Renders like this:
image

application/src/main/kotlin/net/corda/v5/application/persistence/Query.kt
driessamyn: I think the conclusion on the slack discussion was to stick with `execute`. Do you agree?

application/src/main/kotlin/net/corda/v5/application/persistence/PersistenceService.kt
lankydan: I think the API looks really nice when you see it shown for real.

lankydan (dismissed):

conal-smith (dismissed):
Looks great!

application/src/main/kotlin/net/corda/v5/application/persistence/Query.kt
JamesHR3: Yeah that seems fine. Can't imagine we'd want to suspend setting parameters!

application/src/main/kotlin/net/corda/v5/application/persistence/PersistenceService.kt
JamesHR3: Looks good! The whole API with examples thing is probably an epic in itself so let's not worry too much about that here. Something to think about though.

JamesHR3 (dismissed):
Looks good!

application/src/main/kotlin/net/corda/v5/application/persistence/PagedQuery.kt
```diff
@@ -0,0 +1,24 @@
+package net.corda.v5.application.persistence
+
+/**
+ * Used to build a Query that supports limit and offset.
+ *
+ * @param R the type of the results.
+ */
+interface PagedQuery<R> : Query<R> {
```
simon-johnson-r3, line 8: I assume the default offset is 0, but what about the default limit? If that is unlimited, should you also specify a way to explicitly set that using the method for consistency?

simon-johnson-r3, line 8: I assume the default offset is 0, but what about the default limit? If that is unlimited, should you also specify a way to explicitly set that using the method for consistency?

I think the defaults should be documented also.

simon-johnson-r3:
Is paging independent of the chunking of results to meet the avro message size limitation?

application/src/main/kotlin/net/corda/v5/application/persistence/PagedQuery.kt
driessamyn: That's a good point. I assumed the default is `Integer.MAX_VALUE`, which is the JPA default, and I think makes sense, because you basically say "I don't want to set a limit".

I will add to the comments.

application/src/main/kotlin/net/corda/v5/application/persistence/PersistenceService.kt
driessamyn: > Is paging independent of the chunking of results to meet the avro message size limitation?

Yes. And in theory, once we have chunking on the queries, you wouldn't need pages as it will query for all results, then chunk them up, but that may not be great from a performance point of view, so it may actually be worth thinking about whether we should impose a limit, or at least warn people against it.

driessamyn:
> Is paging independent of the chunking of results to meet the avro message size limitation?

Yes. And in theory, once we have chunking on the queries, you wouldn't need pages as it will query for all results, then chunk them up, but that may not be great from a performance point of view, so it may actually be worth thinking about whether we should impose a limit, or at least warn people against it.

driessamyn:
Green `corda-runtime-os` build for this version: https://ci02.dev.r3.com/blue/organizations/jenkins/Corda5%2Fcorda-runtime-os/detail/PR-1694/4/pipeline

simon-johnson-r3 (approved):

JamesHR3 (approved):

PR #482, opened 2022-07-19
CORE-5338: add FindWithNamedQuery.avsc

dickon:
CORE-5338: add a simple FindWithNamedQuery message for the persistence API

### DIFF
--- a/data/avro-schema/src/main/resources/avro/net/corda/data/persistence/FindWithNamedQuery.avsc
+++ b/data/avro-schema/src/main/resources/avro/net/corda/data/persistence/FindWithNamedQuery.avsc
@@ -1,0 +1,31 @@
+{
+  "type": "record",
+  "name": "FindWithNamedQuery",
+  "doc": "Find entities matched by a named query. Parameters for the query may be specified. Pagination is supported, but there's no guarantee of consistency if the underlying data changes or is unsorted.  Replies with {@link EntityResponse}",
+  "namespace": "net.corda.data.persistence",
+  "fields": [
+    {
+      "name": "queryName",
+      "type": "string",
+      "doc": "The name of a named query defined somewhere in the CPK."
+    },
+    {
+      "name": "parameters",
+      "type": {
+        "type": "map",
+        "values": "bytes"
+      },
+      "doc": "Parameters for the query, with each value represented as a bytes array payload serialized in AMQP format. Null values must be tested using IS NULL in the query; it is not possible to pass in a null as a parameter value. "
+    },
+    {
+       "name": "offset",
+       "type" : "int",
+       "doc": "The index of the first result in the query output that should be returned."
+    },
+    {
+        "name": "limit",
+        "type": "int",
+        "doc": "Limit number of rows in the output query results, after applying the offset. Use the maximum int value if you do not want a lower limit."
+    }
+  ]
+}


data/avro-schema/src/main/resources/avro/net/corda/data/persistence/FIndWithNamedQuery.avsc
```diff
@@ -0,0 +1,21 @@
+{
+  "type": "record",
+  "name": "FindWithNamedQuery",
+  "doc": "Find entities which match named query with given parameters.  Replies with {@link EntityResponse}",
+  "namespace": "net.corda.data.persistence",
+  "fields": [
+    {
+      "name": "queryName",
+      "type": "string",
+      "doc": "the name of a named query defined somewhere in the virtual node"
```
driessamyn:
```suggestion
      "doc": "The name of a named query defined in the CPK."
```

data/avro-schema/src/main/resources/avro/net/corda/data/persistence/FIndWithNamedQuery.avsc
```diff
@@ -0,0 +1,21 @@
...
+    },
+    {
+      "name": "parameters",
+      "type": {
+        "type": "map",
+        "values": "bytes"
+      },
+      "doc": "parameters for the query"
```
driessamyn:
```suggestion
      "doc": "Parameters for the query. Value can be a complex type and is an AMQP serialised payload."
```

data/avro-schema/src/main/resources/avro/net/corda/data/persistence/FIndWithNamedQuery.avsc
```diff
@@ -0,0 +1,21 @@
...
+    {
+      "name": "queryName",
+      "type": "string",
+      "doc": "the name of a named query defined somewhere in the virtual node"
+    },
+    {
+      "name": "parameters",
+      "type": {
+        "type": "map",
+        "values": "bytes"
```
driessamyn: I think we said this should be `bytes` or `null` as null is a valid paramater in a JPA query.

data/avro-schema/src/main/resources/avro/net/corda/data/persistence/FIndWithNamedQuery.avsc
```diff
@@ -0,0 +1,21 @@
+{
+  "type": "record",
+  "name": "FindWithNamedQuery",
```
driessamyn: I think it's worth adding the offset and limit parameters too. I know we have 2 JIRAs splitting it up, but adding them now will avoid another API change that then needs a version bump.

data/avro-schema/src/main/resources/avro/net/corda/data/persistence/FIndWithNamedQuery.avsc
```diff
@@ -0,0 +1,31 @@
...
+      "type": {
+        "type": "map",
+        "values": "bytes"
+      },
+      "doc": "parameters for the query, with each value represented as a bytes array payload serialized in AMQP format"
+    },
+    {
+       "name": "offset",
+       "type" : "long",
+       "doc": "offset into the query results"
```
driessamyn: Brought the language in line with the public API:

```suggestion
       "doc": "The index of the first result in the query to return."
```

data/avro-schema/src/main/resources/avro/net/corda/data/persistence/FIndWithNamedQuery.avsc
```diff
@@ -0,0 +1,31 @@
...
+      "type": "string",
+      "doc": "the name of a named query defined somewhere in the virtual node"
+    },
+    {
+      "name": "parameters",
+      "type": {
+        "type": "map",
+        "values": "bytes"
+      },
+      "doc": "parameters for the query, with each value represented as a bytes array payload serialized in AMQP format"
```
driessamyn:
```suggestion
      "doc": "Parameters for the query, with each value represented as a bytes array payload serialized in AMQP format."
```

data/avro-schema/src/main/resources/avro/net/corda/data/persistence/FIndWithNamedQuery.avsc
```diff
@@ -0,0 +1,31 @@
...
+    },
+    {
+       "name": "offset",
+       "type" : "long",
+       "doc": "offset into the query results"
+    },
+    {
+        "name": "limit",
+        "type": "long",
+        "doc": "limit number of query results; 0 for no limit"
```
driessamyn: The public API says Int.Max is the default, so I don't think we should say 0 is a limit here. I think it's ok for it to be a mandatory property in our internal API and we'll take whatever is passed in. (So `0` would mean don't return anything).

Also brought the language in line with the public API.

```suggestion
        "doc": "Maximum number of results to return."
```

driessamyn (changes requested):
Sorry, bit nitpicky about the documentation, but I think it's important style is consistent with other schemas a we may be generating documentation from them.

I realise that some others in this directory are also not consistent, but I will try to fix that up with a separate PR.

driessamyn (dismissed):
LGTM

lankydan (dismissed):

driessamyn (approved):
LGTM

PR #484, opened 2022-07-20
CORE-5145 Merkle tree doc tweaks

vlajos:
* Update Nonce Hash Digest Provider's expected entropy length.
 * Clarify Merkle Tree factory expected options length

Implementation: https://github.com/corda/corda-runtime-os/pull/1683

### DIFF
--- a/crypto/src/main/kotlin/net/corda/v5/crypto/merkle/MerkleTreeFactory.kt
+++ b/crypto/src/main/kotlin/net/corda/v5/crypto/merkle/MerkleTreeFactory.kt
@@ -10,7 +10,7 @@ import net.corda.v5.base.annotations.Suspendable
 const val HASH_DIGEST_PROVIDER_DEFAULT_NAME = "DefaultHashDigestProvider"
 
 /**
- * A merkle tree digest that one-way derives a set of per-leaf nonces from a 16-byte entropy input. The nonces are
+ * A merkle tree digest that one-way derives a set of per-leaf nonces from a 32-byte entropy input. The nonces are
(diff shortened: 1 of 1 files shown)


crypto/src/main/kotlin/net/corda/v5/crypto/merkle/MerkleTreeFactory.kt
```diff
@@ -40,17 +40,17 @@ const val HASH_DIGEST_PROVIDER_NONCE_SIZE_ONLY_VERIFY_NAME = "NonceHashDigestPro
...
 
 /**
- * Tweakable hash digest providers require a ByteArray node prefix option.
+ * Tweakable hash digest providers require a non-empty ByteArray node prefix option.
  */
 const val HASH_DIGEST_PROVIDER_NODE_PREFIX_OPTION = "nodePrefix"
 
 /**
- * Nonce hash digest providers require a ByteArray entropy option.
+ * Nonce hash digest providers require a 32 character long ByteArray entropy option.
```
mnesbit: we should say bytes not characters

mnesbit (dismissed):
Minor wording request, but otherwise fine

crypto/src/main/kotlin/net/corda/v5/crypto/merkle/MerkleTreeFactory.kt
vlajos: Thanks. I have updated it.

blsemo (approved):
LGTM

PR #487, opened 2022-07-25
CORE-5001 Consensual ledger structures

vlajos:
ledger-common:
* PrivacySalt

ledger-consensual:
* ConsensualLedgerService
* ConsensualState
* Party
* ConsensualLedgerTransaction
* ConsensualSignedTransaction
* ConsensualTransactionBuilder

Implementation: https://github.com/corda/corda-runtime-os/pull/1715


ledger/ledger-common/src/main/kotlin/net/corda/v5/ledger/common/transactions/CpkIdentifier.kt
```diff
@@ -0,0 +1,11 @@
+package net.corda.v5.ledger.common.transactions
+
+import net.corda.v5.base.annotations.CordaSerializable
+
+//TODO: Mainly from [net.corda.libs.cpiupload.endpoints.v1.CpkIdentifier] Clarify their relationship
+@CordaSerializable
+data class CpkIdentifier(
```
mnesbit: Interesting question if we should expose the metadata to flows.

ledger/ledger-consensual/src/main/kotlin/net/corda/v5/ledger/consensual/transaction/ConsensualWireTransaction.kt
```diff
@@ -0,0 +1,24 @@
...
+import net.corda.v5.crypto.SecureHash
+import net.corda.v5.crypto.merkle.HASH_DIGEST_PROVIDER_TWEAKABLE_NAME
+import net.corda.v5.ledger.common.transactions.PrivacySalt
+import net.corda.v5.ledger.consensual.ConsensualState
+import java.security.PublicKey
+import java.time.Instant
+
+@DoNotImplement
+@CordaSerializable
+interface ConsensualWireTransaction {
```
mnesbit: I am not quite sure of the value of this interface. We do need a WireTransaction that holds a PrivacySalt and a List<List<ByteArray>>, however it is largely an internal concern of serialization and storage. This interface is presumably meant for informational purposes in which case it should be talking in terms of concrete information e.g. Class<*> for state types and deserialized PublicKeys for signers and states, but that blurs into whatever the 'LedgerTransaction' type is so maybe we don't need to expose this.

ledger/ledger-consensual/src/main/kotlin/net/corda/v5/ledger/consensual/transaction/ConsensualTransactionBuilder.kt
```diff
@@ -0,0 +1,31 @@
...
+import net.corda.v5.ledger.consensual.ConsensualState
+import net.corda.v5.ledger.consensual.Party
+import java.security.PublicKey
+import java.security.SecureRandom
+import java.time.Instant
+
+@DoNotImplement
+interface ConsensualTransactionBuilder {
+    val metadata: ConsensualTransactionMetaData?
+    fun withMetadata(metadata: ConsensualTransactionMetaData): ConsensualTransactionBuilder
```
mnesbit: I think the meta will all be added internally by the builder, but we should discuss

ledger-obsolete/src/main/kotlin/net/corda/v5/ledger/obsolete/transactions/LedgerTransaction.kt
```diff
@@ -13,6 +13,7 @@ import net.corda.v5.ledger.obsolete.contracts.TimeWindow
 import net.corda.v5.ledger.obsolete.contracts.TransactionState
 import net.corda.v5.ledger.obsolete.contracts.TransactionVerificationException
 import net.corda.v5.ledger.obsolete.identity.Party
+import net.corda.v5.ledger.common.transactions.PrivacySalt
```
mnesbit: We should create the consensual state equivalent of LedgerTransaction

ledger/ledger-consensual/src/main/kotlin/net/corda/v5/ledger/consensual/transaction/ConsensualTransactionBuilder.kt
```diff
@@ -0,0 +1,31 @@
...
+
+    val consensualStates: List<ConsensualState>
+    fun withConsensualState(consensualState: ConsensualState) : ConsensualTransactionBuilder
+
+    fun build(
+        merkleTreeFactory: MerkleTreeFactory,
+        digestService: DigestService,
+        secureRandom: SecureRandom,
+        serializer: SerializationService
+    ): ConsensualWireTransaction
```
blsemo: This should return a signed transaction, I believe (which could be a common class for UTXO and Consensual ledger)

ledger/ledger-consensual/src/main/kotlin/net/corda/v5/ledger/consensual/transaction/ConsensualTransactionMetaData.kt
```diff
@@ -0,0 +1,9 @@
+package net.corda.v5.ledger.consensual.transaction
+
+import net.corda.v5.base.annotations.DoNotImplement
+import net.corda.v5.ledger.common.transactions.TransactionMetaData
+
+@DoNotImplement
+interface ConsensualTransactionMetaData : TransactionMetaData {
```
blsemo: I don't think we need this interface?

ledger/ledger-common/src/main/kotlin/net/corda/v5/ledger/common/transactions/TransactionMetaData.kt
```diff
@@ -0,0 +1,16 @@
+package net.corda.v5.ledger.common.transactions
+
+import net.corda.v5.base.annotations.CordaSerializable
+import net.corda.v5.base.annotations.DoNotImplement
+
+/**
+ * TODO(Design question: Maybe just a simple map to make future changes easier?, Or just adding an options map to it?)
```
blsemo: Yes, I think a map + known constants would be better - then ledger specific metadata would just be some extra constants.
Could be a `Map<String, Any>` with some extension functions to fish out things? The extension functions could also be impl only.

ledger/ledger-consensual/src/main/kotlin/net/corda/v5/ledger/consensual/Party.kt
```diff
@@ -0,0 +1,16 @@
...
+
+import net.corda.v5.base.annotations.CordaSerializable
+import net.corda.v5.base.annotations.DoNotImplement
+import net.corda.v5.base.types.MemberX500Name
+import net.corda.v5.crypto.toStringShort
+import java.security.PublicKey
+
+@DoNotImplement
+@CordaSerializable
+interface Party {
```
blsemo, line 16: Do we want the party to be always a well known party? Will we need anonymous parties? @MrMatthewLayton mentioned account parties? 
What do you think @mnesbit ?

MrMatthewLayton: I've always had a few concerns with regards to the `Party` implementation in Corda 4:

**Naming**
The root of the type hierarchy is `AbstractParty`. As far as I'm aware from looking at some Java APIs, the `Abstract` prefix seems to occur **only** when you have an interface for something, and an abstract of that thing, where there would be a naming clash; for example:

```
interface List<T> { ... }

abstract class AbstractList<T> : List<T> { ... }

class ArrayList<T> : AbstractList<T> { ... }
```

Therefore the type hierarchy would have been better described as:

```
abstract class Party
 > class AnonymousParty : Party
 > class WellKnownParty : Party
```

**Extensibility**
I'm not sure whether there was ever an intention to extend `AbstractParty` outside of the platform itself. It almost feels like it should have been declared as `sealed` rather than `abstract`, however, allowing external extensibility has allowed me to build an account model that hooks right into the `AbstractParty` implementation...

https://github.com/onix-labs/onixlabs-corda-identity-framework/blob/main/onixlabs-corda-identity-framework-contract/src/main/kotlin/io/onixlabs/corda/identityframework/contract/accounts/AccountParty.kt

I'm aware that in Corda 5, there is a heavy push towards using interfaces rather than using abstract or sealed classes, however this introduces a trade-off between things that _may_ leak abstraction, and things that are protected from external implementation, and IIRC, sealed interfaces are still experimental.

**Conclusion**
As @blsemo has pointed out, 
 - Do we want anonymous parties? 
 - Do we want accounts that can be implemented as parties?
 - Do we want the ability for consumers to implement these interfaces externally?

ledger/ledger-consensual/src/main/kotlin/net/corda/v5/ledger/consensual/Party.kt
```diff
@@ -0,0 +1,16 @@
...
+    val name: MemberX500Name
+    val owningKey: PublicKey
+
+    fun description() = "$name (owningKey = ${owningKey.toStringShort()})"
```
MrMatthewLayton: Why do we implement things like this as a method/function, rather than as a computed property? i.e.

```
val description: String get() = "$name (owningKey = ${owningKey.toStringShort()})"
```

ledger/ledger-consensual/src/main/kotlin/net/corda/v5/ledger/consensual/transaction/ConsensualWireTransaction.kt
vlajos: I have removed this.

ledger/ledger-consensual/src/main/kotlin/net/corda/v5/ledger/consensual/Party.kt
vlajos: Thanks. I have changed it. (The original came with the DP1 codebase.)

ledger/ledger-consensual/src/main/kotlin/net/corda/v5/ledger/consensual/Party.kt
mnesbit: We need a design meeting on this, because identity is a difficult topic. I certainly don't think that Party is fit for purpose.

ledger/ledger-consensual/src/main/kotlin/net/corda/v5/ledger/consensual/transaction/ConsensualLedgerTransaction.kt
```diff
@@ -0,0 +1,16 @@
...
+import net.corda.v5.ledger.common.transactions.LedgerTransaction
+import net.corda.v5.ledger.consensual.ConsensualState
+import net.corda.v5.ledger.consensual.Party
+import java.time.Instant
+
+@DoNotImplement
+interface ConsensualLedgerTransaction: LedgerTransaction {
+    val metadata: ConsensualTransactionMetaData
+    val timestamp: Instant
+    val requiredSigners: List<Party>
```
mnesbit: Continuing the discussion of Party. For instance here we talk about signers but include a Party not a PublicKey, or hash of a PublicKey. The signature will only confirm the PublicKey it won't help with the name. Worse the name/key association is likely to shift with time as keys are rotated. It also fails to include account information, which might be needed. So personally I would like to make the list of signers just a list of PublicKey. However, people do require a way of categorising state/transactions by the network members involved and they also need to be able to validate notaries and issuers against MemberInfo details.

ledger/ledger-consensual/src/main/kotlin/net/corda/v5/ledger/consensual/Party.kt
blsemo: Let's go with this `Party` definition for now and revisit when I'm back. Everyone feel free to think about it...
I have raised https://r3-cev.atlassian.net/browse/CORE-5936 to tackle this in a few weeks time.

ledger/ledger-consensual/src/main/kotlin/net/corda/v5/ledger/consensual/transaction/ConsensualTransactionMetaData.kt
vlajos: I have removed it.

ledger/ledger-common/src/main/kotlin/net/corda/v5/ledger/common/transactions/TransactionMetaData.kt
vlajos: I have refactored it. LayeredPropertyMap seemed to be a good pattern for it.

ledger/ledger-consensual/src/main/kotlin/net/corda/v5/ledger/consensual/transaction/ConsensualTransactionBuilder.kt
vlajos: I have modified it according this.

ledger/ledger-common/src/main/kotlin/net/corda/v5/ledger/common/transactions/CpkIdentifier.kt
vlajos: I have opened https://r3-cev.atlassian.net/browse/CORE-5940 to address this.

ledger/ledger-consensual/src/main/kotlin/net/corda/v5/ledger/consensual/Party.kt
vlajos: I have moved Matthew's response here: https://r3-cev.atlassian.net/browse/CORE-5936

ledger/ledger-consensual/src/main/kotlin/net/corda/v5/ledger/consensual/transaction/ConsensualLedgerTransaction.kt
vlajos: I have moved this comment here: https://r3-cev.atlassian.net/browse/CORE-5936

ledger/ledger-common/src/main/kotlin/net/corda/v5/ledger/common/transactions/TransactionMetaData.kt
```diff
@@ -0,0 +1,15 @@
...
+@DoNotImplement
+@CordaSerializable
+interface TransactionMetaData{
+    operator fun get(key: String): Any?
+    val entries: Set<Map.Entry<String, Any>>
+}
+
+const val TRANSACTION_META_DATA_LEDGER_MODEL_KEY = "ledgerModel"
+const val TRANSACTION_META_DATA_LEDGER_VERSION_KEY = "ledgerVersion"
+const val TRANSACTION_META_DATA_CPK_IDENTIFIERS_KEY = "cpkIdentifiers"
```
omfR3: You could omit the `TRANSACTION_META_DATA_` part and put them in a companion object inside `TransactionMetaData` maybe?

```kotlin
interface TransactionMetaData{
    operator fun get(key: String): Any?
    val entries: Set<Map.Entry<String, Any>>

    companion object {
      const val LEDGER_MODEL_KEY = "ledgerModel"
      const val LEDGER_VERSION_KEY = "ledgerVersion"
      const val CPK_IDENTIFIERS_KEY = "cpkIdentifiers"
    }
}
```

Then the constants would be namespaced under `TransactionMetaData`.

ledger/ledger-common/src/main/kotlin/net/corda/v5/ledger/common/transactions/WireTransaction.kt
```diff
@@ -0,0 +1,19 @@
...
+@DoNotImplement
+@CordaSerializable
+interface WireTransaction {
+    val id: SecureHash
+    val privacySalt: PrivacySalt
+    val componentGroupLists: List<List<ByteArray>>
+
+    fun getComponentGroupList(componentGroupId: Int): List<ByteArray>
+
+    fun getWrappedLedgerTransactionClassName(): String
```
omfR3: ^ Is this needed for deserialisation later?

ledger/ledger-common/src/main/kotlin/net/corda/v5/ledger/common/transactions/LedgerTransaction.kt
```diff
@@ -0,0 +1,16 @@
...
+import net.corda.v5.base.annotations.CordaSerializable
+import net.corda.v5.base.annotations.DoNotImplement
+import net.corda.v5.crypto.SecureHash
+import java.security.PublicKey
+
+@DoNotImplement
+@CordaSerializable // TODO(???)
+interface LedgerTransaction {
+    val id: SecureHash
+    val privacySalt: PrivacySalt
```
omfR3: Why do we have a `privacySalt` here but also within `wireTransaction`? Are they the same?

ledger/ledger-consensual/src/main/kotlin/net/corda/v5/ledger/consensual/transaction/ConsensualTransactionBuilder.kt
```diff
@@ -0,0 +1,29 @@
...
+    val digestService: DigestService
+    val secureRandom: SecureRandom
+    val serializer: SerializationService
+    val signingService: SigningService
+
+    val timeStamp: Instant?
+    fun withTimeStamp(timeStamp: Instant): ConsensualTransactionBuilder
+
+    val consensualStates: List<ConsensualState>
+    fun withConsensualState(consensualState: ConsensualState) : ConsensualTransactionBuilder
```
omfR3: What do I get if I call `withTimeStamp` or `withConsensualState` -- is this a builder method thing and they return a new `ConsensualTransactionBuilder` copy with those properties added?

ledger/ledger-consensual/src/test/java/net/corda/v5/ledger/consensual/ConsensualStatesJavaApiTest.java
```diff
@@ -0,0 +1,48 @@
...
+        Assertions.assertThat(participants).isEqualTo(participants1);
+    }
+
+    @Test
+    public void customConsensualStateClass() {
+        final CustomConsensualState customConsensualState = new CustomConsensualState(participants);
+
+        Assertions.assertThat(customConsensualState).isNotNull();
+        Assertions.assertThat(customConsensualState.getParticipants()).isEqualTo(participants);
+    }
```
omfR3: Seems like we're just testing the mocks here.

ledger/ledger-common/src/main/kotlin/net/corda/v5/ledger/common/transactions/WireTransaction.kt
vlajos: We'll need something to know what is wrapped inside the wireTx which deserializes componentgroup 0,0 and gets the ledgerModel which will be the classname.

ledger/ledger-common/src/main/kotlin/net/corda/v5/ledger/common/transactions/LedgerTransaction.kt
vlajos: Yes. I was not too sure about it. I have just removed it for the time being.

ledger/ledger-consensual/src/main/kotlin/net/corda/v5/ledger/consensual/transaction/ConsensualTransactionBuilder.kt
vlajos: Yes. https://github.com/corda/corda-runtime-os/pull/1715/files#diff-056e4bffcd08471f085b7fb5255f417ac87f930b4b8bd0552db5e638fa9c2513R61
(Although it is half refactored already to have the services as constructor arguments and return a signed tx, but the concept is roughly the same.)

ledger/ledger-consensual/src/test/java/net/corda/v5/ledger/consensual/ConsensualStatesJavaApiTest.java
vlajos: I have not dug yet too deeply in the tests. I ported what I found in the old code. It is possible that with the later it became pointless... :) I'll check it later.

ledger/ledger-common/src/main/kotlin/net/corda/v5/ledger/common/transactions/TransactionMetaData.kt
vlajos: Thanks. Good idea. I have refactored it.

ledger/ledger-consensual/src/main/kotlin/net/corda/v5/ledger/consensual/transaction/ConsensualTransactionBuilder.kt
vlajos: Thanks. Done.

ledger/ledger-common/src/main/kotlin/net/corda/v5/ledger/common/transactions/LedgerTransaction.kt
```diff
@@ -0,0 +1,12 @@
+package net.corda.v5.ledger.common.transactions
+
+import net.corda.v5.base.annotations.DoNotImplement
+import net.corda.v5.crypto.SecureHash
+import java.security.PublicKey
+
+@DoNotImplement
+interface LedgerTransaction {
+    val id: SecureHash
+    val wireTransaction: WireTransaction // TODO(to be removed if wireTx gets private)
```
mnesbit: I don't think we need this. We don't convert LedgerTransaction back to WireTransaction

ledger/ledger-common/src/main/kotlin/net/corda/v5/ledger/common/transactions/SignedTransaction.kt
```diff
@@ -0,0 +1,33 @@
...
+import net.corda.v5.application.crypto.DigitalSignatureAndMetadata
+import net.corda.v5.base.annotations.CordaSerializable
+import net.corda.v5.base.annotations.DoNotImplement
+import java.security.PublicKey
+
+@DoNotImplement
+@CordaSerializable
+interface SignedTransaction {
+    val wireTransaction: WireTransaction
+    val ledgerTransaction: LedgerTransaction //TODO(do we need this? Or rather delegate some properties through?)
```
mnesbit: We should not have a property for ledgerTransaction as the conversion process is complicated and contextual and this might easily get serialized by mistake. We can have a function that returns a LedgerTransaction on demand and makes it clear that it is not for serialization

ledger/ledger-common/src/main/kotlin/net/corda/v5/ledger/common/transactions/SignedTransaction.kt
```diff
@@ -0,0 +1,33 @@
...
+    val requiredSigningKeys: List<PublicKey> //TODO(was a set previously.)
+
+    /** Returns the same transaction but with an additional (unchecked) signature. */
+    fun withAdditionalSignature(sig: DigitalSignatureAndMetadata): SignedTransaction
+
+    /** Returns the same transaction but with an additional (unchecked) signatures. */
+    fun withAdditionalSignatures(sigList: Iterable<DigitalSignatureAndMetadata>): SignedTransaction
+
+    /** Alias for [withAdditionalSignature] to let you use Kotlin operator overloading. */
+    operator fun plus(sig: DigitalSignatureAndMetadata) = withAdditionalSignature(sig)
```
mnesbit: These are not Java API compatible. We might be able to do them as extension methods, but they should not be on the api interface

ledger/ledger-common/src/main/kotlin/net/corda/v5/ledger/common/transactions/LedgerTransaction.kt
vlajos: Thanks. I have moved it to private/runtime-os.

ledger/ledger-common/src/main/kotlin/net/corda/v5/ledger/common/transactions/SignedTransaction.kt
vlajos: Thanks. Done.

ledger/ledger-common/src/main/kotlin/net/corda/v5/ledger/common/transactions/SignedTransaction.kt
vlajos: Did you mean something like this?
https://github.com/corda/corda-api/pull/487/commits/b3c3c9e41a40f0c54fd9b1b74bf3a0d77a731c4e
Can this stay in corda-api? Or does it need to go to runtime-os?

ledger/ledger-common/src/main/kotlin/net/corda/v5/ledger/common/transactions/LedgerTransaction.kt
```diff
@@ -0,0 +1,11 @@
+package net.corda.v5.ledger.common.transactions
+
+import net.corda.v5.base.annotations.DoNotImplement
+import net.corda.v5.crypto.SecureHash
+import java.security.PublicKey
+
+@DoNotImplement
+interface LedgerTransaction {
```
mnesbit: I am not sure we should create common interfaces at this stage. I think it is quite limiting to force ledgers to have any commonality at the data model level. Lets just stick to Consensual ledger

ledger/ledger-common/src/main/kotlin/net/corda/v5/ledger/common/transactions/TransactionMetaData.kt
```diff
@@ -0,0 +1,17 @@
+package net.corda.v5.ledger.common.transactions
+
+import net.corda.v5.base.annotations.CordaSerializable
+import net.corda.v5.base.annotations.DoNotImplement
+
+@DoNotImplement
+@CordaSerializable
+interface TransactionMetaData{
```
mnesbit: I think the meta will be an internal concern at least for now, so probably should start out in corda-runtime-os. The same is probably true for PrivacySalt, but that is tied into the legacy stuff so we should leave here

ledger/ledger-common/src/main/kotlin/net/corda/v5/ledger/common/transactions/LedgerTransaction.kt
vlajos: OK. I have removed the common one.

ledger/ledger-common/src/main/kotlin/net/corda/v5/ledger/common/transactions/TransactionMetaData.kt
vlajos: Thanks. I have moved TransactionMetadata into runtime-os.

mnesbit (dismissed):
This seems like a sensible point to start with

ledger/ledger-consensual/src/test/java/net/corda/v5/ledger/consensual/ConsensualStatesJavaApiTest.java
vlajos: Thanks. Good catch.
I have fixed it.

ledger/ledger-consensual/src/main/kotlin/net/corda/v5/ledger/consensual/Party.kt
```diff
@@ -1,29 +1,35 @@
...
      */
     val description: String get() = "$name (owningKey = ${owningKey.toStringShort()})"
+
+    override fun toString() = name.toString()
+    override fun equals(other: Any?): Boolean =
+        other === this ||
+                other is Party &&
+                other.owningKey == owningKey &&
+                other.name == name
+    override fun hashCode(): Int = owningKey.hashCode()
```
MrMatthewLayton: @mnesbit Would there ever be a case where `Party` might be constructed with different `MemberX500Name`s, but the same `PublicKey`?

Just wondering if `hashCode` should account for the `name` property as well; for example:

```kotlin
override fun hashCode(): Int = Objects.hash(name, owningKey)
```

vlajos: I think I got that part from here:
https://github.com/corda/corda-runtime-os/blob/release/os/5.0/libs/membership/ledger-identity/src/main/kotlin/net/corda/membership/ledger/identity/PartyImpl.kt#L25

ledger/ledger-common/src/main/java/net/corda/v5/ledger/common/package-info.java
```diff
@@ -0,0 +1,4 @@
+@Export
+package net.corda.v5.ledger.common;
```
JamesHR3: I think you only have types in `ledger.common.transaction`, so at the moment you don't need to export this?

ledger/ledger-common/src/main/kotlin/net/corda/v5/ledger/common/transaction/PrivacySalt.kt
```diff
@@ -0,0 +1,12 @@
+package net.corda.v5.ledger.common.transaction
+
+import net.corda.v5.base.annotations.CordaSerializable
+
+/**
+ * A privacy salt is required to compute nonces per transaction component in order to ensure that an adversary cannot
+ * use brute force techniques and reveal the content of a Merkle-leaf hashed value.
+ */
+@CordaSerializable
```
JamesHR3, line 10: I don't think marking an interface as `@CordaSerializable` actually does anything, as the annotation isn't inherited by the class implementations. (The annotation is meta-annotated as `@Inherited` but my understanding is that that only applies to classes and not interfaces.)

ledger/ledger-common/src/main/kotlin/net/corda/v5/ledger/common/transaction/PrivacySalt.kt
```diff
@@ -0,0 +1,12 @@
...
+interface PrivacySalt{
```
JamesHR3, line 12: Do we expect users to implement this?

ledger/ledger-common/src/main/kotlin/net/corda/v5/ledger/common/transaction/PrivacySalt.kt
```diff
@@ -0,0 +1,12 @@
...
+    val bytes: ByteArray
```
JamesHR3, line 16: The property should have a kdoc.

ledger/ledger-consensual/src/main/kotlin/net/corda/v5/ledger/consensual/ConsensualState.kt
```diff
@@ -0,0 +1,37 @@
...
+     * This needs to be provided to let the State verify its internal state and other parts of the encapsulating
+     * Transaction.
+     * TODO(make services injectable (crypto, etc... CORE-5995)
+     *
+     * @param ledgerTransaction encapsulating transaction
+     *
+     * @return result of the verification
+     *
+     */
+    fun verify(ledgerTransaction: ConsensualLedgerTransaction): Boolean
```
JamesHR3: This feels a bit odd to me, but that may be because I don't have a good enough internal model of the consensual ledger. What happens if you give this verify function a transaction that's unrelated to the state you're trying to verify for example? Do you need to be able to retrieve the correct transaction in order to get a valid result from this? If so, how do you do that?

ledger/ledger-consensual/src/main/kotlin/net/corda/v5/ledger/consensual/transaction/ConsensualLedgerTransaction.kt
```diff
@@ -0,0 +1,52 @@
...
+
+/**
+ * Defines a Consensual ledger transaction.
+ *
+ * Comparing with [ConsensualSignedTransaction]:
+ *  - It has access to the deserialized details.
+ *  - It does not have direct access to the signatures.
+ *  - It requires a serializer.
+ *
+ *  //TODO(below docs are from obsolete part, review them.)
```
JamesHR3: Are you intending to review these docs in this PR or in a follow up?

ledger/ledger-consensual/src/main/kotlin/net/corda/v5/ledger/consensual/transaction/ConsensualTransactionBuilder.kt
```diff
@@ -0,0 +1,54 @@
...
+ * It is a transaction class that's mutable (unlike the others which are all immutable). It is
+ * intended to be passed around parties that may edit it by adding new states. Then, once the states
+ * are right, this class can be used as a holding bucket to gather signatures from multiple parties.
+ */
+@DoNotImplement
+interface ConsensualTransactionBuilder {
+    /**
+     * @property timestamp The timestamp of the [ConsensualTransactionBuilder].
+     */
+    val timestamp: Instant?
```
JamesHR3: Should this be nullable?

What does this instant represent? The time at which the builder was created? Does that define a timestamp placed on the transaction itself?

ledger/ledger-consensual/src/main/kotlin/net/corda/v5/ledger/consensual/transaction/ConsensualTransactionBuilder.kt
```diff
@@ -0,0 +1,54 @@
...
+
+    /**
+     * Sets the specific timestamp to the [ConsensualTransactionBuilder].
+     *
+     * @param timestamp The timestamp to be set to the current [ConsensualTransactionBuilder].
+     * @return Returns a new [ConsensualTransactionBuilder] with the specified timestamp.
+     */
+    fun withTimestamp(timestamp: Instant): ConsensualTransactionBuilder
```
JamesHR3: Ah, so we're allowing a timestamp to be set by the user. I'm not certain that we should allow that - can the user do something mad by setting a timestamp far in the past or similar? It's hard to say whether this API makes sense without having an indication of what the timestamp will be used for.

ledger/ledger-consensual/src/main/kotlin/net/corda/v5/ledger/consensual/transaction/ConsensualTransactionBuilder.kt
```diff
@@ -0,0 +1,54 @@
...
+
+    /**
+     * @property states The output [ConsensualState]s of the [ConsensualTransactionBuilder].
+     */
+    val states: List<ConsensualState>
```
JamesHR3, line 19: I wonder how useful having these read only properties for what has been added to the builder actually is - I would have thought a typical use case would call `withTimestamp` once, `withState` a bunch of times and then `signInitial` to complete the building of the tx.

ledger/ledger-consensual/src/main/kotlin/net/corda/v5/ledger/consensual/transaction/ConsensualTransactionBuilder.kt
```diff
@@ -0,0 +1,54 @@
...
+
+    /**
+     * Adds the specified [ConsensualState] to the [ConsensualTransactionBuilder].
+     *
+     * @param state The state of the output to add to the current [ConsensualTransactionBuilder].
+     * @return Returns a new [ConsensualTransactionBuilder] with the specified output state.
+     */
+    fun withState(state: ConsensualState) : ConsensualTransactionBuilder
```
JamesHR3: Should we allow adding a list of states?

ledger/ledger-consensual/src/main/kotlin/net/corda/v5/ledger/consensual/transaction/ConsensualTransactionBuilder.kt
```diff
@@ -0,0 +1,54 @@
...
+    /**
+     * 1. Verifies the content of the [ConsensualTransactionBuilder]
+     * 2.a Creates
+     * 2.b signs
+     * 2.c and returns a [ConsensualSignedTransaction]
+     *
+     * @param publicKey The key for signing the [ConsensualSignedTransaction].
+     * @return Returns a new [ConsensualSignedTransaction] with the specified details.
+     */
+    fun signInitial(publicKey: PublicKey): ConsensualSignedTransaction
```
JamesHR3, line 45: Not sure if we want to go down this route, but we could do this as `withSigningKey` followed by `build`, which triggers the first signature to be generated and added. That approach might be better if we ever anticipate allowing multiple signers from the same vnode. I think I'd need to see more about the model to understand whether that's a good idea.

ledger/ledger-consensual/src/main/kotlin/net/corda/v5/ledger/consensual/transaction/ConsensualSignedTransaction.kt
```diff
@@ -0,0 +1,79 @@
...
+     */
+    fun addSignature(signature: DigitalSignatureAndMetadata): ConsensualSignedTransaction
+
+    /**
+     * Adds the specified signatures to the current [ConsensualSignedTransaction].
+     *
+     * @param signatures The signatures and metadata to add to the current [ConsensualSignedTransaction].
+     * @return Returns a new [ConsensualSignedTransaction] containing the applied signatures.
+     */
+    fun addSignatures(signatures: Iterable<DigitalSignatureAndMetadata>): ConsensualSignedTransaction
```
JamesHR3: These APIs feel a bit strange to me. In the builder we ask for a public key and then generate a signed transaction (presumably with that signature attached), but then if you want to add more you do it by generating a signature object and putting it on the tx. Can you add a signature over some other data for example? Is that useful?

ledger/ledger-consensual/src/main/kotlin/net/corda/v5/ledger/consensual/transaction/ConsensualSignedTransaction.kt
```diff
@@ -0,0 +1,79 @@
...
+
+    /**
+     * Gets the signing keys for any missing transaction signatures.
+     *
+     * @param serializer The [SerializationService] required to obtain the missing keys.
+     * @return Returns a [Set] of [PublicKey] representing the signing keys for any missing transaction signatures.
+     */
+    fun getMissingSigningKeys(serializer: SerializationService): Set<PublicKey>
```
JamesHR3: The builder requires a list of consensual states to be provided before the transaction is generated, which are presumably not serialized until we generate the transaction itself. Can we extract a list of required signers at build time and prevent the need for a serialization round trip to extract them?

ledger/ledger-consensual/src/main/kotlin/net/corda/v5/ledger/consensual/Party.kt
```diff
@@ -0,0 +1,29 @@
...
+
+    /**
+     * @property owningKey The [PublicKey] owned by the party.
+     */
+    val owningKey: PublicKey
+
+    /**
+     * Concise, readable String representation of a Party.
+     */
+    val description: String get() = "$name (owningKey = ${owningKey.toStringShort()})"
```
JamesHR3: Are we better off just overriding `toString` on the implementation?

JamesHR3 (commented):
Some thoughts. I recognize that we're porting the API over from C4 to an extent, but I wonder if we can make some tweaks to make the API more usable without having to do serious design work, as we're keeping the concepts the same?

ledger/ledger-common/src/main/java/net/corda/v5/ledger/common/package-info.java
vlajos: Thanks. I have removed it.

ledger/ledger-common/src/main/kotlin/net/corda/v5/ledger/common/transaction/PrivacySalt.kt
vlajos: @JamesHR3 
Hmm. 
If I remove `@CordaSerializable` from `Party`. https://github.com/corda/corda-api/pull/487/files#diff-220f68872e64df2e2d86100c1110074e2e35665a0d012096a095a16e87cf675aR13
Whose implementation does not have it either. https://github.com/corda/corda-runtime-os/pull/1715/files#diff-5952c829255e7e76a07adc3fe7ae1a9c4a736fbfa54c4cd151cee4167aa05661R7
This test: https://github.com/corda/corda-runtime-os/pull/1715/files#diff-f4502fe9888db2f9bd6bfec08904c1ec06978d53e18ab85845626fcbf5ac2363R71
fails with this:
```
Class "interface net.corda.v5.ledger.consensual.Party" is not annotated with @CordaSerializable.
java.io.NotSerializableException: Class "interface net.corda.v5.ledger.consensual.Party" is not annotated with @CordaSerializable.
...
```
Otherwise it works.
So it looks like doing something in a similar case.

ledger/ledger-common/src/main/kotlin/net/corda/v5/ledger/common/transaction/PrivacySalt.kt
vlajos: No. Fixed.

ledger/ledger-common/src/main/kotlin/net/corda/v5/ledger/common/transaction/PrivacySalt.kt
vlajos: Thanks. Added.

ledger/ledger-consensual/src/main/kotlin/net/corda/v5/ledger/consensual/ConsensualState.kt
vlajos: Details are not super clear/final yet.
We agreed before Christian's holiday, that we'll need a verify function which will need access to the Transaction and probably some services.
We have this ticket to carry-on with the details: https://r3-cev.atlassian.net/browse/CORE-5982

ledger/ledger-consensual/src/main/kotlin/net/corda/v5/ledger/consensual/transaction/ConsensualTransactionBuilder.kt
vlajos: > Should this be nullable?

It is null only until it gets filled populated with the next function

> What does this instant represent? 

It will be the transaction's timestamp.

ledger/ledger-consensual/src/main/kotlin/net/corda/v5/ledger/consensual/transaction/ConsensualTransactionBuilder.kt
vlajos: Based on my understanding, we try to be quite flexible in the Consensual world.
If any parties do not like the timestamp, they won't sign it.

ledger/ledger-consensual/src/main/kotlin/net/corda/v5/ledger/consensual/transaction/ConsensualTransactionBuilder.kt
vlajos: I do not have strong opinion about them.
They came from @mnesbit 's verifier.
@mnesbit what do you think?

ledger/ledger-common/src/main/kotlin/net/corda/v5/ledger/common/transaction/PrivacySalt.kt
vlajos: (Since yesterday `Party` got a custom serialiser. Hopefully the links survive the related pushes...)

ledger/ledger-consensual/src/main/kotlin/net/corda/v5/ledger/consensual/Party.kt
vlajos: I think I got it from here:
https://github.com/corda/corda5/blob/release/os/5.0/application/src/main/kotlin/net/corda/v5/application/identity/Party.kt#L37

Original `toString()` (https://github.com/corda/corda-runtime-os/pull/1715/files#diff-c174eb8fa66f7c7ee5d09ba7b732751c4ec66ff6e802b892cb5000bcbf7f8c75L20)
was just like this in the previous PartyImpl:
`override fun toString() = name.toString()`

Although compatibility/similarity with C4/DP1 is not a big issue in Consensual world.

Anyway I do not have strong feelings.
Should I change it?

ledger/ledger-consensual/src/main/kotlin/net/corda/v5/ledger/consensual/transaction/ConsensualTransactionBuilder.kt
vlajos: Thanks.
Good idea. I have changed to vararg.

ledger/ledger-consensual/src/main/kotlin/net/corda/v5/ledger/consensual/transaction/ConsensualSignedTransaction.kt
vlajos: These functions came from here:
https://github.com/corda/corda5/blob/40b67c5cff50c69a1e1b792ad11b8d7d2acfbccc/ledger/ledger/src/main/kotlin/net/corda/v5/ledger/transactions/SignedTransaction.kt#L45

But I think you are right and gathering the signatures in different ways does not sound consistent.

(I am not too sure about the answer to your last questions.)

ledger/ledger-consensual/src/main/kotlin/net/corda/v5/ledger/consensual/transaction/ConsensualSignedTransaction.kt
vlajos: Where the states are available the signers can be accessed too. (We planned to calculate them from the states's participants's owningKeys.)
I thought this function's purpose was more useful for verifying if there are any missing signatures.

ledger/ledger-consensual/src/main/kotlin/net/corda/v5/ledger/consensual/transaction/ConsensualLedgerTransaction.kt
vlajos: I have removed these and opened CORE-6251 for a later review.

ledger/ledger-consensual/src/main/kotlin/net/corda/v5/ledger/consensual/transaction/ConsensualSignedTransaction.kt
vlajos: @JamesHR3 suggested on our call to replace addSignature with something like this:
```
fun sign(publickey: ...)
```
to be more consistent with `signInitial()`

ledger/ledger-consensual/src/main/kotlin/net/corda/v5/ledger/consensual/Party.kt
vlajos: I have changed it.

ledger/ledger-consensual/src/main/kotlin/net/corda/v5/ledger/consensual/Party.kt
vlajos: (These changes got reverted here.)

ledger/ledger-consensual/src/main/kotlin/net/corda/v5/ledger/consensual/transaction/ConsensualTransactionBuilder.kt
vlajos: We agreed to keep it.

ledger/ledger-consensual/src/main/kotlin/net/corda/v5/ledger/consensual/transaction/ConsensualTransactionBuilder.kt
vlajos: We agreed that the function stays as it is.

I added these to its doc block:
```
     * Calling this function once renders the current [ConsensualTransactionBuilder] immutable.
```
...
```
     *
     * @throws [UnsupportedOperationException] when called second time on the same object to prevent duplicate
     *      transactions accidentally.
```

ledger/ledger-consensual/src/main/kotlin/net/corda/v5/ledger/consensual/ConsensualState.kt
vlajos: I have changed to throw instead of returning Boolean and added some docs.

ledger/ledger-consensual/src/main/kotlin/net/corda/v5/ledger/consensual/transaction/ConsensualTransactionBuilder.kt
vlajos: I have removed timestamp and withTimestamp.
They will be set in `signInitial()` calling time.

ledger/ledger-consensual/src/main/kotlin/net/corda/v5/ledger/consensual/transaction/ConsensualTransactionBuilder.kt
vlajos: I have removed timestamp and withTimestamp.
They will be set in `signInitial()` calling time.

ledger/ledger-consensual/src/main/kotlin/net/corda/v5/ledger/consensual/transaction/ConsensualSignedTransaction.kt
vlajos: I have modified to `fun addSignature(publicKey: PublicKey)` and removed the plural versions.

ledger/ledger-consensual/src/main/kotlin/net/corda/v5/ledger/consensual/transaction/ConsensualSignedTransaction.kt
vlajos: I have removed the serializer argument.

ledger/ledger-consensual/src/main/kotlin/net/corda/v5/ledger/consensual/ConsensualState.kt
```diff
@@ -0,0 +1,41 @@
...
+     *  - check the transaction's timestamp.
+     *
+     * TODO(make services injectable (crypto, etc... CORE-5995)
+     *
+     * @param ledgerTransaction encapsulating transaction
+     *
+     * @throws [IllegalArgumentException] if the verification failed.
+     *
+     */
+    fun verify(ledgerTransaction: ConsensualLedgerTransaction): Unit
```
mnesbit: Don't need the explicit Unit

JamesHR3: You can leave `Unit` off the end I think.

ledger/ledger-consensual/src/main/kotlin/net/corda/v5/ledger/consensual/transaction/ConsensualSignedTransaction.kt
```diff
@@ -0,0 +1,67 @@
...
+    /**
+     * Gets the signing keys for any missing transaction signatures.
+     *
+     * @return Returns a [Set] of [PublicKey] representing the signing keys for any missing transaction signatures.
+     */
+    fun getMissingSigningKeys(): Set<PublicKey>
+}
+
+/** Alias for [ConsensualSignedTransaction.addSignature] to let you use Kotlin operator overloading. */
+operator fun ConsensualSignedTransaction.plus(publicKey: PublicKey) = addSignature(publicKey)
```
mnesbit: Feels a bit likely to confuse. Think we can live without.

ledger/ledger-consensual/src/main/kotlin/net/corda/v5/ledger/consensual/ConsensualState.kt
```diff
@@ -0,0 +1,41 @@
...
+     *  - verify the state's well-formedness
+     *  - verify compatibility with the other states of the encapsulating transaction
+     *  - check required signing keys
+     *  - check the transaction's timestamp.
+     *
+     * TODO(make services injectable (crypto, etc... CORE-5995)
+     *
+     * @param ledgerTransaction encapsulating transaction
+     *
+     * @throws [IllegalArgumentException] if the verification failed.
```
JamesHR3: Is `IllegalArgumentException` what we want here, or should we throw something verification specific?

ledger/ledger-consensual/src/main/kotlin/net/corda/v5/ledger/consensual/transaction/ConsensualTransactionBuilder.kt
JamesHR3: Not sure immutable is quite the right word, it's more like it consumes the builder so it cannot be used again. Therefore if you want to build two transactions you need two builders.

JamesHR3 (commented):
Minor bits of tidying up, but looks good otherwise.

ledger/ledger-consensual/src/main/kotlin/net/corda/v5/ledger/consensual/ConsensualState.kt
vlajos: Thanks. I have removed it.

ledger/ledger-consensual/src/main/kotlin/net/corda/v5/ledger/consensual/transaction/ConsensualSignedTransaction.kt
vlajos: Thanks. I have removed it.

ledger/ledger-consensual/src/main/kotlin/net/corda/v5/ledger/consensual/ConsensualState.kt
vlajos: Thanks. I have removed it.

ledger/ledger-consensual/src/main/kotlin/net/corda/v5/ledger/consensual/transaction/ConsensualTransactionBuilder.kt
vlajos: I have updated that.

ledger/ledger-consensual/src/main/kotlin/net/corda/v5/ledger/consensual/ConsensualState.kt
vlajos: I have ported the main [TransactionVerificationException] class from the obsolete module.

ledger/ledger-common/src/main/kotlin/net/corda/v5/ledger/common/transaction/TransactionVerificationException.kt
```diff
@@ -0,0 +1,14 @@
...
+import net.corda.v5.crypto.SecureHash
+
+/**
+ * Indicates that some aspect of the transaction named by [txId] violates the platform rules. The exact type of failure
+ * is expressed using a subclass. TransactionVerificationException is a [FlowException] and thus when thrown inside
+ * a flow, the details of the failure will be serialised, propagated to the peer and rethrown.
+ *
+ * @property txId the Merkle root hash (identifier) of the transaction that failed verification.
+ */
+abstract class TransactionVerificationException(val txId: SecureHash, message: String, cause: Throwable?)
```
mnesbit: Do we want this to be abstract? I presume we are throwing some (undeclared) descendants of this, which doesn't seem very friendly. Think we should keep the number of specialized exceptions down to those which you might have an application reason to differentiate.

vlajos: @mnesbit 
The DP1 implementation has 21 subclasses of TransactionVerificationException.
Do we want to discard that hierarchy?
I thought with adding only their ancestor class, we keep the option to populate the childs as we go.

mnesbit (approved):
OK

JamesHR3 (approved):

ledger/ledger-common/src/main/kotlin/net/corda/v5/ledger/common/transaction/TransactionVerificationException.kt
vlajos: I have changed the exception to concrete from abstract.

PR #488, opened 2022-07-25
CORE-5893 - Add missing response type for group policy query

nikinagy:
https://r3-cev.atlassian.net/browse/CORE-5893


charlieR3 (approved):
LGTM

PR #489, opened 2022-07-25
CORE-5897 - Move logicalType property in the right place

dimosr:
The `logicalType` property was in the wrong place in the p2p schemas causing each of the processors to write a WARN log complaining about it e.g.:
```
[pod/corda-membership-worker-57879bd496-wbzlg/corda-membership-worker] 10:54:02.870 [main] WARN  org.apache.avro.Schema - Ignored the net.corda.p2p.app.AuthenticatedMessageHeader.ttl.logicalType property ("timestamp-millis"). It should probably be nested inside the "type" for the field.
```

Moving it in the right place to remove the warning.

This requires a version bump, because adding the `logicalType` in the right place turns the TTL field from a `Long` to an `Instant`.

## Testing
I confirmed the warning has been removed checking the logs of the workers in the e2e tests for the `corda-runtime-os` [build](https://github.com/corda/corda-runtime-os/pull/1725).


williamvigorr3 (approved):
LGTM

PR #492, opened 2022-07-27
NOTICK: Dependency version upgrade for Jackson, JUnit, Hibernate, Detekt and the retry plugin

JamesHR3:
This PR contains the following dependency version upgrades:
- Hibernate 5.6.9.FINAL -> 5.6.10.FINAL
- Jackson 2.13.1 -> 2.13.3
- JUnit 5.8.2 -> 5.9.0
- Detekt 1.20 -> 1.21
- Retry plugin 1.3.2 -> 1.4.0

The detekt plugin upgrade has resulted in some legacy code flagging up new warnings, as well as general warnings being reported for incorrectly declared detekt configuration. These have been addressed.

### DIFF
--- a/detekt-config.yml
+++ b/detekt-config.yml
@@ -185,11 +185,11 @@ style:
   ForbiddenComment:
     active: true
     excludes: "**/buildSrc/**"
-    values: 'TODO:,FIXME:,STOPSHIP:'
+    values: ["TODO:","FIXME:","STOPSHIP:"]
   MagicNumber:
     active: false
     excludes: "**/test/**,**/integration-test/**,**/*Test.kt,**/*Tests.kt,**/buildSrc/**"
-    ignoreNumbers: '-1,0,1,2'
+    ignoreNumbers: ["-1","0","1","2"]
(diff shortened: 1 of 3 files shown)


alexey-kadyrov-r3:
Should you bump the API version? As some dependencies are declared in the runtime-os as well so there might be some version clashes?

JamesHR3:
> Should you bump the API version? As some dependencies are declared in the runtime-os as well so there might be some version clashes?

I think we already have this problem with Jackson - it's at 2.13.1 here and 2.13.2 in runtime-os. It seems that having it declared twice at different versions doesn't affect the build, but it would certainly be better if they were aligned. (I think gradle resolves the clashing dependencies problem for you by picking the higher one, although I don't know how/if it copes with major version skew.) I've just raised https://github.com/corda/corda-runtime-os/pull/1746 that will address the Jackson problem, though there may be others.

alexey-kadyrov-r3 (approved):
LGTM

ronanbrowne (approved):
LGTM

PR #493, opened 2022-07-27
Backup of original UTXO API

MrMatthewLayton:
Added data class `Amount<T>` to ledger/common and some of the unit tests (wip).


ledger/ledger-common/src/main/kotlin/net/corda/v5/ledger/common/Amount.kt
```diff
@@ -0,0 +1,366 @@
...
+        return times(value.quantity)
+    }
+
+    /**
+     * Divides the current [Amount] by the specified amount.
+     *
+     * @param value The amount to divide the current [Amount] by.
+     * @return Returns an [Amount] representing the quotient of the current quantity divided by the specified quantity.
+     */
+    operator fun div(value: BigInteger): Amount<T> {
```
mnesbit, line 239: I intentionally did not provide div in the original, because it won't conserve quantity, hence the splitEvenly.

ledger/ledger-common/src/main/kotlin/net/corda/v5/ledger/common/Amount.kt
```diff
@@ -0,0 +1,366 @@
...
+
+        @JvmStatic
+        fun fromCurrency(quantity: BigInteger, currency: Currency): Amount<Currency> {
+            return fromCurrency(quantity.toBigDecimal(currency.defaultFractionDigits), currency)
+        }
+
+        @JvmStatic
+        fun fromCurrency(quantity: BigDecimal, currency: Currency): Amount<Currency> {
+            val quantityScale = quantity.scale()
+            val currencyScale = currency.defaultFractionDigits
```
mnesbit, line 94: There are so many problems with Java Currency that I would be keen to drop this. There are numerous historical cases in which currencies were revalued, or renamed and Java just doesn't have accurate master reference data. That data would have to live on the ledger.

ledger/ledger-common/src/main/kotlin/net/corda/v5/ledger/common/Amount.kt
```diff
@@ -0,0 +1,366 @@
...
+ *
+ * The quantity of an amount is represented as an integral value, as opposed to a decimal value.
+ *
+ * @param T The underlying type of the item being quantified by the current amount.
+ * @property quantity The quantity of the item of the current amount.
+ * @property item The item being quantified by the current amount.
+ * @property isDivisible Determines whether the current amount can be subdivided into smaller amounts.
+ */
+@CordaSerializable
+data class Amount<T : Any>(val quantity: BigInteger, val item: T) : Comparable<Amount<T>> {
```
mnesbit, line 22: If we don't try to support to/from string representations and hence don't need to know scale and symbol then perhaps we don't need Amount. We just model it as BigInteger for the purposes of tracking fungible token counts.

ledger/ledger-common/src/main/kotlin/net/corda/v5/ledger/common/Amount.kt
MrMatthewLayton: Happy to remove `div` and `rem` if they are useless. Or we could keep them and make them private for now, and use them under the hood to implement `splitEvenly`? 

If then in the (rather unlikely) case that a customer asked why `Amount` is not divisible, we could either offer `splitEvenly`, or make `div` and `rem` public in Corda 5.x should there be any push for it.

ledger/ledger-common/src/main/kotlin/net/corda/v5/ledger/common/Amount.kt
MrMatthewLayton: Okay happy to drop.

ledger/ledger-common/src/main/kotlin/net/corda/v5/ledger/common/Amount.kt
MrMatthewLayton: I guess on balance, if we've (customers have) misunderstood the `T` property, this should perhaps be the first thing to go?

With regards to handling precision, if we were to drop `T`, then `Amount` just becomes a useless wrapper for `BigInteger` and could be removed entirely, but if we do want a safer way to handle precision, then `Amount` sort of feels like we're just trying to re-implement `BigDecimal` with some adjusted behaviour, and I'm not sure we want to be doing that either.

The reason I didn't do anything with precision in the initial implementation was because I came to the realisation that ultimately you just end up dealing with the smallest indivisible unit (i.e. 1 penny/cent/satoshi, etc.) so just allow precision to be handled as a display/presentation issue. The second reason was that some instruments subdivide into different things, and therefore the precision changes based on how you want to interpret the thing you're looking at; for example:

1.00000000 BTC = 1 BTC or 0.00000001 MBTC (MegaBitcoin)
0.00000100 BTC = 1.00 uBTC or 10.0 Finney or 100 Satoshi

mnesbit: Yes the C4 Amount was meant to allow easy conversion between display quantities and the indivisible tokens. It didn't really work that well as display is a complex topic. So it may well be we just use BigInteger. For me it is key that we look at the FungibleTokenState interface as that is where Amount is used. If that becomes: 
```
interface FungibleTokenState<T> :  ContractState {
   val tokenType: T
   val amount: BigInteger
   val issuer: Party // or PublicKey
   val owner: Party // or PublicKey
}
```

Then this does solve problems where issuer isn't a platform concept. Need to think about tokenType more as that is meant to represent different symbols/mixing rules, but that is left up to contract code

mnesbit: Note conservation of tokens is being left up to the contract verify, which does put burden on the cordapp developer

.run/Run Ledger Common Tests.run.xml
```diff
@@ -0,0 +1,12 @@
...
+    <module name="corda-api.ledger.ledger-common.test" />
+    <option name="PACKAGE_NAME" value="" />
+    <option name="MAIN_CLASS_NAME" value="AmountTests" />
+    <option name="METHOD_NAME" value="" />
+    <option name="TEST_OBJECT" value="class" />
+    <method v="2">
+      <option name="Make" enabled="true" />
+    </method>
+  </configuration>
+</component>
```
omfR3, line 12: Is this a new IntelliJ thingy?

ledger/ledger-utxo/src/main/kotlin/net/corda/v5/ledger/utxo/StateRef.kt
```diff
@@ -0,0 +1,42 @@
...
+        /**
+         * Parses the specified value into a [StateRef].
+         *
+         * @param value The value to parse.
+         * @return Returns a [StateRef] representing the specified value.
+         */
+        @JvmStatic
+        fun parse(value: String): StateRef {
+            val transactionHash = value.substringBefore(DELIMITER)
+            val index = value.substringAfterLast(DELIMITER)
```
r3domfox: Would suggest a regex here, both to check that the input value is well-formed (and throw an exception if it is not), and then to extract the sub-values from it.

ledger/ledger-utxo/src/main/kotlin/net/corda/v5/ledger/utxo/UtxoLedgerTransaction.kt
```diff
@@ -0,0 +1,9 @@
+package net.corda.v5.ledger.utxo
+
+/**
+ *
+ */
+interface UtxoLedgerTransaction {
+    fun <T : ContractState> getInputsOfType(type: Class<T>): List<T>
+    fun <T : ContractState> getOutputsOfType(type: Class<T>): List<T>
+}
```
r3domfox, line 299: Can we have elsewhere, as Kotlin helpers,

```kotlin
inline fun <reified T : ContractState> UtxoLedgerTransaction.getInputs(): List<T> = getInputsOfType(T::class.java)
inline fun <reified T : ContractState> UtxoLedgerTransaction.getOutputs(): List<T> = getOutputsOfType(T::class.java)
```

ledger/ledger-utxo/src/main/kotlin/net/corda/v5/ledger/utxo/UtxoSignedTransaction.kt
```diff
@@ -0,0 +1,54 @@
...
+     */
+    fun addSignature(signature: DigitalSignatureAndMetadata): UtxoSignedTransaction
+
+    /**
+     * Adds the specified signatures to the current [UtxoSignedTransaction].
+     *
+     * @param signatures The signatures and metadata to add to the current [UtxoSignedTransaction].
+     * @return Returns a new [UtxoSignedTransaction] containing the applied signatures.
+     */
+    fun addSignatures(signatures: Iterable<DigitalSignatureAndMetadata>): UtxoSignedTransaction
```
r3domfox, line 37: A vararg variant of this might be nice to have, and would probably make the single `addSignature` case redundant.

```kotlin
default fun addSignatures(vararg signatures: DigitalSignatureAndMetadata): UtxoSignedTransaction =
    addSignatures(signatures)
```

ledger/ledger-utxo/src/main/kotlin/net/corda/v5/ledger/utxo/UtxoTransactionBuilder.kt
```diff
@@ -0,0 +1,76 @@
...
+     */
+    fun withInput(stateRef: StateRef): UtxoTransactionBuilder
+
+    /**
+     * Adds the specified inputs to the UTXO transaction.
+     *
+     * @param stateRefs The state references of the inputs to add to the current [UtxoTransactionBuilder].
+     * @return Returns a new [UtxoTransactionBuilder] with the specified input states.
+     */
+    fun withInputs(stateRefs: Iterable<StateRef>): UtxoTransactionBuilder
```
r3domfox: Same here (and throughout) - default vararg funs which in turn call the fun taking an iterable would enable:

```
val myCommands = listOf(command1, command2)
builder.withInputs(input1)
   .withReferenceInputs(refInput1, refInput2)
   .withOutputs(output3)
    .withCommands(myCommands)
```

r3domfox (changes requested):
All looks plausible to me - please consider the vararg suggestion, but discard if it causes problems for Java callers, or default methods don't work with OSGi, or whatever.

ledger/ledger-utxo/src/main/kotlin/net/corda/v5/ledger/utxo/UtxoSignedTransaction.kt
```diff
@@ -0,0 +1,54 @@
...
+
+    /**
+     * Gets the signing keys for any missing transaction signatures.
+     *
+     * @param serializer The [SerializationService] required to obtain missing signed transaction signatures.
```
vlajos, line 42:
```suggestion
     * @param serializer The [SerializationService] required to obtain the missing keys.
```

MrMatthewLayton:
DO NOT DELETE! We may need this to refer back, beyond MVP.

PR #494, opened 2022-07-27
CORE-5389: add offset and limit to FindAll

dickon:
For now we are not supporting sorting. Pagination will be most useful in situations where the database
returns consistent order without us having to specify an ordering.


lankydan (dismissed):

driessamyn (approved):
LGTM

PR #500, opened 2022-08-02
CORE-5991 Flow Context propagation API readiness

simon-johnson-r3:
All Api changes required for Flow Worker Context Propagation


application/src/main/kotlin/net/corda/v5/application/flows/FlowContextProperties.kt
```diff
@@ -0,0 +1,45 @@
...
+    fun put(key: String, value: String)
+
+    /**
+     * Gets a value from the context property store.
+     * Also opens up [] operator access for getting values in Kotlin.
+     *
+     * @param key The property key
+     */
+    operator fun get(key: String): String?
+
```
simon-johnson-r3: Tested these methods are Java compliant, they are both callable just as "normal" methods.

application/src/main/kotlin/net/corda/v5/application/flows/FlowContextProperties.kt
```diff
@@ -0,0 +1,45 @@
...
+    /**
+     * Puts a user value into the context property store.
+     * Opens up [] operator access for setting values in Kotlin.
+     *
+     * @param key The property key
+     * @param value The property value
+     */
+    operator fun set(key: String, value: String)
```
lankydan: Do we need both `put` and `set`?

application/src/main/kotlin/net/corda/v5/application/flows/FlowContextProperties.kt
```diff
@@ -0,0 +1,45 @@
+package net.corda.v5.application.flows
+
+/**
+ * Context properties of a [Flow] are key value pairs of Strings. They are comprised of two sets of properties, those
+ * set by the platform when a [Flow] is started, and those which are added by the CorDapp developer during the execution
+ * of their flow. The latter set are referred to as user context properties. All context properties are immutable once
+ * set.
+ * Where keys exist as platform properties already, setting user properties is an effective no-o as they will always be
```
lankydan: "setting user properties is an effective no-o as they will always be", there is probably a better way to write this for public apis 😛

application/src/main/kotlin/net/corda/v5/application/flows/FlowContextProperties.kt
```diff
@@ -0,0 +1,45 @@
...
+     */
+    fun put(key: String, value: String)
+
+    /**
+     * Gets a value from the context property store.
+     * Also opens up [] operator access for getting values in Kotlin.
+     *
+     * @param key The property key
+     */
+    operator fun get(key: String): String?
```
lankydan, line 40: Should have a `@return`.

data/avro-schema/src/main/resources/avro/net/corda/data/flow/FlowStartContext.avsc
```diff
@@ -43,7 +43,18 @@
...
+      "doc": "The body of the request provided when the flow was started if this flow was started via RPC. Null otherwise."
+    },
+    {
+      "name": "contextPlatformProperties",
+      "type": ["null",
+        {
+          "type": "map",
+          "values": "string"
+        }
+      ],
```
lankydan: Can probably be formatted better.

data/avro-schema/src/main/resources/avro/net/corda/data/flow/event/session/SessionInit.avsc
```diff
@@ -26,6 +26,22 @@
       "type": ["null","string"],
       "doc": "Flow ID of the flow initiating the session. This is only required to inform the Flow Mapper. It should be null when sent to counterparties."
     },
+    {
+      "name": "contextUserProperties",
```
lankydan, line 33: So is the idea that the `FlowContextProperties` fills the `contextUserProperties` and then an internal API/access to the map alters `contextPlatformProperties`?

application/src/main/kotlin/net/corda/v5/application/flows/FlowContextProperties.kt
simon-johnson-r3: set() is a Kotlin operator which gives you [] =, but put() makes it more Java Map like, the latter being inspired by what you said in last Friday's demo :-D

application/src/main/kotlin/net/corda/v5/application/flows/FlowContextProperties.kt
simon-johnson-r3: typo, should have been no-op, but I assume you also mean that is inappropriate for the public api so I will change it

data/avro-schema/src/main/resources/avro/net/corda/data/flow/event/session/SessionInit.avsc
simon-johnson-r3: Yes exactly that. Platform properties are immutable as far as the user is concerned, they will contain things like "account", list TBD but outside the scope of context propagation implementation. User context is whatever the user wants it to be. Only internal apis can set platform properties, we don't expose the ability to do that to users.

application/src/main/kotlin/net/corda/v5/application/flows/FlowContextProperties.kt
lankydan: Does Java end up seeing both?

data/avro-schema/src/main/resources/avro/net/corda/data/flow/FlowStartContext.avsc
simon-johnson-r3: I'll run it through a formatter, the IDE wont do this for me unfortunately.

lankydan: There is a plugin for `avsc` files that I'm using which handles formatting. Intellij recommended it when I opened an `avsc` file.

application/src/main/kotlin/net/corda/v5/application/flows/FlowContextProperties.kt
simon-johnson-r3: Yes, if it didn't Java users wouldn't be able to see get() either. So the only potential change is to remove the Kotlin-isms here I think. The is only a user facing class so I have no concerns that it will affect us internally if we make it less Kotlin like. I don't mind either way in other words, happy to go with consensus. I guess we don't hide the fact the api is really Kotlin made Java compatible generally, so I'm not sure if this is any more confusing that other parts of the api for Java users or not.

data/avro-schema/src/main/resources/avro/net/corda/data/flow/FlowStartContext.avsc
simon-johnson-r3: I remember seeing that actually, will install, good idea.

application/src/main/kotlin/net/corda/v5/application/flows/FlowContextProperties.kt
lankydan: You should be able to use an extension function to bypass this. For example:

```kotlin
interface FlowContextProperties {
    /**
     * Puts a user value into the context property store.
     * If an existing user value exists anywhere in the current context stack, this value will overwrite it from this
     * point on down, but not up, the context stack.
     * If an existing platform value exists anywhere in the context stack, the value is ignored in preference for the
     * platform value.
     *
     * @param key The property key
     * @param value The property value
     */
    fun put(key: String, value: String)

    /**
     * Gets a value from the context property store.
     * Also opens up [] operator access for getting values in Kotlin.
     *
     * @param key The property key
     */
    operator fun get(key: String): String?
}

/**
 * Puts a user value into the context property store.
 * Opens up [] operator access for setting values in Kotlin.
 *
 * @param key The property key
 * @param value The property value
 */
operator fun FlowContextProperties.set(key: String, value: String) = put(key, value)

class FlowContextPropertiesImpl : FlowContextProperties {
    override fun put(key: String, value: String) {
        TODO("Not yet implemented")
    }

    override fun get(key: String): String? {
        TODO("Not yet implemented")
    }
}

fun main() {
    val map = FlowContextPropertiesImpl()
    map[""] = ""
    map.put("", "")
    val a = map[""]
}
```

Extension functions are _harder_ to call from Java because they become static functions so there's no reason for Java users to call it when we use it like this. It ends up like this from Java:

```java
public class Test {

    public static void main(String[] args) {
        FlowContextProperties map = new FlowContextPropertiesImpl();
        // I defined FlowContextProperties in ExternalEventResponseFactoryKt
        ExternalEventResponseFactoryKt.set(map, "", "");
    }
}
```

At the end of the day its not a big deal, I just want to remove the possible confusion of "do I call put or set?".

simon-johnson-r3: Yes, that's a good point, it also saves the effort of implementing it on the other side. I'll make this change.

application/src/main/kotlin/net/corda/v5/application/flows/FlowContextProperties.kt
```diff
@@ -0,0 +1,46 @@
...
+}
+
+/**
+ * Puts a user value into the context property store.
+ * Opens up [] operator access for setting values in Kotlin.
+ *
+ * @param key The property key
+ * @param value The property value
+ */
```
lankydan, line 65: Probably just copy the Kdoc from the `put` method.

application/src/main/kotlin/net/corda/v5/application/flows/FlowContextProperties.kt
```diff
@@ -0,0 +1,46 @@
+package net.corda.v5.application.flows
+
+/**
+ * Context properties of a [Flow] are key value pairs of Strings. They are comprised of two sets of properties, those
+ * set by the platform when a [Flow] is started, and those which are added by the CorDapp developer during the execution
+ * of their flow. The latter set are referred to as user context properties. All context properties are immutable once
+ * set.
+ * Where keys exist as platform properties already, setting user properties will always be ignored, only the platform
```
lankydan: Is there a reason why we ignore setting user properties if a platform one exists? Would it be better to throw an exception so its clear what the behaviour is?

lankydan (dismissed):
Left 2 comments but generally happy.

data/avro-schema/src/main/resources/avro/net/corda/data/flow/event/session/SessionInit.avsc
```diff
@@ -26,6 +26,22 @@
...
+      "type": {
+        "type": "map",
+        "values": "string"
+       },
```
JamesHR3: We might need to store these as arrays rather than maps, because I don't think the serializer is guaranteed to serialize the entries in the same order every time, which can result in problems when signing/verifying serialized objects.

simon-johnson-r3: As discussed offline, this will create a bit of churn downstream, but there is at least avro types for this already.

application/src/main/kotlin/net/corda/v5/application/flows/FlowContextProperties.kt
simon-johnson-r3: I did think about this and I can't come up with a good answer, except this avoids runtime exceptions. It's not possible to guess whether that's better than some unusual behaviour.

Something else we could suggest here is that users always scope their properties like packages. We could then do the same, e.g. `net.corda.v5.account`?

I'm open to suggestions.

lankydan: @JamesHR3 @mnesbit Any thoughts?

simon-johnson-r3: Decided offline to throw. Also throw on any "corda." properties.

application/src/main/kotlin/net/corda/v5/application/flows/FlowContextProperties.kt
```diff
@@ -0,0 +1,59 @@
...
+ *
+ * Both sets of context properties are passed from the originating [Flow] to sub-flows, initiated flows, and services.
+ * Where sub-flows and initiated flows have extra user properties added, these are only visible in the scope of those
+ * flows and any of their sub-flows, initiated flows or services, but not back up the flow stack to any flows which
+ * sit above the one in which they were added. In this way context properties can also be thought of as a stack,
+ * unwinding along with the flow stack.
+ */
+interface FlowContextProperties {
+    /**
+     * Puts a user value into the context property store.
```
mnesbit, line 11: We need to explain the scoping of the properties. i.e. when set by a flow/subflow it is available in all child subflows unless overwritten by one of them. If overwritten then the original value is restored when that child subflow returns.

mnesbit (dismissed):
OK

application/src/main/kotlin/net/corda/v5/application/flows/FlowContextProperties.kt
simon-johnson-r3: on further discussion offline it seems this is resolved by moving the existing explanation to the methods, plus adding a bit about overwrites, so that's what I've done

mnesbit (dismissed):
OK

JamesHR3 (dismissed):
The PR looks good, but I'm unclear why the build is failing?

simon-johnson-r3:
> The PR looks good, but I'm unclear why the build is failing?

Flakey CI build, re-running it removes the failure. I'm not sure of the mechanism behind the running of the compatibility checks when there's a breaking change, superficially I'd expect that not to run, but I guess it does something slightly different instead, for some reason that was erroneously failing.

JamesHR3 (approved):

PR #502, opened 2022-08-08
CORE-5165 Uniqueness DB schema

relyafi:
This change introduces the database schema required for the uniqueness checker component. See https://github.com/corda/corda-runtime-os/pull/1826 for the changes that utilise this schema.

This change must be merged prior to https://github.com/corda/corda-runtime-os/pull/1826

### DIFF
--- a/data/db-schema/src/main/kotlin/net/corda/db/schema/CordaDb.kt
+++ b/data/db-schema/src/main/kotlin/net/corda/db/schema/CordaDb.kt
@@ -8,6 +8,7 @@ package net.corda.db.schema
 enum class CordaDb(val persistenceUnitName: String) {
     CordaCluster("corda-cluster"),
     RBAC("corda-rbac"),
+    Uniqueness("corda-uniqueness"),
     Vault("corda-vault"),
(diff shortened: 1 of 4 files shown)


ifrankrui (approved):
LGTM. 

On the `bigint` data type for `issue_tx_output_idx` column in `uniqueness_state_details` table. Do we want to consider changing the `int` [default size](https://www.cockroachlabs.com/docs/stable/int.html#considerations-for-64-bit-signed-integers) on the CockroachDB side?

relyafi:
> LGTM.
> 
> On the `bigint` data type for `issue_tx_output_idx` column in `uniqueness_state_details` table. Do we want to consider changing the `int` [default size](https://www.cockroachlabs.com/docs/stable/int.html#considerations-for-64-bit-signed-integers) on the CockroachDB side?

@ifrankrui yes - this is part of the scope of [CORE-4958](https://r3-cev.atlassian.net/browse/CORE-4958). There's a TODO in the corresponding `corda-runtime-os` change to capture this as well.

PR #505, opened 2022-08-09
CORE-5401 Trusted Root Certificates - Add DB table to store trusted root certificates

mbrkic-r3:
CORE-5401 Added DB table to store trusted root certificates for packaging verification


josephzunigadaly (approved):

mbrkic-r3:
Closing as code signing certs can be stored in the existing cluster_certificates table that is used by P2P, for example (https://r3-cev.atlassian.net/wiki/spaces/CB/pages/3560505628/Decision+Log).

PR #508, opened 2022-08-10
CORE-5593: Explicitly constrain config `integer` values

vkolomeyko:
To prevent passing bogus values at config update time.
Looks like when no min/max values specified JSON schema does not do much of a validation in terms of range of the numbers as per: `com.networknt.schema.TypeValidator#isInteger`.

### DIFF
--- a/data/config-schema/src/main/resources/net/corda/schema/configuration/messaging/1.0/corda.messaging.json
+++ b/data/config-schema/src/main/resources/net/corda/schema/configuration/messaging/1.0/corda.messaging.json
@@ -58,7 +58,7 @@
                         "type": "integer",
                         "minimum": 1,
                         "maximum": 100000,
-                        "default": 5
+                        "default": 100
                       }
                     },
                     "additionalProperties": false
--- a/data/config-schema/src/main/resources/net/corda/schema/configuration/messaging/1.0/corda.messaging.json
+++ b/data/config-schema/src/main/resources/net/corda/schema/configuration/messaging/1.0/corda.messaging.json
@@ -55,7 +55,10 @@
                     "properties": {
                       "maxPollRecords": {
                         "description": "Max records to poll from db in a single poll.",
-                        "type": "integer"
+                        "type": "integer",
+                        "minimum": 1,
+                        "maximum": 100000,
+                        "default": 5
                       }
                     },
                     "additionalProperties": false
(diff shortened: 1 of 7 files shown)


data/config-schema/src/main/resources/net/corda/schema/configuration/messaging/1.0/corda.messaging.json
```diff
@@ -55,7 +55,10 @@
                     "properties": {
                       "maxPollRecords": {
                         "description": "Max records to poll from db in a single poll.",
-                        "type": "integer"
+                        "type": "integer",
```
JamesHR3, line 58: @fowlerrr what is the current default that we use? I'm a bit surprised we weren't specifying one here before actually, does that imply that we're ignoring this entirely?

JamesHR3 (commented):
Thanks @vkolomeyko for this. I suspect the maxima for these config values will need tweaking for  our stuff, which I'll raise a tech debt item for. That should hopefully be good enough to avoid overflows in the short term though.

I think we need to quickly verify what will happen when we change the DB bus poll default (as we weren't specifying one before) - it's not out the question we'll see issues otherwise. That's one for @fowlerrr.

data/config-schema/src/main/resources/net/corda/schema/configuration/messaging/1.0/corda.messaging.json
fowlerrr: This is in one of the default configs.
```
# Default properties for user configurable fields
consumer = {
    maxPollRecords = 100
}
```
It gets merged in with the config passed on subscription creation, which I _think_ comes from the config read service.

vkolomeyko: I have put defaulr value to 100.

gradle.properties
```diff
@@ -7,7 +7,7 @@ cordaProductVersion = 5.0.0
 # NOTE: update this each time this module contains a breaking change
 ## NOTE: currently this is a top level revision, so all API versions will line up, but this could be moved to
 ##   a per module property in which case module versions can change independently.
-cordaApiRevision = 155
```
JamesHR3: How come you need to bump this, and is there a corresponding runtime-os PR?

vkolomeyko: Yes, I wanted to test this change with `corda-runtime-os` and thought the easiest would be to bump the version. PR: https://github.com/corda/corda-runtime-os/pull/1853 which is green now.

JamesHR3 (approved):

kyriathar (approved):

PR #509, opened 2022-08-11
NOTICK: Update Avro to 1.11.1 and remove constraints that are no longer required

JamesHR3:
Update the Avro version to 1.11.1 from 1.11.0. The `commons-compress` dependency was updated in 1.11.0, so a constraints block we were previously using to force this dependency version can also be removed.


vkolomeyko (approved):
LGTM

ronanbrowne (approved):
LGTM

PR #513, opened 2022-08-15
CORE-5957: Registration status endpoint

yift-r3:
This PR adds API changes to allow persisting the registration request status (within the member database) and to read the requests statuses.

See runtime PR in https://github.com/corda/corda-runtime-os/pull/1889


data/avro-schema/src/main/resources/avro/net/corda/data/membership/p2p/MembershipRegistrationRequest.avsc
```diff
@@ -15,7 +15,7 @@
     {
       "name": "memberContext",
       "doc": "Member provided data in MemberInfo serialised as byte array by using KeyValuePairList.",
-      "type": "bytes"
```
charlieR3: We always send data as serialized bytes when we include a signature over that data for signature verification. We were told to do this in case serialization on different platforms returns slightly different byte arrays.

data/avro-schema/src/main/resources/avro/net/corda/data/membership/rpc/response/RegistrationStatus.avsc
```diff
@@ -0,0 +1,15 @@
+{
```
charlieR3: Is it possible to find a better package for this since it's used in a number of places now. e.g. it's in `rpc.response` but it's used in `db.request`

data/avro-schema/src/main/resources/avro/net/corda/data/membership/rpc/response/RegistrationStatusDetails.avsc
```diff
@@ -0,0 +1,44 @@
+{
```
charlieR3: It would be good to move this to a common place also since it's used in `db.response.query` also

data/avro-schema/src/main/resources/avro/net/corda/data/membership/rpc/response/RegistrationStatus.avsc
yift-r3: There is `net.corda.data.membership.state`. We can also put it in `net.corda.data.membership.common` or in `net.corda.data.membership` - Where would it be better?

charlieR3: `net.corda.data.membership.common` sounds good to me.

charlieR3 (dismissed):

charlieR3 (approved):

PR #514, opened 2022-08-15
CORE-6192 - remove-persistence-module

driessamyn:
- Removed `MappedSchema`
- Moved `CordaPersistenceException` and `UUIDConverter` into application module
- Removed persistence module
- Fixed up obsolete ledger module to support these changes.

Runtime os PR: https://github.com/corda/corda-runtime-os/pull/1890


ledger-obsolete/src/main/kotlin/net/corda/v5/ledger/obsolete/persistence/MappedSchema.kt
```diff
@@ -1,6 +1,9 @@
-package net.corda.v5.persistence
+package net.corda.v5.ledger.obsolete.persistence
 
 /**
+ * NOTE: MOVED THIS HERE AS THIS CLASS IS NOW REDUNDANT IN THE API. (CORE-6192)
```
driessamyn, line 4: FYI @vlajos 
Moved this class into this module as it is not needed anymore elsehwere.

lankydan: How will developers create custom tables in C5?

Before we would use `MappedSchema` for both ledger and non-ledger tables. With this change, I think we would lose that.

lankydan: Discussed with @driessamyn, due to CPK metadata now knowing what entities we have as well as liquibase scripts executing as defined in a CPK, we no longer need this class to create custom tables of any type.

lankydan (approved):

ledger-obsolete/src/main/kotlin/net/corda/v5/ledger/obsolete/persistence/MappedSchema.kt
driessamyn: Classes annotated with `@Entity` and `@CordaSerializable` are listed in the `Corda-Entity-Classes` CPK manifest attribute. This is what is used when creating a DB Sandbox.

ronanbrowne (approved):
LGTM

PR #518, opened 2022-08-16
CORE-4729 Create the UniquenessCheckerClientService API, and Result

nkovacsx:
Create a basic API for the Uniqueness Checker Client Service, and a result class (Success/Failure) it will return to the caller.


application/build.gradle
```diff
@@ -15,6 +15,7 @@ dependencies {
     api project(":crypto")
     api project(":membership")
     api project(":serialization")
+    api project(":data:avro-schema")
```
relyafi: Respect the existing alphabetical ordering of these includes instead of adding to the end

application/src/main/kotlin/net/corda/v5/application/uniqueness/client/UniquenessCheckerClientService.kt
```diff
@@ -0,0 +1,55 @@
...
+ * The request will be processed asynchronously and a [Future] will
+ * be returned that can be used to check the outcome.
+ *
+ * This service can be injected to either flows or other services.
+ *
+ * The actual request processing depends on which uniqueness service
+ * is invoked from the client.
+ *
+ * The API contains only one function that is used to queue up a request.
+ */
```
relyafi:
```suggestion
/**
 * This service is used to request uniqueness checking. The request will be processed asynchronously
 * and a [Future] will be returned that can be used to check the outcome.
 *
 * This service can be injected to either flows or other services.
 */
```


 
 Think this captures what is needed at the interface level. Regarding the removed content:
 
 - Queueing is an implementation detail
 - It's self-evident that an interface could have multiple implementations; stating this doesn't add value to the interface description
 - Similarly, it's self evident the interface contains one function
```

File: application/src/main/kotlin/net/corda/v5/application/uniqueness/client/UniquenessCheckerClientService.kt [hunk 3]
```diff
@@ -0,0 +1,55 @@
...
+@DoNotImplement
+interface UniquenessCheckerClientService {
+
+    /**
+     * This function is used to queue up a request for the client service,
+     * which will process it in an async manner, and return a [Future].
+     *
+     * @param txId The ID of the transaction that needs processing
+     *
+     * @param inputs A list of the input state refs that belongs
```
relyafi: Line lengths seem very conservative in this file; using 100 characters would probably reduce the number of lines in the file.

application/src/main/kotlin/net/corda/v5/application/uniqueness/client/UniquenessCheckerClientService.kt
```diff
@@ -0,0 +1,55 @@
...
+ * is invoked from the client.
+ *
+ * The API contains only one function that is used to queue up a request.
+ */
+@DoNotImplement
+interface UniquenessCheckerClientService {
+
+    /**
+     * This function is used to queue up a request for the client service,
+     * which will process it in an async manner, and return a [Future].
```
relyafi: Seems duplicate information from the interface description

application/src/main/kotlin/net/corda/v5/application/uniqueness/client/UniquenessCheckerClientService.kt
```diff
@@ -0,0 +1,55 @@
...
+     *
+     * @param txId The ID of the transaction that needs processing
+     *
+     * @param inputs A list of the input state refs that belongs
+     * o the given transaction, in a <TX_ID>:<STATE_NUM> format.
```
relyafi:
```suggestion
     * to the given transaction, in a <TX_ID>:<STATE_NUM> format.
```

application/src/main/kotlin/net/corda/v5/application/uniqueness/client/UniquenessCheckerClientService.kt
```diff
@@ -0,0 +1,55 @@
...
+     * This list might be empty if the given transaction is only
+     * doing issuance.
+     *
+     * @param references A list of the reference state refs that belongs
+     * o the given transaction, in a <TX_ID>:<STATE_NUM> format.
```
relyafi:
```suggestion
     * to the given transaction, in a <TX_ID>:<STATE_NUM> format.
```

application/src/main/kotlin/net/corda/v5/application/uniqueness/client/UniquenessCheckerClientService.kt
```diff
@@ -0,0 +1,55 @@
...
+     * @param numberOfOutputStates The number of output states the
+     * given transaction has.
+     *
+     * @param timeWindowLowerBound The earliest date/time from which
+     * the transaction is considered valid. This is an optional parameter.
+     *
+     * @param timeWindowUpperBound The latest date/time until the transaction
+     * is considered valid.
+     */
+    fun commitRequest(
```
relyafi: I think `commitRequest` is a carry-over from the implementation dating back to C4, and I'd question if this still makes sense? Something like `requestUniquenessCheck` would probably make more sense, because again, the commit behavior is a (server-side) implementation detail.

application/src/main/kotlin/net/corda/v5/application/uniqueness/client/UniquenessCheckerClientService.kt
```diff
@@ -0,0 +1,55 @@
...
+        txId: String,
+        inputs: List<String>,
+        references: List<String>,
+        numberOfOutputStates: Int,
+        timeWindowLowerBound: Instant?,
+        timeWindowUpperBound: Instant,
```
relyafi: Make sure the names of these align with `UniquenessCheckRequest` - there are a few discrepencies. In particular, ensuring inputs and references specifically mention _states_ may be important in future if the interface ends up being overloaded to support generic uniqueness checking.

relyafi (changes requested):
As a general comment, more thought needs to be given into ensuring consistency with the existing repo, and ensuring the interface documentation is only referring to things relevant to the interface and not implementation details.

application/src/main/kotlin/net/corda/v5/application/uniqueness/client/UniquenessCheckerClientService.kt
nkovacsx: Good point, I'll change it to `requestUniquenessCheck`, that makes more sense in the current context.

relyafi (commented):
LGTM now, but will hold off on final approval until the corresponding implementation PR in `corda-runtime-os` is raised and approved.

application/src/main/kotlin/net/corda/v5/application/uniqueness/model/UniquenessCheckError.kt
```diff
@@ -0,0 +1,51 @@
+package net.corda.v5.application.uniqueness.model
+
+import com.fasterxml.jackson.annotation.JsonSubTypes
+import com.fasterxml.jackson.annotation.JsonTypeInfo
+import net.corda.v5.base.annotations.CordaSerializable
+import java.time.Instant
+
+/**
+ * TODO Rewrite KDocs
```
relyafi: Is there a reason not to do this now?

application/src/main/kotlin/net/corda/v5/application/uniqueness/model/UniquenessCheckResponse.kt
```diff
@@ -0,0 +1,11 @@
+package net.corda.v5.application.uniqueness.model
+
+import net.corda.v5.application.crypto.DigitalSignatureAndMetadata
+
+/**
+ * TODO Write KDocs
```
relyafi: Is there a reason not to do this now?

application/src/main/kotlin/net/corda/v5/application/uniqueness/model/UniquenessCheckResult.kt
```diff
@@ -0,0 +1,130 @@
...
+import net.corda.data.uniqueness.UniquenessCheckExternalResultMalformedRequest
+import net.corda.data.uniqueness.UniquenessCheckExternalResultReferenceStateConflict
+import net.corda.data.uniqueness.UniquenessCheckExternalResultReferenceStateUnknown
+import net.corda.data.uniqueness.UniquenessCheckExternalResultTimeWindowOutOfBounds
+import net.corda.v5.base.exceptions.CordaRuntimeException
+import org.apache.avro.specific.SpecificRecord
+import java.time.Instant
+
+/**
+ * TODO Rewrite KDocs
```
relyafi: Is there a reason not to do this now?

application/src/main/kotlin/net/corda/v5/application/uniqueness/model/UniquenessCheckResult.kt
```diff
@@ -0,0 +1,130 @@
...
+ */
+sealed class UniquenessCheckResult(val commitTimestamp: Instant) {
+    companion object {
+        const val RESULT_ACCEPTED_REPRESENTATION = 'A'
+        const val RESULT_REJECTED_REPRESENTATION = 'R'
+
+        /**
+         * Converts an Avro error to a [Failure]. Since there's no common superclass
+         * for avro classes, the parameter can be anything, hence the [Any] type.
```
relyafi: But the parameter isn't `Any`?

application/src/main/kotlin/net/corda/v5/application/uniqueness/model/UniquenessCheckResult.kt
```diff
@@ -0,0 +1,130 @@
...
+         */
+        fun fromExternalError(avroResponse: UniquenessCheckExternalResponse): Failure {
```
relyafi: Given this is taking a response (not specifically an error response) wouldn't it make more sense to call this `fromAvroResponse` and have it handle the success case too?

application/src/main/kotlin/net/corda/v5/application/uniqueness/model/UniquenessCheckResult.kt
```diff
@@ -0,0 +1,130 @@
...
+                        )
+                    )
+                }
+                is UniquenessCheckExternalResultMalformedRequest -> {
+                    // FIXME This should be handled differently
+                    throw CordaRuntimeException("")
+                }
+                else -> {
+                    // TODO should this be handled differently?
+                    throw CordaRuntimeException("")
```
relyafi: Yes, but I don't see a reason to defer this? These are legitimate cases that should be properly handled now.

For the malformed request, it may be you need to extend the formerly internal class - there was no reason to have this before (the uniqueness check server doesn't store these) but as these classes are now dual purpose there may now be a need.

Regardless, it feels like we shouldn't be throwing exceptions here.

application/src/main/kotlin/net/corda/v5/application/uniqueness/model/UniquenessCheckStateDetails.kt
```diff
@@ -0,0 +1,11 @@
+package net.corda.v5.application.uniqueness.model
+
+import net.corda.v5.crypto.SecureHash
+
+/**
+ * TODO Rewrite KDocs
```
relyafi: Is there a reason not to do this now?

application/src/main/kotlin/net/corda/v5/application/uniqueness/model/UniquenessCheckStateRef.kt
```diff
@@ -0,0 +1,21 @@
+package net.corda.v5.application.uniqueness.model
+
+import net.corda.v5.crypto.SecureHash
+
+/**
+ * TODO Rewrite KDocs
```
relyafi: Is there a reason not to do this now?

application/src/main/kotlin/net/corda/v5/application/uniqueness/model/UniquenessCheckTransactionDetails.kt
```diff
@@ -0,0 +1,11 @@
+package net.corda.v5.application.uniqueness.model
+
+import net.corda.v5.crypto.SecureHash
+
+/**
+ * TODO Rewrite KDocs
```
relyafi: Is there a reason not to do this now?

data/avro-schema/src/main/resources/avro/net/corda/data/uniqueness/UniquenessCheckExternalRequest.avsc
```diff
@@ -1,6 +1,6 @@
 {
   "type": "record",
-  "name": "UniquenessCheckRequest",
+  "name": "UniquenessCheckExternalRequest",
```
relyafi: I think it might be better to clearer to append `Avro` to all of these rather than use External, e.g. `UniquenessCheckRequestAvro`. This is accurate (they are Avro specific) and is a bit more future proof if we did use something other than Avro in future.

data/avro-schema/src/main/resources/avro/net/corda/data/uniqueness/UniquenessCheckExternalResponse.avsc
```diff
@@ -0,0 +1,23 @@
+{
```
relyafi: Is there a reason this file was added / deleted instead of moved like the others? Moving preserves change history.

data/avro-schema/src/main/resources/avro/net/corda/data/uniqueness/UniquenessCheckExternalRequest.avsc
```diff
@@ -1,6 +1,6 @@
 {
```
relyafi: As we are making breaking API changes, I'm assuming you will need to bump the platform API version, as per this process: https://r3-cev.atlassian.net/wiki/spaces/CB/pages/4071850170/Working+with+corda-api+and+corda-runtime-os+repos+for+Corda+5#What-do-I-need-to-do-to-merge-a-breaking-change-in-corda-api%3F

relyafi (changes requested):

application/src/main/kotlin/net/corda/v5/application/uniqueness/model/UniquenessCheckError.kt
```diff
@@ -0,0 +1,51 @@
+package net.corda.v5.application.uniqueness.model
```
relyafi: Having seen the corresponding [corda-runtime-os PR](https://github.com/corda/corda-runtime-os/pull/1952/files), I actually wonder if this is in the right package - do we really want our backing store and uniqueness checker components and libraries to be dependent on `corda-application`? Conceptually that feels a bit back to front.

Maybe it should live in its own `uniqueness` module, similar to other services such as ledger, membership, persistence etc.

application/src/main/kotlin/net/corda/v5/application/uniqueness/model/UniquenessCheckResult.kt
nkovacsx: Good point, I was thinking about how to properly handle this but I think this is the best way.

data/avro-schema/src/main/resources/avro/net/corda/data/uniqueness/UniquenessCheckExternalResponse.avsc
nkovacsx: I have no idea tbh, I did the same thing as with the other files but for some reason it was deleted/recreated. Maybe some IntelliJ issue?

application/src/main/kotlin/net/corda/v5/application/uniqueness/model/UniquenessCheckError.kt
nkovacsx: Yeah that's a valid point!

application/build.gradle
```diff
@@ -13,6 +13,7 @@ dependencies {
     api platform(project(':corda-api'))
     api project(":base")
     api project(":crypto")
+    api project(":data:avro-schema")
```
mnesbit: This is a red flag sorry!

mnesbit (changes requested):

application/build.gradle
nkovacsx: That's exactly why I'm trying to move the whole API into a separate module as @relyafi suggested above. The whole conversation originated from that question on Slack.

application/src/main/kotlin/net/corda/v5/application/uniqueness/model/UniquenessCheckError.kt
```diff
@@ -0,0 +1,57 @@
...
+ * Representation of errors that can be raised by the uniqueness checker. These errors are returned
+ * through the uniqueness client service as well. This representation is agnostic to both the message
+ * bus API and any DB schema that may be used to persist data by the backing store.
+ *
+ * Unfortunately since this is a sealed class with multiple implementations we need to tell Jackson
+ * what types to look for. Hence the annotations. This will be removed when we start using standard
+ * Corda serialization.
+ */
+@CordaSerializable
+@JsonTypeInfo(use = JsonTypeInfo.Id.NAME, include = JsonTypeInfo.As.PROPERTY, property = "type")
```
mnesbit: None of this should be here, this should be invisible to the api user and on the actual implementations in corda-runtime-os. Only interfaces should be exposed here unless the objects are truly trivial POJOs.

application/build.gradle
nkovacsx: https://github.com/corda/corda-api/pull/518#discussion_r955164226

application/src/main/kotlin/net/corda/v5/application/uniqueness/model/UniquenessCheckError.kt
```diff
@@ -0,0 +1,57 @@
...
+@JsonSubTypes(
+    JsonSubTypes.Type(value = UniquenessCheckError.InputStateConflict::class),
+    JsonSubTypes.Type(value = UniquenessCheckError.InputStateUnknown::class),
+    JsonSubTypes.Type(value = UniquenessCheckError.ReferenceStateConflict::class),
+    JsonSubTypes.Type(value = UniquenessCheckError.ReferenceStateUnknown::class),
+    JsonSubTypes.Type(value = UniquenessCheckError.TimeWindowOutOfBounds::class),
+)
+sealed class UniquenessCheckError {
```
mnesbit: We should not put sealed classes into public API as we can never change them in future. This should be a small hierarchy of interfaces really.

application/src/main/kotlin/net/corda/v5/application/uniqueness/client/UniquenessCheckerClientService.kt
```diff
@@ -0,0 +1,44 @@
...
+    @Suppress("LongParameterList")
+    @Suspendable
+    fun requestUniquenessCheck(
+        txId: String,
+        inputStates: List<String>,
+        referenceStates: List<String>,
+        numOutputStates: Int,
+        timeWindowLowerBound: Instant?,
+        timeWindowUpperBound: Instant
+    ): Future<UniquenessCheckResponse>
```
mnesbit: Futures are not permitted in flows. This should be regarded as a blocking call that will return when done.

mnesbit:
As discussed in our meeting I don't think we should create a new module for this relatively small part of our API, especially considering the OSGi complexity of that. I also don't think the priority is to create a generic uniqueness service, even though I am keen to have that in the future. Therefore the most logical place for the API classes to live is in a package within :ledger rather than in the more abstract :application module which is all about simple flows with messaging, crypto and persistence.

ledger/ledger-utxo/build.gradle
```diff
@@ -0,0 +1,20 @@
+plugins {
+    id 'corda-api.common-library'
+    id 'corda.common-publishing'
+    id 'org.jetbrains.kotlin.jvm'
+}
+
+description 'Corda Uniqueness'
```
relyafi: Should this be updated?

ledger/ledger-utxo/detekt-baseline.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version='1.0' encoding='UTF-8'?>
```
relyafi: Is this file necessary given no suppressions?

ledger/ledger-utxo/src/main/kotlin/net/corda/v5/ledger/utxo/uniqueness/model/UniquenessCheckErrors.kt
```diff
@@ -0,0 +1,48 @@
+package net.corda.v5.ledger.utxo.uniqueness.model
+
+import java.time.Instant
+
+/**
+ * Representation of errors that can be raised by the uniqueness checker. These errors are returned
+ * through the uniqueness client service.
```
relyafi:
```suggestion
 * Representation of errors that can be raised by the uniqueness checker. These errors are returned
 * by the uniqueness checker back to the uniqueness client service, which propagates the errors to callers.
```

ledger/ledger-utxo/src/main/kotlin/net/corda/v5/ledger/utxo/uniqueness/model/UniquenessCheckResponse.kt
```diff
@@ -0,0 +1,16 @@
+package net.corda.v5.ledger.utxo.uniqueness.model
+
+import net.corda.v5.application.crypto.DigitalSignatureAndMetadata
+
+/**
+ * This is a response class that wraps the result of the uniqueness checking and a potential signature,
+ * if the request was processed successfully.
+ *
+ * @property result The result produced by the uniqueness checker
+ * @property signature Contains the signature provided by the uniqueness checker. Might be `null`
```
relyafi: _Might be_, or _will be_?

ledger/ledger-utxo/src/main/kotlin/net/corda/v5/ledger/utxo/uniqueness/model/UniquenessCheckTransactionDetails.kt
```diff
@@ -0,0 +1,12 @@
+package net.corda.v5.ledger.utxo.uniqueness.model
+
+import net.corda.v5.crypto.SecureHash
+
+/**
+ * Representation of the transaction details that must be persisted by the uniqueness checker,
+ * used by the uniqueness checker and backing store.
```
relyafi: Should also mention usage by the client?

relyafi (commented):
I strongly object to much of this being included in a `ledger-utxo` module. I understand this is because @mnesbit is insisting on it, however, it seems to me to be a bad idea on two fronts:

- While arguably certain elements of the API (e.g. `UniquenessCheckStateRef`) are specific to ledgers generally (states don't make sense outside of a ledger), there's nothing to say they will be specific to this specific ledger implementation. We already know that going forward, there may be multiple UTXO ledger implementations, which could well have the same view on how a state reference is represented.
- More concerning, is that the classes under `net.corda.v5.ledger.utxo.uniqueness.model` have actually been ported from `corda-runtime-os`, where they were previously used by the uniqueness checker implementation internally, including for persisting to the database. This means we are now introducing a dependency on the UTXO ledger in the uniqueness checker implementation, which is completely wrong. We went to great lengths to avoid this given @mnesbit's previous instructions to look towards a generic uniqueness service. Whilst I understand his view has now changed and we should not be actively focusing on this now given the time constraints, it does not seem like a good idea to be actively taking steps that undermine what has already been done. Of course, this could be avoided by having yet another set of classes (client / avro / server) but this also seems a bad choice, leading to duplication and more risk of introducing bugs due to not keeping interfaces in step.

Tagging @blsemo and @rick-r3 for visibility as well as I think you guys should be aware of the impacts of this decision.

Bottom line - @mnesbit if you are insisting on this approach in spite of the points raised above, please can you reply to this comment to confirm this.

relyafi:
Following further discussion verbally with @mnesbit regarding my earlier comment (https://github.com/corda/corda-api/pull/518#pullrequestreview-1097252607), we have agreed that:

- The client service that is injected into flows will remain in the `ledger-utxo` module
- Everything in the `net.corda.v5.ledger.utxo.uniqueness.model` package will move to a package under `application`, e.g. `net.corda.v5.application.uniqueness.model`. Strictly speaking, this does contain certain things that are arguably ledger specific, e.g. specific errors that relate to states. We won't try and dis-entangle this now, but may revisit in future.

@mnesbit please can you confirm this matches your understanding of what we agreed?

mnesbit:
Yes I am fine with that summary. I definitely don't want to go creating more modules unless they have distinct sandbox/OSGi requirements, or represent alternative higher level ledger models.

ledger/ledger-utxo/build.gradle
nkovacsx: Good point, once the ledger team merge their stuff this will be overridden anyway but I'll update it just in case.

ledger/ledger-utxo/detekt-baseline.xml
nkovacsx: Hmm, seems like it can be removed. I remember in DP1 this always had to be present!

ledger/ledger-utxo/src/main/kotlin/net/corda/v5/ledger/utxo/uniqueness/model/UniquenessCheckResponse.kt
nkovacsx: Yeah the phrasing here is not the best. I meant "This field might be null, in case the result is a failure", I'll rephrase it.

ledger/ledger-utxo/src/main/kotlin/net/corda/v5/ledger/utxo/uniqueness/model/UniquenessCheckTransactionDetails.kt
nkovacsx: Actually, I just realised that this class is not used in the API at all, so can be moved to the runtime-os.

relyafi (dismissed):
LGTM

ledger/ledger-utxo/src/main/kotlin/net/corda/v5/ledger/utxo/uniqueness/client/LedgerUniquenessCheckerClientService.kt
```diff
@@ -0,0 +1,44 @@
...
+     * @param timeWindowLowerBound The earliest date/time from which the transaction is considered valid.
+     * This is an optional parameter.
+     *
+     * @param timeWindowUpperBound The latest date/time until the transaction is considered valid.
+     */
+    @Suppress("LongParameterList")
+    @Suspendable
+    fun requestUniquenessCheck(
+        txId: String,
+        inputStates: List<String>,
```
mnesbit, line 39: Can we add an extension method version of this function in this file that takes `List<StateRef>` directly and does the toString operation

mnesbit (dismissed):
OK

ledger/ledger-utxo/src/main/kotlin/net/corda/v5/ledger/utxo/uniqueness/client/LedgerUniquenessCheckerClientService.kt
```diff
@@ -0,0 +1,67 @@
...
+ * and a [Future] will be returned that can be used to check the outcome. This service can be injected
+ * to either flows or other services.
+ */
+@DoNotImplement
+interface LedgerUniquenessCheckerClientService {
+
+    /**
+     * @param txId The ID of the transaction that needs processing
+     *
+     * @param inputStates A list of the input state refs that belongs to the given transaction in a
```
mnesbit, line 21: We should be clear about the case sensitivity of these string inputs. (I am assuming they are case sensitive, but that in turn does mean that the StateRef.toString() has to be formally stable)

ledger/ledger-utxo/src/main/kotlin/net/corda/v5/ledger/utxo/uniqueness/client/LedgerUniquenessCheckerClientService.kt
```diff
@@ -0,0 +1,67 @@
...
+fun LedgerUniquenessCheckerClientService.requestUniquenessCheck(
+    txId: String,
+    inputStates: List<UniquenessCheckStateRef>,
+    referenceStates: List<UniquenessCheckStateRef>,
+    numOutputStates: Int,
+    timeWindowLowerBound: Instant?,
+    timeWindowUpperBound: Instant
+) = requestUniquenessCheck(
+    txId,
+    inputStates.map { it.toString() },
```
mnesbit, line 62: Actually is this right? That converts to <txid>(<output>) https://github.com/corda/corda-api/blob/a064bbfe01498385db2d52846e0ca306e24fc966/ledger-obsolete/src/main/kotlin/net/corda/v5/ledger/obsolete/contracts/States.kt#L89

ledger/ledger-utxo/src/main/kotlin/net/corda/v5/ledger/utxo/uniqueness/client/LedgerUniquenessCheckerClientService.kt
nkovacsx: As discussed this is using the `UniquenessCheckStateRef` interface where the toString functionality will depend on the implementation.

ledger/ledger-utxo/src/main/kotlin/net/corda/v5/ledger/utxo/uniqueness/client/LedgerUniquenessCheckerClientService.kt
nkovacsx: As discussed this is using the UniquenessCheckStateRef interface where the toString functionality will depend on the implementation.

relyafi (dismissed):

relyafi (dismissed):

relyafi (dismissed):

relyafi (approved):

ronanbrowne (approved):
Build changes approved, but please remember you will need to update the composite build logic in runtime-os settings.gradle to ensure not to break any non R3er building via the composite build method , logic here https://github.com/corda/corda-runtime-os/blob/release/os/5.0/settings.gradle#L404

PR #519, opened 2022-08-17
INFRA-1811 add arm64 config to allow for further testing of this architecture 

ronanbrowne:
- add needed config to use arm64 machine with this job


yarivus (approved):
LGTM

PR #520, opened 2022-08-17
CORE-5939 crypto config schema

alexey-kadyrov-r3:
Defined the new crypto schema


data/config-schema/src/main/resources/net/corda/schema/configuration/cryptoLibrary/1.0/corda.cryptoLibrary.json
```diff
@@ -5,6 +5,406 @@
   "description": "Configuration schema for the crypto library subsection.",
   "type": "object",
   "properties": {
-    "additionalProperties": true
-  }
+    "cryptoConnectionFactory": {
+      "description": "Settings for database connections factory's cache of EntityManagerFactory(s)",
+      "type": "object",
+      "default": {},
+      "properties": {
```
charlieR3, line 12: Could nest this under `cache` to align with `signingService`

data/config-schema/src/main/resources/net/corda/schema/configuration/cryptoLibrary/1.0/corda.cryptoLibrary.json
```diff
@@ -5,6 +5,406 @@
...
+          "type": "object",
+          "default": {},
+          "properties": {
+            "maxAttempts": {
+              "description": "Maximum attempts to process a message",
+              "type": "integer",
+              "default": 3
+            },
+            "waitBetweenMills": {
+              "description": "Timeouts between attempts in milliseconds, if the number of value less than intervals then the last item is repeated",
```
charlieR3:
```suggestion
              "description": "Time between attempts in milliseconds, if the number of values is less than attempts then the last item is repeated",
```

data/config-schema/src/main/resources/net/corda/schema/configuration/cryptoLibrary/1.0/corda.cryptoLibrary.json
```diff
@@ -5,6 +5,406 @@
...
```
charlieR3:
```suggestion
              "description": "Time between attempts in milliseconds, if the number of values is less than attempts then the last item is repeated",
```

data/config-schema/src/main/resources/net/corda/schema/configuration/cryptoLibrary/1.0/corda.cryptoLibrary.json
```diff
@@ -5,6 +5,406 @@
...
```
charlieR3:
```suggestion
              "description": "Time between attempts in milliseconds, if the number of values is less than attempts then the last item is repeated",
```

data/config-schema/src/main/resources/net/corda/schema/configuration/cryptoLibrary/1.0/corda.cryptoLibrary.json
```diff
@@ -5,6 +5,406 @@
...
+                "name",
+                "masterKeyPolicy",
+                "capacity",
+                "supportedSchemes",
+                "categories",
+                "cfg"
+              ]
+            }
+          },
+          "additionalProperties": false        }
```
charlieR3, line 405: Formatting

charlieR3 (dismissed):
Some comments added. Approved as I don't think they're major.

data/config-schema/src/main/resources/net/corda/schema/configuration/cryptoLibrary/1.0/corda.cryptoLibrary.json
```diff
@@ -5,6 +5,406 @@
...
+        "expireAfterAccessMins": {
+          "description": "Expiration time in minutes for cached EntityManagerFactory",
+          "type": "integer",
+          "default": 5
+        },
+        "maximumSize": {
+          "description": "Maximum number of cached EntityManagerFactory",
+          "type": "integer",
+          "default": 3
+        }
```
driessamyn, line 22: Slightly separate to this PR actually, but I don't think caching of db connections and EMFs should happen at the component level. Instead we should do this centrally as the same connections are used by multiple components.
So I'm not too keen on having this on the config schema.

data/config-schema/src/main/resources/net/corda/schema/configuration/cryptoLibrary/1.0/corda.cryptoLibrary.json
alexey-kadyrov-r3: The service which provides the connections is more global than the signing service only, it's used in other places

charlieR3 (approved):

data/config-schema/src/main/resources/net/corda/schema/configuration/cryptoLibrary/1.0/corda.cryptoLibrary.json
driessamyn: CC @mnesbit , WDYT ^ ?

My worry is if components each have their own cache of connections, it'll be very hard to tune this as vnode connections, for example, are used across different components, so all these caches will be fighting eachother when a central cache would make more sense.

alexey-kadyrov-r3: I would prefer that to be central as well

PR #522, opened 2022-08-18
CORE-5402 Trusted Root Certificates - Retrieve certificates and pass to verification library

mbrkic-r3:
- Updated net.corda.data.certificates.rpc schemas to support requesting all certificates for specific tenant

### DIFF
--- a/data/avro-schema/src/main/resources/avro/net/corda/data/certificates/rpc/request/CertificateRpcRequest.avsc
+++ b/data/avro-schema/src/main/resources/avro/net/corda/data/certificates/rpc/request/CertificateRpcRequest.avsc
@@ -6,7 +6,7 @@
   "fields": [
     {
       "name": "tenantId",
-      "doc": "'p2p', 'rpc-api', or holding identity identity ID.",
+      "doc": "'p2p', 'rpc-api', 'codesigner', or holding identity identity ID.",
       "type": "string"
     },
     {
--- a/data/avro-schema/src/main/resources/avro/net/corda/data/certificates/rpc/request/RetrieveCertificateRpcRequest.avsc
+++ b/data/avro-schema/src/main/resources/avro/net/corda/data/certificates/rpc/request/RetrieveCertificateRpcRequest.avsc
@@ -6,8 +6,11 @@
   "fields": [
     {
       "name": "alias",
-      "doc": "The certificate alias.",
-      "type": "string"
+      "doc": "The certificate alias. If null, all certificates (for specified tenant) will be retrieved",
+      "type": [
+        "string",
+        "null"
+      ]
     }
   ]
 }
--- a/data/avro-schema/src/main/resources/avro/net/corda/data/certificates/rpc/response/CertificateRetrievalRpcResponse.avsc
+++ b/data/avro-schema/src/main/resources/avro/net/corda/data/certificates/rpc/response/CertificateRetrievalRpcResponse.avsc
@@ -2,15 +2,15 @@
   "type": "record",
   "name": "CertificateRetrievalRpcResponse",
   "namespace": "net.corda.data.certificates.rpc.response",
-  "doc": "RPC operations response with a certificate chain.",
+  "doc": "RPC operations response with a collection of certificates.",
   "fields": [
     {
       "name": "certificates",
-      "doc": "The certificates in PEM format.",
-      "type": [
-        "string",
-        "null"
-      ]
+      "doc": "Collection of certificates encoded in PEM format.",
+      "type": {
+        "type": "array",
+        "items": "string"
+      }
     }
   ]
 }
--- a/gradle.properties
+++ b/gradle.properties
@@ -7,7 +7,7 @@ cordaProductVersion = 5.0.0
 # NOTE: update this each time this module contains a breaking change
 ## NOTE: currently this is a top level revision, so all API versions will line up, but this could be moved to
 ##   a per module property in which case module versions can change independently.
-cordaApiRevision = 158
+cordaApiRevision = 159
 
 # Main
 kotlinVersion = 1.7.10


data/avro-schema/src/main/resources/avro/net/corda/data/certificates/rpc/request/CertificateRpcRequest.avsc
```diff
@@ -6,7 +6,7 @@
   "fields": [
     {
       "name": "tenantId",
-      "doc": "'p2p', 'rpc-api', or holding identity identity ID.",
+      "doc": "'p2p', 'rpc-api', 'codesigner', or holding identity identity ID.",
```
alexey-kadyrov-r3, line 9: What's the `codesigner` tenant?

mbrkic-r3: It will be used for code signing certs (packaging verification), see https://r3-cev.atlassian.net/wiki/spaces/CB/pages/3560505628/Decision+Log (2022 08 10)

driessamyn: Tenant ID is a wrong name in this case (it should be cert type), but I think we said this would be refactored later on.

dimosr: I had a chat with @charlieR3 about this that will probably come up as part of the refactoring. But, I think both might be needed. Tenant ID really corresponds to the tenant ID under which the private key for the corresponding certificate is stored under (might not apply to all the certificates, so it could become optional with the new changes). Certificate type is a separate piece of information. It is possible one can be inferred from the other, but not 100% sure whether that's better than retaining both as the mapping logic might end up being quite complicated and hard to maintain.

dimosr: without having deep understanding of that use-case, I got the impression this is a case where we don't store the corresponding private key. So, it might make sense to have the tenant-id be empty. But, in that case, we'd need to introduce the extra column (e.g. cert type) for information purposes. Makes me wonder should we do that now (not sure how much work it would be)?

alexey-kadyrov-r3: Am I right assuming that in C5 the certs tale (and service) will hold certificates belonging to "entities" (as persons, organisations, etc.) which are not transacting members but involved in signing the CPIs?

alexey-kadyrov-r3: If so then @dimosr is right we cannot use the tenant concept and be better with the type or the tenant being null

driessamyn: In this usecase, the "tenant" is the cluster, but I thought that was the same for RPC, for example, and IIRC from the conversation last week we thought that because this is the cluster-wide table (not the VNode one), this would be the case for all certs.
What slightly concerns me is that the docs also mention "holding identity identity ID". Shouldn't those be in the vnode DB?

alexey-kadyrov-r3: The tenancy assumes some kind of ownership and operations on behalf of the tenant (in other places ownership of the private key and signing). If we just store those certs and using them as authentication/validation of entities then there is no notion of tenancy here, as we don't go beyond that, so I would be reluctant to use the tenant id here

driessamyn: What is a tenant? And do we need to add this to https://r3-cev.atlassian.net/wiki/spaces/CB/pages/1843101750/Corda+5+Glossary?

Tenant to me sounds like it is related to an identity, but in the list we have values like "P2P" and "RPC-API". To me, this sounds more like type of certs. But I'm still confused why Holding Identity ID is also a possible option as I wouldn't expect that to be in the cluster DB, but rather be in the VNode DB.

mbrkic-r3: Tenant is definitely not the right name for code signing use case. Code signing certificate is a cluster wide certificate that has a specific purpose, it's a trusted certificate for packaging signature verification. I guess we'll also have other cluster wide certificates for different purposes. This PR is just a temporary workaround in order to reuse existing code but, as written in the decision log, proper solution and required refactoring should be done later with https://r3-cev.atlassian.net/browse/CORE-6130

alexey-kadyrov-r3: > What is a tenant?

Anything which delegates to crypto library to handle its private keys - like members (vnodes) or platform systems where each system is separate tenant

alexey-kadyrov-r3: > And do we need to add this...

Probably as that is exposed in the current shape of our RPC API, but in it probably should be called there as `crypto tenant`

dimosr: > What slightly concerns me is that the docs also mention "holding identity identity ID". Shouldn't those be in the vnode DB?

I think this is how it is. The requests are routed differently depending on the tenant ID. For cluster-level tenants (e.g. p2p , rpc-api), requests are sent to the cluster DB. For holding ID tenants, requests are sent to the vnode DB.

dimosr: speaking of which made me think.. seems like tenant-id started as an isolation term for the crypto processor, but it is now being used in a wider context (e.g. DB routing for cert storage). We could leave it as-is for now and refactor it as part of the other ticket on the membership team's backlog as it might require some more thinking.

However, I want to emphasise that tenant ID is much more than just a determinant of the certificate type.

alexey-kadyrov-r3: Yes, the tenant id is defined only for crypto processor.

driessamyn: Discussing this further with @mbrkic-r3 last night, I'm not sure we actually need this change in api.
We only need the certificates from the DB worker, when validating a CPI, so it doesn't make sense to use this Kafka API to fetch them.

mbrkic-r3:
As per comment above, certificates for packaging verification won't be retrieved via Kafka so I'm closing this PR

PR #525, opened 2022-08-23
CORE-6335 cleaned up crypto API

alexey-kadyrov-r3:


charlieR3 (dismissed):

charlieR3 (dismissed):

application/src/main/kotlin/net/corda/v5/application/crypto/HashingServiceUtils.kt
```diff
@@ -0,0 +1,47 @@
...
+ *
+ * @param str The algorithm name followed by a delimiter and the sequence of hexadecimal digits that represents a digest.
+ *
+ * @throws IllegalArgumentException The input string does not contain the expected number of hexadecimal digits, or it contains
+ * incorrectly-encoded characters.
+ *
+ * @see SecureHash.create
+ */
+@Suppress("TooGenericExceptionCaught")
+fun HashingService.create(str: String): SecureHash {
```
mnesbit: I think we generally refer to this as parse on other objects

alexey-kadyrov-r3: The name was left from C4 but anyway I've renamed it as well as SecureHash.create

mnesbit (approved):
OK

ronanbrowne (approved):
LGTM

PR #526, opened 2022-08-23
CORE-5375: vnode db reset

benmcmahon100:
https://github.com/corda/corda-runtime-os/pull/1943


data/avro-schema/src/main/kotlin/net/corda/data/chunking/PropertyKeys.kt
```diff
@@ -6,4 +6,6 @@ object PropertyKeys {
      * Boolean property specifying whether same data content can re-upload again
      */
     const val FORCE_UPLOAD = "forceUpload"
-}
\ No newline at end of file
+    const val RESET_DB = "resetDb"
```
vkolomeyko, line 12: Might be worth putting some comments for the purpose of those properties.

data/avro-schema/src/main/resources/avro/net/corda/data/virtualnode/VirtualNodeDBResetRequest.avsc
```diff
@@ -0,0 +1,19 @@
...
+      "name": "holdingIdentityShortHashs",
+      "type": {
+        "type": "array",
+        "items": "string"
+      }
+    },
+    {
+      "name": "updateActor",
+      "type": "string",
+      "doc": "ID of RPC user that requested the virtual node state update.."
```
vkolomeyko:
```suggestion
      "doc": "ID of RPC user that requested the virtual node state update."
```

data/avro-schema/src/main/resources/avro/net/corda/data/virtualnode/VirtualNodeDBResetRequest.avsc
```diff
@@ -0,0 +1,19 @@
+{
+  "type": "record",
+  "name": "VirtualNodeDBResetRequest",
+  "namespace": "net.corda.data.virtualnode",
+  "fields": [
+    {
+      "name": "holdingIdentityShortHashs",
```
vkolomeyko:
```suggestion
      "name": "holdingIdentityShortHashes",
```

data/avro-schema/src/main/resources/avro/net/corda/data/virtualnode/VirtualNodeDBResetResponse.avsc
```diff
@@ -0,0 +1,14 @@
+{
+  "type": "record",
+  "name": "VirtualNodeDBResetResponse",
+  "namespace": "net.corda.data.virtualnode",
+  "fields": [
+    {
+      "name": "holdingIdentityShortHashs",
```
vkolomeyko:
```suggestion
      "name": "holdingIdentityShortHashes",
```

vkolomeyko:
This PR is in conflicted state and need to be resolved.
Also I am wondering if this PR represents a breaking change... Looks like we are just additing new message types, therefore we can merge this ahead of/independently from: https://github.com/corda/corda-runtime-os/pull/1943

vkolomeyko (approved):
LGTM

data/avro-schema/src/main/kotlin/net/corda/data/chunking/PropertyKeys.kt
```diff
@@ -6,4 +6,12 @@ object PropertyKeys {
...
-}
\ No newline at end of file
+    /**
+     * Boolean property specifying whether the vnode databases should be deleted and recreated
+     */
+    const val RESET_DB = "resetDb"
+    /**
+     * String property specifying the RPC actor that made the request
+     */
+    const val ACTOR = "actor"
```
sprintengine, line 16: Ok, but seems strange to have actor as a property rather than as a String on these APIs seeing as it seems to be mandatory - we seem to always set it in forceCpiUpload and uploadCpi, and CpiValidatorImpl has a validation check to make sure its set. 

It would require changing a lot of places to do this so happy enough to go with this approach using property instead.

PR #528, opened 2022-08-24
CORE-5735 Cleaned up application API

owenstanford:
Removing unimplemented APIs from the application


charlieR3 (dismissed):
`membership` changes LGTM.

JamesHR3 (dismissed):
This looks about right in terms of what we're removing (possibly we don't need `HashingService` yet? That needs a bit more discussion though). I suspect you're behind the API version curve though.

lankydan (dismissed):

JamesHR3 (approved):

PR #530, opened 2022-08-25
NOTICK: Remove superfluous inline UntrustworthyData.unwrap() method.

chrisr3:
Declare `UntrustworthyData.Validator` to be a Kotlin functional interface so that Kotlin can use it to generate lambdas. This allows us to remove the inline extension method, and hence make `UntrustworthyData.fromUntrustedWorld` private too.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -7,7 +7,7 @@ cordaProductVersion = 5.0.0
 # NOTE: update this each time this module contains a breaking change
 ## NOTE: currently this is a top level revision, so all API versions will line up, but this could be moved to
 ##   a per module property in which case module versions can change independently.
-cordaApiRevision = 162
+cordaApiRevision = 163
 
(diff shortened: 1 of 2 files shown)


gradle.properties
```diff
@@ -7,7 +7,7 @@ cordaProductVersion = 5.0.0
 # NOTE: update this each time this module contains a breaking change
 ## NOTE: currently this is a top level revision, so all API versions will line up, but this could be moved to
 ##   a per module property in which case module versions can change independently.
-cordaApiRevision = 162
```
JamesHR3, line 10: Do you need to update the API version number? Does this cause any breaking changes?

JamesHR3: Never mind - I missed the second PR :)

JamesHR3 (approved):

PR #532, opened 2022-08-25
NOTICK: Allow Crypto functions to be invoked from Java.

chrisr3:
Remove Kotlin types from Crypto API signatures and apply `@JvmStatic` annotations, allowing them to be invoked from Java.

Note also that Java doesn't support Kotlin's default parameters either :wink:


alexey-kadyrov-r3 (approved):
LGTM

chrisr3:
> LGTM

Thanks, I'll just check my composite build still works before merging it :wink:

PR #533, opened 2022-08-25
CORE-6415: Remove Kotlin-specific references from our API signatures.

chrisr3:
API signatures containing Kotlin classes cannot be invoked from Java, e.g. Kotlin lambda parameters. Replace these with standard Java functional interfaces instead.

Also apply `@JvmStatic` annotations to `companion object` functions, and assign non-default names to `*Kt` classes.


alexey-kadyrov-r3 (approved):
LGTM

PR #536, opened 2022-08-28
CORE-6384: Add capability to publish to a S3 Bucket 

ronanbrowne:
Add capability to publish to a S3 Bucket to the corda-api submodule. 

Other submodules are already catered for by the logic in corda-gradle-internal plugin however this submodule uses custom publishing logic and does not consume the r3 publish plugin.

Longer term this should be refactored into corda-gradle-internal plugins, however this requires a larger re work of that logic .Tech debt Jira for this [here](https://r3-cev.atlassian.net/browse/CORE-6425)

### DIFF
--- a/corda-api/build.gradle
+++ b/corda-api/build.gradle
@@ -60,6 +60,7 @@ dependencies {
     }
 }
 
+
 publishing {
     publications {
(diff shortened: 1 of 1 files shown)


yarivus (approved):
LGTM

PR #538, opened 2022-08-29
CORE-6420 extracting crypto api extensions

alexey-kadyrov-r3:
* Moved DigestAlgorothm and DigestAlgorithmFactory to a new `crypto-extensions` API module
* Removed BackoffStrategy from API


chrisr3 (commented):
This is a new _and small_ API module, and so this is the _perfect_ time to reimplement it in Java instead of Kotlin. And _not least_ because both `DigestAlgorithm` and `DigestAlgorithmFactory` contain `val` properties which we are inviting third parties to implement:
```kotlin
interface DigestAlgorithm {
    val algorithm: String

    val digestLength: Int
```
Converting these to Java later would be a _breaking API change_ because Kotlin only allows a  `val` to override / implement another `val`. (See [KT-6653](https://youtrack.jetbrains.com/issue/KT-6653).)

This bundle's OSGi metadata also includes:
```
Export-Package: net.corda.v5.crypto.extensions;uses:=kotlin;version="5.0.0"
Import-Package: java.io,java.lang,kotlin;version="[1.7,2)"
```
which means that CPKs implementing `DigestAlgorithm` will no longer resolve when Corda eventually upgrades to Kotlin 2.0. In fact, Kotlin may break this CPK before then if it drops support for decoding Kotlin 1.7 metadata. In either case, we would need to provide sandboxes with their own Kotlin bundles to fix this, but Corda API bundles written in Kotlin could only be used by one version of the Kotlin libraries.

alexey-kadyrov-r3:
I could do that in Java but that should be concerted approach otherwise it'll look strange why only one module is in Java.

charlieR3 (approved):

ronanbrowne (approved):
LGTM

PR #540, opened 2022-08-30
CORE-5735 Clean-up config schemas

owenstanford:
- Remove unused configuration schemas 
- Add missing schema properties


data/config-schema/src/main/resources/net/corda/schema/configuration/reconciliation/1.0/corda.reconciliation.json
```diff
@@ -14,28 +14,28 @@
...
-      "description": "",
+      "description": "Interval in ms between updating Kafka with missing CPI metadata compared to CPI metadata existing in the DB.",
       "type": "integer",
       "minimum": 5000,
       "maximum": 2147483647,
       "default": 30000
     },
     "configIntervalMs": {
-      "description": "",
+      "description": "Interval in ms between updating Kafka with missing Configuration sections compared to Configuration sections existing in the DB.",
```
kyriathar: Actually my bad, can we update these to: 

"Interval in ms between aligning Kafka X state with DB X state." 

to be more precise please, because reconciliation also includes updates and deletes.

owenstanford: Updated.

data/config-schema/src/main/resources/net/corda/schema/configuration/sandbox/1.0/corda.sandbox.json
```diff
@@ -4,5 +4,19 @@
...
+    "cache": {
+      "description": "Settings for sandbox caching",
+      "type": "object",
+      "default": {},
+      "properties": {
+        "size": {
+          "description": "The maximum number cached sandboxes.",
+          "type": "integer",
+          "minimum": 0,
+          "default": 2
```
mnesbit, line 17: This is only suitable for enhancing our testing, not for production so may need a think

driessamyn: The problem is that we don't have any idea of what a reasonable number is atm. (10, 100, 1000?)
From the soak tests, in the sample app used, it doesn't look like creating a sandbox has a massive impact, so I suggested we leave it at this so that we continue to stress both caching and eviction. Once we have metrics in place, we'll do some additional testing and try to come up with a sensible default. I planned that to be in the beta 1 scope.

lankydan (approved):

PR #542, opened 2022-08-31
CORE-6249 FlowSession local context

simon-johnson-r3:
* Additional property on FlowSession to expose a context local to that session. The immutable property describes the context of the counterparty if this is an initiated Flow. For an initiating Flow the context can be mutated to add extra context for this FlowSession's initiated Flows only.
* Moved the corda reserved property prefix to the api because it forms part of the contract.


application/src/main/kotlin/net/corda/v5/application/messaging/FlowSession.kt
```diff
@@ -31,6 +37,26 @@ interface FlowSession {
      */
     val counterparty: MemberX500Name
```
owenstanford, line 39: As discussed, the concept of creating a session, then being able to mutate the context, but only until the point the initiated flow is created (first send or receive) is confusing. 
Effectively the context on a session is an immutable snapshot. If the user wants to set context properties for a given session they should only be able to do this at the point of creating the session, the API should be updated to reflect this and the contextProperties on the FlowSession should be read-only

application/src/main/kotlin/net/corda/v5/application/messaging/FlowMessaging.kt
```diff
@@ -8,12 +9,38 @@ import net.corda.v5.base.types.MemberX500Name
...
+     * Note that this function does not communicate in itself, the counter-flow will be kicked off by the first
+     * send/receive.
+     *
+     * This overload takes a mutator of context properties. Any properties set or modified against the context passed to
+     * this mutator will be propagated to initiated flows and all that flow's initiated flows and sub flows down the
+     * stack. The properties passed to the mutator are pre-populated with the parent context properties, see
+     * [FlowContextProperties].
+     *
+     * @param x500Name The X500 name of the member to communicate with.
+     * @param flowContextPropertiesMutator A mutator of context properties
```
owenstanford: Sorry to be pedantic:
Missing full stop on flowContextProperiesMutator
The style for the API KDocs is to have a new line between the sets of @param, @retrun and @throws

application/src/main/kotlin/net/corda/v5/application/messaging/FlowMessaging.kt
```diff
@@ -8,12 +9,38 @@ import net.corda.v5.base.types.MemberX500Name
...
+     * @return The session.
+     */
+    @Suspendable
+    fun initiateFlow(x500Name: MemberX500Name, flowContextPropertiesMutator: FlowContextPropertiesMutator): FlowSession
```
owenstanford: Should update the FlowMessagingJavaApiTest for this as well.

application/src/main/kotlin/net/corda/v5/application/messaging/FlowMessaging.kt
```diff
@@ -8,12 +9,40 @@ import net.corda.v5.base.types.MemberX500Name
...
+     * stack. The properties passed to the mutator are pre-populated with the parent context properties, see
+     * [FlowContextProperties].
+     *
+     * @param x500Name The X500 name of the member to communicate with.
+     * @param flowContextPropertiesMutator A mutator of context properties.
+     *
+     * @return The session.
+     */
+    @Suspendable
+    fun initiateFlow(x500Name: MemberX500Name, flowContextPropertiesMutator: FlowContextPropertiesMutator): FlowSession
```
owenstanford: This method can throw now, if the mutator callback tries to set an invalid key?

application/src/main/kotlin/net/corda/v5/application/messaging/FlowMessaging.kt
mnesbit: Is this overcomplicating things? When exactly is the mutator applied? Is it applied immediately, or is it applied at first packet when we start handshake? If applied immediately should we just pass in sessionContext: Map<String, String> instead?

owenstanford (dismissed):

application/src/main/kotlin/net/corda/v5/application/messaging/FlowMessaging.kt
mnesbit: Also I am trying to think what happens when we create a session in an outer flow and then pass the session down into an inline subflow. Is the base session context derived from the original non-inline subflow, or from the point of use?

simon-johnson-r3: This was also considered, but ruled out because
* then you present two different apis to the user to set context
* you don't get any of the validation you get from FlowContext
* when used with a lambda in Java or Kotlin the api is trivial to use:

```
In Java this can be passed a lambda, e.g.
 session = flowMessaging.initiateFlow(vnName, (flowContextProperties) -> {
                // mutate context
              });

In Kotlin:
        val session =
            flowMessaging.initiateFlow(flowEngine.virtualNodeName) { flowContextProperties ->
              // mutate context
            }
```
No user created maps need to be involved in the process.

I attempted to document that the properties are snapshotted here: https://github.com/corda/corda-api/pull/542/files#diff-143329fd07352dd773d710271bd76e65b232e150ff0788873c3573b8a0ecb901R34
perhaps that isn't clear enough? On re-reading I see it's probably clearer on the other overload.

simon-johnson-r3: Re: subflows. Yes, the session context remains immutable, even if you pass it to a subflow, so changing the subflow context would then have no effect at all. Whilst this version prohibits that kind of "advanced" context capturing, the previous iteration captured context at the first point you sent a message and then, no matter what you did, it stayed the same, even though that wasn't quite so apparent to the user because they could continue to modify it.

So this is limiting the feature a little, but it always works the same way and results shouldn't be unexpected.

application/src/main/kotlin/net/corda/v5/application/flows/FlowContextProperties.kt
```diff
@@ -1,20 +1,30 @@
...
  * set by the platform when a [Flow] is started, and those which are added by the CorDapp developer during the execution
  * of their flow. The latter set are referred to as user context properties. All context properties are immutable once
  * set.
  */
 interface FlowContextProperties {
+    companion object {
+        /**
+         * A constant containing the Corda reserved prefix.
+         */
+        const val CORDA_RESERVED_PREFIX = "corda." // constant must be lowercase
```
lankydan: Should include `@JvmStatic`.

application/src/main/kotlin/net/corda/v5/application/messaging/FlowMessaging.kt
lankydan: Discussed with @simon-johnson-r3, think his solution seems like the best solution and decided that "builder" rather than "mutator" would be better.

owenstanford (dismissed):

application/src/main/kotlin/net/corda/v5/application/messaging/FlowMessaging.kt
```diff
@@ -8,12 +10,47 @@ import net.corda.v5.base.types.MemberX500Name
...
+     *
+     * @return The session.
+     *
+     * @throws IllegalArgumentException if the builder tries to set a property for which a platform property already
+     * exists or if the key is prefixed by [CORDA_RESERVED_PREFIX]. See also [FlowContextProperties]. Any other
+     * exception thrown by the builder will also be thrown through here and should be avoided in the provided
+     * implementation, see [FlowContextPropertiesBuilder].
+     */
+    @Suspendable
+    fun initiateFlow(x500Name: MemberX500Name, flowContextPropertiesBuilder: FlowContextPropertiesBuilder): FlowSession
```
lankydan, line 65: Since we're going through and putting examples in some of the Kdocs, I think that this might be a good candidate for an example.

The rest of the kdocs look nice though.

application/src/main/kotlin/net/corda/v5/application/messaging/FlowMessaging.kt
```diff
@@ -79,3 +116,18 @@ interface FlowMessaging {
     @Suspendable
     fun close(sessions: Set<FlowSession>)
 }
+
+/**
+ * Builder of context properties. Instances of this interface can optionally be passed to [FlowMessaging] when sessions
+ * are initiated if there are requirements to modify context properties which are sent to the initiated flow.
+ */
+fun interface FlowContextPropertiesBuilder {
```
lankydan: Let's put this in its own file.

application/src/main/kotlin/net/corda/v5/application/messaging/FlowSession.kt
```diff
@@ -31,6 +37,23 @@ interface FlowSession {
...
+     *
+     * If this session was passed to an initiated flow by Corda, the context properties are associated with the
+     * initiating flow at the other end of the connection. They differ from the [FlowContextProperties] available to all
+     * flows via the [FlowEngine] in that they are not a description of the context of the currently executing flow, but
+     * instead the flow which initiated it.
+     *
+     * Any calls to modify these contextProperties will throw a [CordaRuntimeException], they should be considered
+     * immutable.
+     */
+    val contextProperties: FlowContextProperties
```
lankydan, line 55: Mmmmm, so if `FlowContextProperties` should only be modified in `FlowContextPropertiesBuilder`, should we change things a bit so `FlowContextProperties` cannot be modified at all by users? 

Hopefully, this isn't treading on things we already talked about since we discussed having a read-only version before hand. 

However, with the introduction of the builder, we can have a read only version without needing changes to `FlowSession`.

If `FlowContextPropertiesBuilder` was actually a builder, then that can be where it is mutable and then after calling a `.build` method it becomes a `FlowContextProperties` which only has a `get`. Annoyingly, the naming issue becomes if `FlowContextPropertiesBuilder` is a real builder, then the functional interface you current have named `FlowContextPropertiesBuilder` would need a different name itself, and needs to pass the "new" `FlowContextPropertiesBuilder` into its `apply` method.

Doing something like that means you can remove the `Any calls to modify these contextProperties will throw a [CordaRuntimeException], they should be considered immutable.` clause you have here.

But, as always, assuming what I've said makes any sort of sense, dealing with the naming problem here doesn't look simple...

lankydan: Happy to discuss this if needed @simon-johnson-r3.

application/src/main/kotlin/net/corda/v5/application/messaging/FlowMessaging.kt
simon-johnson-r3: yeah, we discussed already and I forgot, will add

lankydan (approved):

PR #545, opened 2022-09-01
CORE-6380: activate publishing to S3 bucket for corda-api

wzur-r3:
* use newly introduced functionality of shared pipeline and set
  `publishToMavenS3Repository`
* removed obsolete conditional activation of source jar, always create
  them


ronanbrowne (approved):
LGTM

PR #547, opened 2022-09-01
CORE-6146 - add configuration for websocket idle timeout

conal-smith:
https://github.com/corda/corda-runtime-os/pull/2002

### DIFF
--- a/data/config-schema/src/main/resources/net/corda/schema/configuration/rpc/1.0/corda.rpc.json
+++ b/data/config-schema/src/main/resources/net/corda/schema/configuration/rpc/1.0/corda.rpc.json
@@ -55,7 +55,7 @@
       "default": {},
       "properties": {
         "idleTimeoutMs": {
-          "description": "Amount of time in milliseconds for a websocket connection to be idle before timing out and closing the connection.",
+          "description": "Amount of time in milliseconds for a websocket connection to be idle (i.e. no messages are either sent or received) before timing out and closing the connection. Value of 0 means such idle check will be disabled.",
           "type": "integer",
           "minimum": 0,
           "maximum": 2147483647,
--- a/data/config-schema/src/main/resources/net/corda/schema/configuration/rpc/1.0/corda.rpc.json
+++ b/data/config-schema/src/main/resources/net/corda/schema/configuration/rpc/1.0/corda.rpc.json
@@ -48,6 +48,20 @@
           "default": 12000
         }
       }
+    },
+    "websocket": {
+      "description": "Configuration settings for WebSockets.",
+      "type": "object",
+      "default": {},
+      "properties": {
+        "idleTimeoutMs": {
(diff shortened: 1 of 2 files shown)


data/config-schema/src/main/resources/net/corda/schema/configuration/rpc/1.0/corda.rpc.json
```diff
@@ -48,6 +48,20 @@
...
+    },
+    "websocket": {
+      "description": "Configuration settings for WebSockets.",
+      "type": "object",
+      "default": {},
+      "properties": {
+        "idleTimeoutMs": {
+          "description": "Amount of time in milliseconds for a websocket connection to be idle before timing out and closing the connection.",
+          "type": "integer",
+          "minimum": 0,
```
vkolomeyko, line 60: Shall put something like:
```suggestion
          "minimum": 100,
```
or else value of `0` may be interpreted as "never".

data/config-schema/src/main/resources/net/corda/schema/configuration/rpc/1.0/corda.rpc.json
```diff
@@ -48,6 +48,20 @@
...
         }
       }
+    },
+    "websocket": {
+      "description": "Configuration settings for WebSockets.",
+      "type": "object",
+      "default": {},
+      "properties": {
+        "idleTimeoutMs": {
+          "description": "Amount of time in milliseconds for a websocket connection to be idle before timing out and closing the connection.",
```
vkolomeyko: Value of `0` will be interpreted as "never". From `org.eclipse.jetty.io.EndPoint#setIdleTimeout`: `Params:
idleTimeout – the idle timeout in MS. Timeout <= 0 implies an infinite timeout`.

Therefore:
```suggestion
          "description": "Amount of time in milliseconds for a websocket connection to be idle (i.e. no messages are either sent or received) before timing out and closing the connection. Value of 0 means such idle check will be disabled.",
```

gradle.properties
```diff
@@ -7,7 +7,7 @@ cordaProductVersion = 5.0.0
 # NOTE: update this each time this module contains a breaking change
 ## NOTE: currently this is a top level revision, so all API versions will line up, but this could be moved to
 ##   a per module property in which case module versions can change independently.
-cordaApiRevision = 172
+cordaApiRevision = 173
```
vkolomeyko: The version has moved on by now.

data/config-schema/src/main/resources/net/corda/schema/configuration/rpc/1.0/corda.rpc.json
sprintengine: Well spotted, thanks!

vkolomeyko (dismissed):
LGTM

vkolomeyko (approved):
LGTM

PR #549, opened 2022-09-02
CORE-6475 API Documentation, updated KDocs with better description of interfaces for custom digests

alexey-kadyrov-r3:


driessamyn (approved):
Changes LGTM.

Is this actually supported currently? If not, should we remove this (or can we, for example, not publish the JAR) until it is supported?
@mnesbit , do you have a view?

alexey-kadyrov-r3:
Yes it's currently supported, there are several integration tests

PR #556, opened 2022-09-07
CORE-6472 - add new event schema for successful mgm onboarding and fix docs

nikinagy:
https://r3-cev.atlassian.net/browse/CORE-6472

We need to wait until MGM is registered to return callback to link manager in order to have the tls trustroots.
An event has been added to signal when the MGM finished its onboarding.


yift-r3 (dismissed):
LGTM but we need to update the version number

yift-r3 (dismissed):

yift-r3 (dismissed):

yift-r3 (approved):

dimosr (approved):

PR #558, opened 2022-09-07
CORE-6627: Fix Java incompatibility of our serialization API.

chrisr3:
The `CordaSerializationTransformEnumDefault` annotation _cannot_ be used from Java because its `new` parameter clashes with a Java reserved word. Rename this to `newName`, and its companion to `oldName`.

Also refactor `ConstructorForDeserialization` and `DeprecatedConstructorForDeserialization` into the `base` module so that they are available for serializable Java classes in `base ` that have multiple constructors.

And declare `CordaSerializationTransformEnumDefault` and `CordaSerializationTransformRename` as `@Repeatable` annotations, because they are!

Rewriting this small module in Java prevents any further Java incompatibilities from creeping back in. Kotlin's nullability checks have been preserved by applying JetBrains' `@Nullable` and `@NotNull` annotations appropriately, effectively replicating what the Kotlin compiler used to do.


dickon (approved):
LGTM

ronanbrowne (approved):
LGTM

PR #559, opened 2022-09-08
CORE-6250 Context Propagation to External processors

simon-johnson-r3:
* Flow Context is added as a single flat map to external event context.
* Not in scope of this story how those processors use the context. As little is populated right now, this is more about plumbing it in that using it usefully.


LWogan (dismissed):
Makes sense to me.

mnesbit (dismissed):
LGTM

lankydan (dismissed):

lankydan (approved):

PR #561, opened 2022-09-08
CORE-6651 Remove `FlowException`

lankydan:
`FlowException` is not being used anymore, since we are not encouraging the sending of exceptions between sessions to communicate errors. Instead they should pass around POJOs that signify an error.

`FlowException`s specified an exception as needing to be seen as it was on a peer flow. This behaviour is no longer required.


mnesbit (dismissed):
OK

ledger-obsolete/src/main/kotlin/net/corda/v5/ledger/obsolete/contracts/TransactionVerificationException.kt
```diff
@@ -33,26 +31,26 @@ open class TransactionResolutionException(val hash: SecureHash, message: String)
...
 class BrokenAttachmentException(val attachmentId: AttachmentId, message: String?, cause: Throwable?)
-    : FlowException("Attachment $attachmentId has error (${message ?: "no message"})", cause)
+    : CordaRuntimeException("Attachment $attachmentId has error (${message ?: "no message"})", cause)
 
 /**
  * Indicates that some aspect of the transaction named by [txId] violates the platform rules. The exact type of failure
- * is expressed using a subclass. TransactionVerificationException is a [FlowException] and thus when thrown inside
- * a flow, the details of the failure will be serialised, propagated to the peer and rethrown.
+ * is expressed using a subclass. TransactionVerificationException is a [CordaRuntimeException] and thus when thrown
+ * inside a flow, the details of the failure will be serialised, propagated to the peer and rethrown.
```
blsemo: Is that behaviour still true with the change to `CordaRuntimeException`?

blsemo (dismissed):
Generally fine with it, but have on question - see comment in the code.

ledger-obsolete/src/main/kotlin/net/corda/v5/ledger/obsolete/contracts/TransactionVerificationException.kt
lankydan: Yes, that is wrong. Want me to change it, or ignore it since its in the obsolete module?

blsemo (approved):

PR #563, opened 2022-09-08
CORE-6627: Fix publishing of Java artifacts.

chrisr3:
I've updated the `corda.common-publishing` convention plugin _not_ to create a `sourcesJar` task, should one already exist. We instruct the `java` Gradle plugin to publish a `sources` artifact, but I suspect the Kotlin Gradle plugin discards it. Or else Gradle just adds the `sources` artifact to `components.java` already? However, now that the `:serialization` module is pure Java we end up with two `sources` artifacts, which is forbidden of course.

We also need to instruct `java-only` modules to create a `Javadoc` artifact, because Dokka only works for Kotlin modules. Which has given me some Javadoc errors to fix too.

And I've also discovered that we must apply the Kotlin plugin _before_ we apply `corda.common-publishing`, or else we don't publish the main artifact!? This could be another example of multiple `Project.afterEvaluate {}` handlers executing in a unspecified order tripping us up. I _cannot_ find any other explanation.

### DIFF
--- a/application/build.gradle
+++ b/application/build.gradle
@@ -1,7 +1,7 @@
 plugins {
     id 'corda-api.common-library'
-    id 'corda.common-publishing'
     id 'org.jetbrains.kotlin.jvm'
+    id 'corda.common-publishing'
 }
 
 description 'Corda Application'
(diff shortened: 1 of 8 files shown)


ronanbrowne (approved):
LGTM

PR #564, opened 2022-09-09
CORE-6657: Handle checked exceptions in MemberX500Name

YashNabar:
https://r3-cev.atlassian.net/browse/CORE-6657

### DIFF
--- a/base/src/main/kotlin/net/corda/v5/base/types/MemberX500Name.kt
+++ b/base/src/main/kotlin/net/corda/v5/base/types/MemberX500Name.kt
@@ -3,6 +3,7 @@ package net.corda.v5.base.types
 import net.corda.v5.base.annotations.CordaSerializable
 import java.util.Locale
 import java.util.Objects
+import javax.naming.InvalidNameException
 import javax.naming.directory.BasicAttributes
 import javax.naming.ldap.LdapName
 import javax.naming.ldap.Rdn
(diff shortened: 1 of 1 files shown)


gradle.properties
```diff
@@ -7,7 +7,7 @@ cordaProductVersion = 5.0.0
 # NOTE: update this each time this module contains a breaking change
 ## NOTE: currently this is a top level revision, so all API versions will line up, but this could be moved to
 ##   a per module property in which case module versions can change independently.
-cordaApiRevision = 179
+cordaApiRevision = 181
```
nikinagy: Do we really need to increase the api version here? I think this is only for breaking changes which require a second PR on runtime-os

nikinagy (dismissed):

gradle.properties
YashNabar: the version compatibility PR gate was failing for this without the version bump

YashNabar: Passes without it now, removed

base/src/main/kotlin/net/corda/v5/base/types/MemberX500Name.kt
```diff
@@ -126,23 +127,27 @@ class MemberX500Name(
...
+                        }
+                        require(!result.containsKey(it.id)) {
+                            "Duplicate attribute ${it.id}"
+                        }
+                        result[it.id] = value
                     }
-                    result[it.id] = value
                 }
+            } catch (e: InvalidNameException) {
+                throw IllegalArgumentException(e)
```
chrisr3: I think the convention is to include the exception message too so that it's not buried or lost:
```kotlin
throw new IllegalArgumentException(e.message, e)
```

base/src/main/kotlin/net/corda/v5/base/types/MemberX500Name.kt
```diff
@@ -225,23 +230,29 @@ class MemberX500Name(
...
+                organisationUnit?.let {
+                    add(Rdn(BasicAttributes(ATTRIBUTE_ORGANISATION_UNIT, it)))
+                }
+                commonName?.let {
+                    add(Rdn(BasicAttributes(ATTRIBUTE_COMMON_NAME, it)))
+                }
             }
+            X500Principal(LdapName(rdns).toString())
+        } catch (e: InvalidNameException) {
+            throw IllegalArgumentException(e)
```
chrisr3: Same as above - needs exception message too.

chrisr3 (changes requested):
Small thing...

base/src/main/kotlin/net/corda/v5/base/types/MemberX500Name.kt
YashNabar: Updated

nikinagy (dismissed):
LGTM

chrisr3 (approved):

dimosr (approved):

PR #566, opened 2022-09-09
CORE-4980 deleted hsmId and fix for secrets

alexey-kadyrov-r3:
Secrets are converted to a string

### DIFF
--- a/data/config-schema/src/main/resources/net/corda/schema/configuration/cryptoLibrary/1.0/corda.cryptoLibrary.json
+++ b/data/config-schema/src/main/resources/net/corda/schema/configuration/cryptoLibrary/1.0/corda.cryptoLibrary.json
@@ -142,11 +142,6 @@
       },
       "additionalProperties": false
     },
-    "hsmId": {
-      "description": "The HSM's id which is handled by this instance, the value must match one of the keys in the hsmMap",
-      "type": "string",
-      "default": "SOFT"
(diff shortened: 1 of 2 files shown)


data/config-schema/src/main/resources/net/corda/schema/configuration/cryptoLibrary/1.0/corda.cryptoLibrary.json
```diff
@@ -306,16 +301,8 @@
...
-                                  "encryptedSecret": {
-                                    "type": "string"
-                                  }
-                                },
-                                "additionalProperties": false,
-                                "required": [
-                                  "encryptedSecret"
-                                ]
+                                "type": ["object", "string"],
+                                "additionalProperties": true
```
driessamyn: I don't think this is right.
I think `passphrase` needs to be of type `string`. 
Turning a string into a secret is optional and we don't expose that in the schema.

alexey-kadyrov-r3: It was changed to string now as the https://github.com/corda/corda-runtime-os/pull/2071 was merged

kyriathar (dismissed):

driessamyn (approved):
API Change looks ok.

PR #569, opened 2022-09-13
CORE-4869 - Add marker for rejected messages

dimosr:
Adding a marker that can be used to indicate messages that were discarded. This is going to be used by the link manager for scenarios of invalid messages.

### DIFF
--- a/data/avro-schema/src/main/resources/avro/net/corda/data/p2p/markers/LinkManagerDiscardedMarker.avsc
+++ b/data/avro-schema/src/main/resources/avro/net/corda/data/p2p/markers/LinkManagerDiscardedMarker.avsc
@@ -1,0 +1,18 @@
+{
+  "type": "record",
+  "name": "LinkManagerDiscardedMarker",
+  "namespace": "net.corda.p2p.markers",
+  "fields": [
+    {
+      "name": "message",
+      "type": "net.corda.p2p.AuthenticatedMessageAndKey",
+      "doc": "A copy of the original message in P2P_OUT_TOPIC."
+    },
+    {
+      "name": "reason",
+      "type": "string",
+      "doc": "The reason why the message was discarded."
+    }
+  ],
+  "doc": "The message was discarded by the sending Link Manager (e.g. because it was invalid)."
+}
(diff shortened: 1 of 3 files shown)


data/avro-schema/src/main/resources/avro/net/corda/data/p2p/markers/LinkManagerDiscardedMarker.avsc
```diff
@@ -0,0 +1,18 @@
+{
+  "type": "record",
+  "name": "LinkManagerDiscardedMarker",
```
yift-r3, line 3: Do we need to add it to the `AppMessageMarker`?

dimosr: of course - good catch. I'll add it

yift-r3 (dismissed):
LGTM (just need to do the api version change dance)

yift-r3 (dismissed):

data/avro-schema/src/main/resources/avro/net/corda/data/p2p/markers/LinkManagerDiscardedMarker.avsc
```diff
@@ -0,0 +1,18 @@
...
+  "namespace": "net.corda.p2p.markers",
+  "fields": [
+    {
+      "name": "message",
```
williamvigorr3, line 7: Do you need the entire message? We use a union of `AuthenticatedMessageHeader` and `UnauthenticatedMessageHeader` to save space.

dimosr: was a bit on the fence between adding the full message or just the message ID. Settled on the former thinking that it would be useful for investigation purposes given these are invalid messages, which shouldn't be too frequent.

williamvigorr3: Fair enough.

williamvigorr3 (dismissed):
LGTM

ronanbrowne (approved):
LGTM

PR #570, opened 2022-09-13
CORE-6325 - Document Membership

nikinagy:
Addressing the review comments under:
https://r3-cev.atlassian.net/browse/CORE-6305

And some additional changes like properly documenting MemberX500Name, LayeredPropetyMap which are outside of the membership module, but still they are mainly belonging to our team.

Guidelines and how to get the final doc pages can be found here: https://r3-cev.atlassian.net/wiki/spaces/CB/pages/4106027171/Guidance+for+completing+Corda-API+KDocs

Adding @alexey-kadyrov-r3 too as in some cases I put him as author based on simply just the latest git blame.

### DIFF
--- a/base/src/main/kotlin/net/corda/v5/base/types/LayeredPropertyMap.kt
+++ b/base/src/main/kotlin/net/corda/v5/base/types/LayeredPropertyMap.kt
@@ -22,18 +22,16 @@ package net.corda.v5.base.types
  * The default implementation of the [LayeredPropertyMap] is extendable by supplying implementations of custom
  * converters using OSGi. Out of box it supports conversion to simple types like Int, Boolean,
  * as well as MemberX500Name.
+ *
+ * @property entries Returns [Set] of all entries in the underlying map.
  */
 interface LayeredPropertyMap {
+    val entries: Set<Map.Entry<String, String?>>
 
     /**
      * Returns the value of the given key or null if the key doesn't exist.
      */
     operator fun get(key: String): String?
-
-    /**
-     * Returns [Set] of all entries in the underlying map.
-     */
-    val entries: Set<Map.Entry<String, String?>>
 
     /**
      * Converts the value of the given key to the specified type.
--- a/base/src/main/kotlin/net/corda/v5/base/types/LayeredPropertyMap.kt
+++ b/base/src/main/kotlin/net/corda/v5/base/types/LayeredPropertyMap.kt
@@ -42,9 +42,6 @@ package net.corda.v5.base.types
  * as well as [MemberX500Name].
  *
  * @property entries Returns [Set] of all entries in the underlying map.
- *
- * @author Alexey Kadyrov
- * @since DP2
  */
 interface LayeredPropertyMap {
     val entries: Set<Map.Entry<String, String?>>
@@ -55,9 +52,6 @@ interface LayeredPropertyMap {
      * @param key Key for the entry we are looking for.
      *
      * @return The value of the given key or null if the key doesn't exist.
-     *
-     * @author Alexey Kadyrov
-     * @since DP2
      */
     operator fun get(key: String): String?
 
@@ -73,9 +67,6 @@ interface LayeredPropertyMap {
      * the [T] is different from it was called for the first time.
      *
      * @return The parsed values for given type.
-     *
-     * @author Alexey Kadyrov
-     * @since DP2
      */
     fun <T> parse(key: String, clazz: Class<out T>): T
 
@@ -91,9 +82,6 @@ interface LayeredPropertyMap {
      * the [T] is different from it was called for the first time.
      *
      * @return The parsed values for given type or null if the key doesn't exist.
-     *
-     * @author Alexey Kadyrov
-     * @since DP2
      * */
     fun <T> parseOrNull(key: String, clazz: Class<out T>): T?
 
@@ -118,9 +106,6 @@ interface LayeredPropertyMap {
      *  corda.endpoints.3.protocolVersion = 1
      *
      *  @return A parsed list of elements for given type.
-     *
-     *  @author Alexey Kadyrov
-     *  @since DP2
      */
     fun <T> parseList(itemKeyPrefix: String, clazz: Class<out T>): List<T>
 
@@ -142,9 +127,6 @@ interface LayeredPropertyMap {
      *  corda.ledgerKeyHashes.3 = <hash value of ledger key 3>
      *
      * @return A parsed set of elements for given type.
-     *
-     * @author Yash Nabar
-     * @since DP2
      */
     fun <T> parseSet(itemKeyPrefix: String, clazz: Class<out T>): Set<T>
 }
--- a/base/src/main/kotlin/net/corda/v5/base/types/LayeredPropertyMap.kt
+++ b/base/src/main/kotlin/net/corda/v5/base/types/LayeredPropertyMap.kt
@@ -7,8 +7,8 @@ package net.corda.v5.base.types
  * The layered property map provides simple conversions to a possibly complex objects which can use several keys in
  * dot-notation. Take as an example of the map:
  *
- * - "corda.name" to "CN=me, O=R3, L=Dublin, C=Ireland",
- * - "corda.sessionKey" to "ABCDEF...",
+ * - "corda.name" to "CN=me, O=R3, L=Dublin, C=IE",
+ * - "corda.sessionKey" to "-----BEGIN PUBLIC KEY-----Base64–encoded public key-----END PUBLIC KEY-----",
  * - "corda.endpoints.0.url" to "localhost",
  * - "corda.endpoints.0.protocolVersion" to "1",
  * - "corda.endpoints.1.url" to "localhost",
--- a/base/src/main/kotlin/net/corda/v5/base/types/LayeredPropertyMap.kt
+++ b/base/src/main/kotlin/net/corda/v5/base/types/LayeredPropertyMap.kt
@@ -7,39 +7,75 @@ package net.corda.v5.base.types
  * The layered property map provides simple conversions to a possibly complex objects which can use several keys in
  * dot-notation. Take as an example of the map:
  *
- * "corda.name" to "CN=me, O=R3, L=Dublin, C=Ireland",
- * "corda.sessionKey" to "ABCDEF...",
- * "corda.endpoints.0.url" to "localhost",
- * "corda.endpoints.0.protocolVersion" to "1",
- * "corda.endpoints.1.url" to "localhost",
- * "corda.endpoints.1.protocolVersion" to "2"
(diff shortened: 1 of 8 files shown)


base/src/main/kotlin/net/corda/v5/base/types/LayeredPropertyMap.kt
```diff
@@ -7,57 +7,102 @@ package net.corda.v5.base.types
...
+ * ```
+ *
  * The default implementation of the [LayeredPropertyMap] is extendable by supplying implementations of custom
  * converters using OSGi. Out of box it supports conversion to simple types like Int, Boolean,
- * as well as MemberX500Name.
+ * as well as [MemberX500Name].
+ *
+ * @property entries Returns [Set] of all entries in the underlying map.
+ *
+ * @author Alexey Kadyrov
```
alexey-kadyrov-r3: I think we are not using @author tag

base/src/main/kotlin/net/corda/v5/base/types/LayeredPropertyMap.kt
```diff
@@ -7,57 +7,102 @@ package net.corda.v5.base.types
...
+ * @since DP2
```
alexey-kadyrov-r3: Not sure that we are using  @since either

base/src/main/kotlin/net/corda/v5/base/types/LayeredPropertyMap.kt
nikinagy: As i understood from the guidelines, we should use them:
 https://r3-cev.atlassian.net/wiki/spaces/CB/pages/4106027171/Guidance+for+completing+Corda-API+KDocs

base/src/main/kotlin/net/corda/v5/base/types/LayeredPropertyMap.kt
nikinagy: Same comment as above.

base/src/main/kotlin/net/corda/v5/base/types/MemberX500Name.kt
```diff
@@ -16,32 +16,120 @@ import javax.security.auth.x500.X500Principal
...
+ * val nameByPrincipal = MemberX500Name.build(principalForNewName)
+ * val nameByParse = MemberX500Name.parse(name)
+ *
+ * val map = MemberX500Name.toAttributesMap("CN=alice, OU=Accounting, O=R3, L=Seattle, ST=Washington, C=US")
+ * ```
+ *
+ * @author Alexey Kadyrov
+ * @since DP2
+ *
+ * @param commonName Optional name by which the entity is usually known. Used only for services (for
```
alexey-kadyrov-r3: Not clear what is `service` and what is `organisation` and how they fit the member name. There should be some explanation and reference how to find that information.

base/src/main/kotlin/net/corda/v5/base/types/MemberX500Name.kt
```diff
@@ -16,32 +16,120 @@ import javax.security.auth.x500.X500Principal
...
+ * @param state The full name of the state or province the organisation is based in. Corresponds to the "ST"
  * attribute type.
- * @property country country the organisation is in, as an ISO 3166-1 2-letter country code. Corresponds to the "C"
+ * @param country Country the organisation is in, as an ISO 3166-1 2-letter country code. Corresponds to the "C"
  * attribute type.
 */
 @Suppress("LongParameterList")
 @CordaSerializable
 class MemberX500Name(
+    /**
```
alexey-kadyrov-r3, line 97: Out of curiosity why the @property where changed to @param ?

base/src/main/kotlin/net/corda/v5/base/types/LayeredPropertyMap.kt
alexey-kadyrov-r3: There was a discussion for that attribute in the doc - it's useless in the context of R3, e.g. I'm not an author of the MemberX500Name as it was ported from C4 codebase with some minor modifications.

base/src/main/kotlin/net/corda/v5/base/types/LayeredPropertyMap.kt
alexey-kadyrov-r3: Here as well, some of the APIs are from C4 some new and DP2 of which version, C5? As there is no real guidance on how the value should be determined juts don't set until that is clarified.

@nadinequinr3 there must be clear guidance on how to set that tag which is currently doesn't exists. I would be of the opinion that we need it but probably it's premature ATM

base/src/main/kotlin/net/corda/v5/base/types/MemberX500Name.kt
nikinagy: Because without @param, this constructor stayed undocumented in the generated html pages.

base/src/main/kotlin/net/corda/v5/base/types/MemberX500Name.kt
```diff
@@ -16,32 +16,120 @@ import javax.security.auth.x500.X500Principal
  * The class also guaranties the reliable equality comparison regardless which order the attributes are specified when
  * parsing from the string or X500principal as well outputs the attributes to string in predictable order.
  *
- * @property commonName optional name by the which the entity is usually known. Used only for services (for
+ * Example usages:
```
alexey-kadyrov-r3, line 26: It would be useful to specify the constrains, what values are parsed, which are required, which are optional and in which order the name would be built. That exists in some form but it would be difficult to piece together the different bits to get the overall picture.

base/src/main/kotlin/net/corda/v5/base/types/MemberX500Name.kt
nikinagy: They are still listed as properties as well, by putting the comments above each fields.

alexey-kadyrov-r3: That need to be brought to attention to all developers as in the most cases we document those properties using just @property tag

However there is a slight chance that the doc generation is not used correctly and the tool which we use to do it is capable generating the docs correctly - but I'm not exert so have no idea.

base/src/main/kotlin/net/corda/v5/base/types/LayeredPropertyMap.kt
nadinequinr3: @mattbradburyr3 As discussed, I think we should remove @author as a requirement?

base/src/main/kotlin/net/corda/v5/base/types/LayeredPropertyMap.kt
nadinequinr3: @mattbradburyr3 Do we want use "Corda 5 DP2" in @ since for everything? Does it matter if it was modified from Corda 4 code?

nadinequinr3 (dismissed):

base/src/main/kotlin/net/corda/v5/base/types/MemberX500Name.kt
```diff
@@ -9,39 +9,130 @@ import javax.naming.ldap.Rdn
...
  * parsing from the string or X500principal as well outputs the attributes to string in predictable order.
  *
- * @property commonName optional name by the which the entity is usually known. Used only for services (for
- * organisations, the [organisation] property is the name). Corresponds to the "CN" attribute type.
- * @property organisationUnit optional name of a unit within the [organisation]. Corresponds to the "OU" attribute type.
- * @property organisation name of the organisation. Corresponds to the "O" attribute type.
- * @property locality locality of the organisation, typically the nearest major city. For distributed services this would be
- * where one of the organisations is based. Corresponds to the "L" attribute type.
- * @property state the full name of the state or province the organisation is based in. Corresponds to the "ST"
+ * There may be additional network specific requirements which need to be taken into account when creating a name.
```
alexey-kadyrov-r3: I don’t particular like that statement

```
 * There may be additional network specific requirements which need to be taken into account when creating a name.
 * For example, the network operator may require a particular format for names so that they can issue suitable certificates.
 ```

As it implies that the class can do something about that but in reality that is something which is external to this class, so if you reword it in a way that is clear that it must be done outside that class.

File: base/src/main/kotlin/net/corda/v5/base/types/LayeredPropertyMap.kt [hunk 7]
```diff
@@ -7,57 +7,90 @@ package net.corda.v5.base.types
  * The layered property map provides simple conversions to a possibly complex objects which can use several keys in
  * dot-notation. Take as an example of the map:
  *
- * "corda.name" to "CN=me, O=R3, L=Dublin, C=Ireland",
- * "corda.sessionKey" to "ABCDEF...",
- * "corda.endpoints.0.url" to "localhost",
- * "corda.endpoints.0.protocolVersion" to "1",
- * "corda.endpoints.1.url" to "localhost",
- * "corda.endpoints.1.protocolVersion" to "2"
+ * - "corda.name" to "CN=me, O=R3, L=Dublin, C=Ireland",
```
YashNabar: C attribute should be a two-letter country code, in this case "IE".

base/src/main/kotlin/net/corda/v5/base/types/LayeredPropertyMap.kt
```diff
@@ -7,57 +7,90 @@ package net.corda.v5.base.types
...
+ * - "corda.sessionKey" to "ABCDEF...",
```
YashNabar: Can we include a truncated key here starting with "BEGIN PUBLIC KEY..."?

base/src/main/kotlin/net/corda/v5/base/types/MemberX500Name.kt
```diff
@@ -76,9 +169,13 @@ class MemberX500Name(
...
 
         /**
-         * Creates an instance of [MemberX500Name] from specified [X500Principal]
+         * Creates an instance of [MemberX500Name] from specified [X500Principal].
+         *
+         * @param principal The X500 principal used for building [MemberX500Name].
          *
          * @throws [IllegalArgumentException] if required attributes are missing, constrains are not satisfied.
+         *
+         * @return [MemberX500Name] based on param.
```
YashNabar: Would it be better to link the actual param here, i.e. [principal], instead of "param"? Same for other methods in this class

base/src/main/kotlin/net/corda/v5/base/types/LayeredPropertyMap.kt
nikinagy: Nice catch!

alexey-kadyrov-r3 (dismissed):
LGTM

YashNabar (approved):

alexey-kadyrov-r3 (approved):
LGTM

PR #573, opened 2022-09-13
CORE-6376: use DP2 specific downstream job to ensure compatibility

ronanbrowne:
- downstream job we use should not be the `release/os/5.0` job but rather the DP2 specific branch for the compatibility job


yarivus (approved):
LGTM

PR #577, opened 2022-09-14
CORE-6561 remove method not used nor needed

LWogan:
Remove the decodePublicKey method from the SigingingService. This was a tmp method used to help get public keys. Public key objects can be gotten now via memberInfo from the member lookup service so this method is no longer required


JamesHR3 (commented):
Guessing this will need a version update but looks fine in principle.

@lankydan, when you're back, I wouldn't mind a quick verification that we don't need this any more.

lankydan:
Is there a runtime PR?

```kotlin
@Suspendable
    private fun signAndVerify(input: RpcSmokeTestInput): String {
        val publicKey = signingService.decodePublicKey(input.getValue("publicKey"))
        val bytesToSign = byteArrayOf(1, 2, 3, 4, 5)
        log.info("Crypto - Signing bytes $bytesToSign with public key '$publicKey'")
        val signedBytes = signingService.sign(bytesToSign, publicKey, SignatureSpec.RSA_SHA256)
        log.info("Crypto - Signature $signedBytes received")
        digitalSignatureVerificationService.verify(publicKey, SignatureSpec.RSA_SHA256, signedBytes.bytes, bytesToSign)
        log.info("Crypto - Verified $signedBytes as the signature of $bytesToSign")
        return true.toString()
    }
```

JamesHR3:
Runtime PR here: https://github.com/corda/corda-runtime-os/pull/2108

JamesHR3 (commented):
This will need updating to the right version number.

JamesHR3 (approved):
Would you mind leaving a description in the commit about what method was removed and why it's not needed?

PR #595, opened 2022-09-19
CORE-6779

MrMatthewLayton:
Added Contract Interfaces and Tests
**NOTE : PR AND MERGE CORE-6778 BEFORE THIS ONE!**


yarivus (approved):
LGTM

ledger/ledger-utxo/src/main/kotlin/net/corda/v5/ledger/utxo/Contracts.kt
```diff
@@ -26,3 +27,29 @@ interface Contract {
...
+}
+
+/**
+ * Defines a command, and the signatories associated with the specified command.
+ *
+ * @property command The command to verify.
+ * @property signatories The signatory signing keys associated with the specified command.
+ */
+@CordaSerializable
+interface CommandAndSignatories<out T : Command> {
```
mnesbit: This is problematic in terms of serialization and indeed doesn't work in C4 due to need to see the required signers before knowing how to deserialize commands. So probably the information structure needs some thinking. I am fine with this as a synthesized view onto the wire/ledger transaction commands, but really they will end up in two (same sized) component groups

ledger/ledger-utxo/src/main/kotlin/net/corda/v5/ledger/utxo/Contracts.kt
```diff
@@ -26,3 +27,29 @@ interface Contract {
...
+ */
+interface VerifiableCommand : Command {
+
+    /**
+     * Verifies the specified transaction associated with the current contract command.
+     *
+     * @param transaction The transaction to verify.
+     * @param signatories The signatories associated with the current contract command.
+     */
+    fun verify(transaction: UtxoLedgerTransaction, signatories: Iterable<PublicKey>)
```
mnesbit: UtxoLedgerTransaction will contain the commands and signatories, so this is getting a bit circular. Perhaps we just pass in the self index in the list of commands

ledger/ledger-utxo/src/main/kotlin/net/corda/v5/ledger/utxo/Contracts.kt
```diff
@@ -26,3 +27,29 @@ interface Contract {
  */
 @CordaSerializable
 interface Command
+
+/**
+ * Defines a verifiable command; that is, a command that implements its own contract verification logic.
+ */
+interface VerifiableCommand : Command {
```
lankydan: If we're going to have a few types of `Command`s, then I think a separate file/files and possibly a `commands` package if we split into multiple files.

ledger/ledger-utxo/src/main/kotlin/net/corda/v5/ledger/utxo/Contracts.kt
MrMatthewLayton: If we end up with 2 distinct component groups, 1 for commands and 1 for signatories, how will we distinguish which keys belong to which command, or is this to be decided?

ledger/ledger-utxo/src/main/kotlin/net/corda/v5/ledger/utxo/Contracts.kt
MrMatthewLayton: Happy to split. I actually had everything in it's own file (which I prefer), but the list of files was becoming a bit massive.

ledger/ledger-utxo/src/main/kotlin/net/corda/v5/ledger/utxo/Contracts.kt
mnesbit: The component group will be some holder of a list of PublicKeys so one item for one command

mnesbit: Main thing is to separate things that need to already know the CPI and make a sandbox to be deserialized and things which can be interrogated without. This is especially important when we do partial reveal data structures.

ledger/ledger-utxo/src/main/kotlin/net/corda/v5/ledger/utxo/Contracts.kt
mnesbit: This is a 'new' feature we will have to think a bit in terms of if this is going to give us problems with evolution. The rule for Contract is clear in that we identify what contracts are used on all inputs+outputs and invoke them all. If the input comes from an older version of the contract we require that the new sandbox environment has an equivalent Contract that we can map to, otherwise this is an invalid transaction. I guess we can do the same for commands, but it may be more complex to remember as a coder.

mnesbit: Again my view is that we shouldn't be putting any new design elements into the first few PR's on UTXO ledger, but rather hold off till we are in a position to see the consequences of adding them in

ledger/ledger-utxo/src/main/kotlin/net/corda/v5/ledger/utxo/Contracts.kt
MrMatthewLayton: @mnesbit Not sure what you mean by this?

yarivus (commented):
LGTM

PR #598, opened 2022-09-19
CORE-6782

MrMatthewLayton:
Added StatePointer Interfaces, Implementations and Tests

### DIFF
--- a/ledger/ledger-utxo/src/main/kotlin/net/corda/v5/ledger/utxo/StatePointers.kt
+++ b/ledger/ledger-utxo/src/main/kotlin/net/corda/v5/ledger/utxo/StatePointers.kt
@@ -1,0 +1,347 @@
+package net.corda.v5.ledger.utxo
+
+import net.corda.v5.base.annotations.CordaSerializable
+import java.util.*
+
+/**
+ * Defines a pointer to a [ContractState].
+ *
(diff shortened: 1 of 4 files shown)


mnesbit (commented):
I feel this is an advanced feature that can come in later. I am not sure if this couldn't in fact be deleted if we change way builder works

ledger/ledger-utxo/src/main/kotlin/net/corda/v5/ledger/utxo/StatePointers.kt
```diff
@@ -0,0 +1,347 @@
...
+
+/**
+ * Represents a state pointer that points statically to a specific [StateRef].
+ *
+ * @constructor Creates a new instance of the [StaticPointer] data class.
+ * @property identifier The identifier of the [ContractState] being pointed to.
+ * @property type The type of the [ContractState] being pointed to.
+ * @property isResolved Determines whether the current pointer should be resolved to a reference input when used in a transaction.
+ */
+data class StaticPointer<T : ContractState>(
```
lankydan, line 28: With the amount of code in these classes, they should be split into separate files with a appropriate package.

yarivus (commented):
LGTM

blsemo:
I agree with @mnesbit that we probably don't need that as a day 1 feature.

PR #603, opened 2022-09-20
CORE-6681: update JSoup transitive build plugin Dependency 

ronanbrowne:
Ensure Jsoup which is brought in as a transitive dependency of dokka is bumped to the latest version


yarivus (approved):
LGTM

PR #605, opened 2022-09-21
CORE-6608: ensure plugins and app dependencies are bumped for jsoup

ronanbrowne:
Jsoup is brought in via plugin dependencies and application dependencies, therefore removing buildscript block which just handles the plugin use case and using dependencySubstitution which captures transitive dependencies from build and app dependencies

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -6,15 +6,6 @@ buildscript {
     ext {
         vcsUrl = System.getenv('GIT_URL') ?: 'https://github.com/corda/corda-api.git'
     }
-    
-    dependencies {
-        classpath "org.jetbrains.dokka:dokka-core:$dokkaVersion"
-        constraints {
-            classpath("org.jsoup:jsoup:1.15.3") {
-                because "required until dokka plugin updates it's internal version of jsoup, not fixed as of dokka 1.7.10"
-            }
-        }
-    }
 }
 
 plugins {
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -162,6 +153,12 @@ subprojects {
             }
         }
 
+        configurations.all {
+            resolutionStrategy.dependencySubstitution {
+                substitute module('org.jsoup:jsoup:1.14.3') using module('org.jsoup:jsoup:1.15.3')
```
chrisr3: Do you need to specify the original version here?
```gradle
substitute module('org.jsoup:jsoup') using module("org.jsoup:jsoup:$jsoupVersion")
```
Although I'm not sure that this will still apply to our published artfacts. We'd need a `constraint` for that, which would become a `<dependencyManagement>` element in the published POMs.

ronanbrowne: your right re version , updated,

this is brought in via dokka kdoc generation dependency, dokka seems to land dependencies build on the classpath for plugs (e.g ./gradlew buildEnviorment) and actually in the dependency graph where it creates a number of dokka* configuration decencies. , so I don't think this will be ending up in any artifact.

we can not specifically only scan the runtimeClasspath of corda-api as we do in other projects, as snyk falls over due to this configuration not being available in all submodules as logged here by another user https://github.com/snyk/snyk-gradle-plugin/issues/192

chrisr3 (approved):
LGTM - probably :wink:

mcgovc (approved):
LGTM

PR #607, opened 2022-09-21
CORE-6810 update the DigitalSignatureVerificationService doc to throw the correct exceptions

LWogan:


application/src/main/kotlin/net/corda/v5/application/crypto/DigitalSignatureVerificationService.kt
```diff
@@ -22,10 +20,9 @@ interface DigitalSignatureVerificationService {
      * @param signatureSpec The signature spec.
      * @param clearData The clear data/message that was signed (usually the Merkle root).
      *
-     * @throws InvalidKeyException If the key is invalid.
-     * @throws SignatureException If verification fails.
      * @throws IllegalArgumentException If the signature scheme is not supported or if any of the clear or signature
      * data is empty.
+     * @throws CryptoSignatureException Thrown if verification of the digital signature fails.
```
lankydan: Do "If verification of the digital signature fails." instead.

application/src/main/kotlin/net/corda/v5/application/crypto/DigitalSignatureVerificationService.kt
```diff
@@ -22,10 +20,9 @@ interface DigitalSignatureVerificationService {
      * @param signatureSpec The signature spec.
```
lankydan, line 20: Can you correct the param order while we're here.

JamesHR3 (approved):

PR #608, opened 2022-09-21
CORE-5375: vnode db rollback

benmcmahon100:
Add audit table to keep track of cpi changelogs 

https://github.com/corda/corda-runtime-os/pull/2172


data/db-schema/src/main/resources/net/corda/db/schema/config/migration/cpx-creation-v1.0.xml
```diff
@@ -176,6 +176,34 @@
...
+            <column name="cpk_name" type="VARCHAR(255)">
+                <constraints nullable="false"/>
+            </column>
+            <column name="cpk_version" type="VARCHAR(255)">
+                <constraints nullable="false"/>
+            </column>
+            <column name="cpk_signer_summary_hash" type="VARCHAR(255)">
+                <constraints nullable="false"/>
+            </column>
+            <column name="cpk_file_checksum" type="VARCHAR(255)">
```
sprintengine, line 189: I think we should be able to make `cpk_file_checksum` the PK?

benmcmahon100: I'd had a tonne of thoughts about the pk. Ideally it'd include the entity version, but it doesn't seem to update properly, though I'm convinced that's an issue in the Cpi code and I don't want to start touching that too much.

cpk_file_checksum isn't a runner, as you always end up with multiple entries per checksum.
image

benmcmahon100: You could also upload a cpi, replace it, and the reupload it and it will create the same entires again, so even if you have the checksum and the filepath, that can also reoccur.

sprintengine: Yeah `cpk_file_checksum` as a single PK won't work, as a CPK can run multiple migrations.

benmcmahon100: We may also want to re-run the exact same one from earlier, but probably still want a record of that having been done. Might need to add more info later, or another table to track which migrations are actually ran, but I'm leaving that until a little later

data/db-schema/src/main/resources/net/corda/db/schema/config/migration/cpx-creation-v1.0.xml
```diff
@@ -176,6 +176,40 @@
...
+                <constraints nullable="false"/>
+            </column>
+            <column name="file_path" type="VARCHAR(255)">
+                <constraints nullable="false"/>
+            </column>
+            <column name="content" type="TEXT">
+                <constraints nullable="false"/>
+            </column>
+            <column name="entity_version" type="INT">
```
sprintengine, line 198: I'm not sure about using an optimistic locking `entity_version` column to identify records.

For example, an entity version may increment on read queries if the OPTIMISTIC_FORCE_INCREMENT flag is set.

If there's no other way, at least remove the version annotation from the from entity, I presume you won't be changing any audit records thus won't need optimistic locking.

benmcmahon100: It's not an optimistic locking version though - it's just a copy of the other one. It means you can look up the changelogs for a cpi and just get the ones with the highest entity version in order to know what the last set applied were.

Conscious this comment just might've been out of sync with the version annotation being removed.

data/db-schema/src/main/resources/net/corda/db/schema/config/migration/cpx-creation-v1.0.xml
```diff
@@ -176,6 +176,40 @@
...
+            <column name="is_deleted" type="BOOLEAN">
+                <constraints nullable="false"/>
+            </column>
+            <column name="insert_ts" type="TIMESTAMP"
+                    defaultValueComputed="CURRENT_TIMESTAMP">
+                <constraints nullable="false"/>
+            </column>
+        </createTable>
+
+        <addPrimaryKey tableName="cpk_db_change_log_audit" columnNames="cpk_name, cpk_version, cpk_signer_summary_hash, cpk_file_checksum, entity_version, file_path"
```
driessamyn, line 210: `cpk_file_checksum, entity_version, file_path` may be enough?

data/db-schema/src/main/resources/net/corda/db/schema/config/migration/cpx-creation-v1.0.xml
```diff
@@ -176,6 +176,40 @@
                                  baseTableSchemaName="${schema.name}"
                                  referencedTableSchemaName="${schema.name}"/>
 
+        <createTable tableName="cpk_db_change_log_audit" schemaName="${schema.name}">
```
vkolomeyko, line 179: I think it be worth adding a comment **why** we are introducing another table which looks very similar to `cpk_db_change_log`.

data/db-schema/src/main/resources/net/corda/db/schema/config/migration/cpx-creation-v1.0.xml
benmcmahon100: I don't think so, as you could end up with breakages where you upload cpi a, then b and then a again. Also the key for the original has all those fields minus entity version so thought it made sense to be consistent

data/db-schema/src/main/resources/net/corda/db/schema/config/migration/cpx-creation-v1.0.xml
```diff
@@ -175,6 +175,40 @@
                                  constraintName="FK_cpk_db_change_log_cpk_metadata"
                                  baseTableSchemaName="${schema.name}"
                                  referencedTableSchemaName="${schema.name}"/>
+        <!-- Write only audit table for the cpk_db_change_logs -->
```
vkolomeyko:
```suggestion
        <!-- Append only audit table. Unlike cpk_db_change_logs which will be updated during CPI update process. -->
```

vkolomeyko (approved):
LGTM

data/db-schema/src/main/resources/net/corda/db/schema/config/migration/cpx-creation-v1.0.xml
driessamyn: Y. Makes sense to be consistent.

driessamyn (approved):

PR #613, opened 2022-09-21
CORE-6784

MrMatthewLayton:
Added the majority of the UTXO Transaction types, builder and service (WIP)


.editorconfig
```diff
@@ -0,0 +1,6 @@
+root = true
```
mnesbit: This should not go in

ledger/ledger-utxo/src/main/kotlin/net/corda/v5/ledger/utxo/TransactionComponents.kt
```diff
@@ -33,7 +33,21 @@ import java.util.jar.JarInputStream
  * @property REFERENCES_METADATA The references metadata component group. Ordinal = 14.
  */
 enum class ComponentGroup {
-    TRANSACTION_PARAMETERS, MEMBERSHIP_PARAMETERS, PACKAGES, SIGNATORIES, ATTACHMENTS, NOTARY, TIME_WINDOW, COMMANDS, COMMANDS_METADATA, INPUTS, INPUTS_METADATA, OUTPUTS, OUTPUTS_METADATA, REFERENCES, REFERENCES_METADATA
+    TRANSACTION_PARAMETERS,
```
mnesbit: This enum should not be in corda-api

ledger/ledger-utxo/src/main/kotlin/net/corda/v5/ledger/utxo/transaction/UtxoLedgerService.kt
```diff
@@ -0,0 +1,60 @@
...
+interface UtxoLedgerService {
+
+    /**
+     * Gets a UTXO transaction builder
+     *
+     * @param notary The notary associated with the transaction builder.
+     * @return Returns a new [UtxoTransactionBuilder] instance.
+     */
+    @Suspendable
+    fun getTransactionBuilder(notary: Party): UtxoTransactionBuilder
```
mnesbit: Why do we set notary at this point? Any reason this can't be standard builder semantics?

ledger/ledger-utxo/src/main/kotlin/net/corda/v5/ledger/utxo/transaction/UtxoLedgerService.kt
```diff
@@ -0,0 +1,60 @@
...
+
+    /**
+     * Resolves the specified [StateRef] instances into [StateAndRef] instances of the specified [ContractState] type.
+     *
+     * @param T The underlying [ContractState] type.
+     * @param stateRefs The [StateRef] instances to resolve.
+     * @return Returns a [List] of [StateAndRef] of the specified [ContractState] type.
+     */
+    @Suspendable
+    fun <T : ContractState> resolve(vararg stateRefs: StateRef): List<StateAndRef<T>>
```
mnesbit: This isn't needed if we have the iterable input form, remove as varargs only complicate things at JVM level

ledger/ledger-utxo/src/main/kotlin/net/corda/v5/ledger/utxo/transaction/UtxoLedgerService.kt
```diff
@@ -0,0 +1,60 @@
...
+    @Suspendable
+    fun <T : ContractState> verify(stateAndRefs: Iterable<StateAndRef<T>>)
+
+    /**
+     * Verifies the specified [StateAndRef] instances.
+     *
+     * @param stateAndRefs The [StateAndRef] instances to verify.
+     */
+    @Suspendable
+    fun <T : ContractState> verify(vararg stateAndRefs: StateAndRef<T>)
```
mnesbit: not needed

ledger/ledger-utxo/src/main/kotlin/net/corda/v5/ledger/utxo/transaction/UtxoLedgerTransaction.kt
```diff
@@ -1,6 +1,383 @@
...
+ * @property outputTransactionStates The output transaction states to be created by the current [UtxoLedgerTransaction].
+ * @property outputContractStates The output contract states to be created by the current [UtxoLedgerTransaction].
+ */
 @DoNotImplement
-interface UtxoLedgerTransaction
\ No newline at end of file
+@Suppress("TooManyFunctions")
+interface UtxoLedgerTransaction {
+
+    val timeWindow: TimeWindow?
```
mnesbit, line 34: In C5 this is mandatory and not nullable

ledger/ledger-utxo/src/main/kotlin/net/corda/v5/ledger/utxo/transaction/UtxoLedgerTransaction.kt
```diff
@@ -1,6 +1,383 @@
...
+    val attachments: List<Attachment>
+    val commands: List<CommandAndSignatories<*>>
+
+    val inputStateAndRefs: List<StateAndRef<*>>
+    val inputTransactionStates: List<TransactionState<*>> get() = inputStateAndRefs.map { it.state }
```
mnesbit, line 40: Lets start without all these derived methods

ledger/ledger-utxo/src/main/kotlin/net/corda/v5/ledger/utxo/transaction/UtxoLedgerTransaction.kt
```diff
@@ -1,6 +1,383 @@
...
+
+    /**
+     * Obtains a single ledger transaction [CommandAndSignatories] that matches the specified [Command] type.
+     *
+     * @param T The underlying type of the [Command].
+     * @param type The type of the [Command].
+     * @return Returns a single ledger transaction [CommandAndSignatories] that matches the specified [Command] type.
+     * @throws IllegalArgumentException if a single ledger transaction [CommandAndSignatories] that matches the specified [Command] type cannot be found.
+     */
+    fun <T : Command> getCommandAndSignatories(type: Class<T>): CommandAndSignatories<T>
```
mnesbit: This is ambiguous for multiple commands lets remove for now

ledger/ledger-utxo/src/main/kotlin/net/corda/v5/ledger/utxo/transaction/UtxoLedgerTransaction.kt
```diff
@@ -1,6 +1,383 @@
...
+
+    /**
+     * Obtains a single ledger transaction [StateAndRef] reference input that match the specified [ContractState] type.
+     *
+     * @param T The underlying type of the [ContractState].
+     * @param type The type of the [ContractState].
+     * @return Returns a single ledger transaction [StateAndRef] reference input that match the specified [ContractState] type.
+     * @throws IllegalArgumentException if a single ledger transaction [StateAndRef] reference input that matches the specified [ContractState] type cannot be found.
+     */
+    fun <T : ContractState> getReferenceInputStateAndRef(type: Class<T>): StateAndRef<T>
```
mnesbit: All these getXXX(clazz):T are not really safe with composition. Better to remove the non-list form methods and make developers of verify think about this stuff

ledger/ledger-utxo/src/main/kotlin/net/corda/v5/ledger/utxo/transaction/UtxoLedgerTransaction.kt
```diff
@@ -1,6 +1,383 @@
...
+
+    /**
+     * Obtains a single ledger transaction [ContractState] input that match the specified [ContractState] type.
+     *
+     * @param T The underlying type of the [ContractState].
+     * @param type The type of the [ContractState].
+     * @return Returns a single ledger transaction [ContractState] input that match the specified [ContractState] type.
+     * @throws IllegalArgumentException if a single ledger transaction [ContractState] input that matches the specified [ContractState] type cannot be found.
+     */
+    fun <T : ContractState> getInputState(type: Class<T>): T
```
mnesbit: Lets delete for now

ledger/ledger-utxo/src/main/kotlin/net/corda/v5/ledger/utxo/transaction/UtxoSignedTransaction.kt
```diff
@@ -0,0 +1,54 @@
...
+     */
+    fun addSignatures(vararg signatures: DigitalSignatureAndMetadata): UtxoSignedTransaction
+
+    /**
+     * Gets the missing signatories from the current [UtxoSignedTransaction].
+     *
+     * @param serializer The [SerializationService] required to obtain missing signed transaction signatures.
+     * @return Returns a [Set] of [PublicKey] representing the missing signatories from the current [UtxoSignedTransaction].
+     */
+    fun getMissingSignatories(serializer: SerializationService): Set<PublicKey>
```
mnesbit: Interesting question if serializer should be passed in. It is done to allow user specified serializers, but question is if we are considering that in our logic. I think at this point it is probably too late to change your mind as TxId needs to be same as one in builder. I do think you might need this though if you did need custom formats. Need for this has also been reduced recently, so agin perhaps we defer this and just provide the 'defaulted' no parameter form

ledger/ledger-utxo/src/main/kotlin/net/corda/v5/ledger/utxo/transaction/UtxoTransactionBuilder.kt
```diff
@@ -0,0 +1,228 @@
...
+    fun addOutputState(contractState: ContractState, notary: Party): UtxoTransactionBuilder
+
+    /**
+     * Adds an output state to the current [UtxoTransactionBuilder].
+     *
+     * @param contractState The [ContractState] to add to the current [UtxoTransactionBuilder].
+     * @param contractId The class name of the [Contract] associated with the transaction state.
+     * @return Returns a [UtxoTransactionBuilder] including the additional output state.
+     */
+    fun addOutputState(contractState: ContractState, contractId: String): UtxoTransactionBuilder
```
mnesbit: I would be keen to settle on either annotation based linkage to contracts, or a parameter (clazz, not string?) in builder. Not sure what having both implies.

mnesbit (commented):
A few initial thoughts. Will need to mull over a bit more yet

ledger/ledger-utxo/src/main/kotlin/net/corda/v5/ledger/utxo/TransactionComponents.kt
vlajos: Also WireTransaction assumes that the first component group is Metadata.

ledger/ledger-utxo/src/main/kotlin/net/corda/v5/ledger/utxo/transaction/UtxoSignedTransaction.kt
```diff
@@ -0,0 +1,54 @@
...
+ *
+ * @property id The ID of the transaction.
+ * @property signatures The signatures that have been applied to the transaction.
+ */
+@DoNotImplement
+@CordaSerializable
+interface UtxoSignedTransaction {
+
+    val id: SecureHash
+    val signatures: Set<DigitalSignatureAndMetadata>
```
vlajos, line 20: Consensual friend of this class has some differences:
* signatures is a List<>
* toLedgerTransaction does not have the serializer as an argument. (It will get injected as a constructor parameter) (The same for other methods where serializer will be needed.)

ledger/ledger-utxo/src/main/kotlin/net/corda/v5/ledger/utxo/transaction/UtxoWireTransaction.kt
```diff
@@ -0,0 +1,3 @@
+package net.corda.v5.ledger.utxo.transaction
+
+interface UtxoWireTransaction
```
vlajos: Ideally this file won't be needed.

ledger/ledger-utxo/src/test/java/net/corda/v5/ledger/utxo/AbstractMockTestHarness.java
```diff
@@ -1,10 +1,14 @@
...
+import net.corda.v5.application.crypto.DigitalSignatureAndMetadata;
+import net.corda.v5.application.crypto.DigitalSignatureMetadata;
+import net.corda.v5.application.serialization.SerializationService;
 import net.corda.v5.base.types.MemberX500Name;
 import net.corda.v5.cipher.suite.DigestService;
+import net.corda.v5.crypto.DigitalSignature;
 import net.corda.v5.crypto.SecureHash;
 import net.corda.v5.ledger.common.transaction.Party;
-import net.corda.v5.ledger.utxo.transaction.UtxoLedgerTransaction;
+import net.corda.v5.ledger.utxo.transaction.*;
```
vlajos: AFAIK `*` imports are accepted.

.editorconfig
MrMatthewLayton: @mnesbit `.editorconfig` entirely, or just that line?

ledger/ledger-utxo/src/main/kotlin/net/corda/v5/ledger/utxo/transaction/UtxoLedgerService.kt
MrMatthewLayton: @mnesbit Since all transactions need a notary, it's better that it's not declared nullable. It should be trivial to obtain a notary from either an underlying service, or a pre-existing state before creating a transaction builder.

ledger/ledger-utxo/src/main/kotlin/net/corda/v5/ledger/utxo/transaction/UtxoLedgerService.kt
MrMatthewLayton: @mnesbit Should we have an equivalent singular function instead; for example

```kotlin
fun <T : ContractState> resolve(stateRef: StateRef): StateAndRef<T>
```

ledger/ledger-utxo/src/main/kotlin/net/corda/v5/ledger/utxo/transaction/UtxoLedgerService.kt
MrMatthewLayton: As above, should we have a singular function?

ledger/ledger-utxo/src/main/kotlin/net/corda/v5/ledger/utxo/transaction/UtxoTransactionBuilder.kt
MrMatthewLayton: @mnesbit I was always skeptical about the `BelongsToContract` annotation, because it means that a state only works with one contract. Before, I'm sure it was possible to use states with different contracts; for example:

```kotlin
data class Person(val owner: PublicKey, val name: String, val birthday: Date) : ContractState {
    val participants: Set<PublicKey> get() = setOf(owner)
}

class UpdatableNameContract : Contract {
    // Impl allows create, update (name only) and delete
}

class UpdatableBirthdayContract : Contract {
    // Impl allows create, update (birthday only) and delete
}
```

From the perspective of transaction construction, once your state is created and bound to a particular contract, you have to stick with it until the state is consumed:

```kotlin
val builder = TransactionBuilder(notary)
    .addOutputState(person, UpdatableNameContract::class)
    .addCommand(UpdatableNameContract.Create, key);
```

In a subsequent transaction, you would not be able to switch contract:
```kotlin
val builder = TransactionBuilder(input.state.notary)
    .addInputState(input)
    .addOutputState(person, UpdatableBirthdayContract.class) // NOT ALLOWED!!!
    .addOutputState(person, UpdatableNameContract.class)
    .addCommand(UpdatableNameContract.UpdateName, key);
```

So, once you've bound your state to a particular contract type, that binding must be applied for all subsequent transactions using that state until consumption, but the important thing is, you can choose different bindings at creation, which lends itself to code reuse.

I'd be inclined to remove the annotation, favouring this...
```kotlin
fun addOutputState(state: ContractState, type: Class<out Contract>): UtxoTransactionBuilder
```

That do you think?

ledger/ledger-utxo/src/main/kotlin/net/corda/v5/ledger/utxo/transaction/UtxoSignedTransaction.kt
MrMatthewLayton: @vlajos The reason I chose `Set<T>` here was that there is no reason for having duplicate signatures in a signed transaction, which `List<T>` would allow, but `Set<T>` would not (assuming that the value semantics of `DigitalSignatureAndMetadata` are implemented correctly).

ledger/ledger-utxo/src/main/kotlin/net/corda/v5/ledger/utxo/transaction/UtxoLedgerService.kt
mnesbit: That is not normal builder semantics though, which generally allow any order of operations and validate at build step. Also, in C5 TimeWindow is now required, but is most reasonably the last thing added, so not sure that stands up to scrutiny.

ledger/ledger-utxo/src/main/kotlin/net/corda/v5/ledger/utxo/transaction/UtxoLedgerService.kt
mnesbit: I think a singular one would be useful

ledger/ledger-utxo/src/main/kotlin/net/corda/v5/ledger/utxo/transaction/UtxoLedgerService.kt
MrMatthewLayton: @mnesbit Lots of builder patterns allow you to pass initial data via a constructor (which is essentially what this is doing). `TransactionBuilder` in Corda 4 allowed you to pass the notary via the constructor. `StringBuilder` allows you to pass in an initial string `val x = StringBuilder("Hello, World!")`.

The only way I can think of having a `setNotary` function that aligns with `setTimeWindow` would be to default it at creation, which would then need service access to get the default notary (and I'm not even sure that's useful to the developer). For time window, we could have a `DefaultTimeWindowImpl` where `from` is `Instant.MIN` and `until` is `Instant.MAX`, which would give the equivalent behaviour to Corda 4 with a `null` time window.

ledger/ledger-utxo/src/test/java/net/corda/v5/ledger/utxo/AbstractMockTestHarness.java
vlajos: Hmm. I missed only the point here...
Correctly:

AFAIK * imports are *NOT* accepted.

ledger/ledger-utxo/src/main/kotlin/net/corda/v5/ledger/utxo/transaction/UtxoSignedTransaction.kt
vlajos: It sounds convincing.
I think the main reason for `List<>` in consensual was only to follow the C4 pattern.
If @mnesbit or @blsemo are happy with `Set<>`, I will change the consensual signed Tx as well.

.editorconfig
```diff
@@ -0,0 +1,6 @@
...
+
+[*]
+end_of_line = lf
+insert_final_newline = true
+max_line_length = 140
```
mnesbit: Just a reminder that you need to add this file to .gitignore!

ledger/ledger-utxo/src/main/kotlin/net/corda/v5/ledger/utxo/transaction/UtxoLedgerService.kt
mnesbit: Lets just go with setNotary on builder. I think it is just nicer and my quick poll of Rick's view was similar

ledger/ledger-utxo/src/main/kotlin/net/corda/v5/ledger/utxo/transaction/UtxoSignedTransaction.kt
mnesbit: You can't really have a Set of signatures as they will likely all be unique. We can have a Map<PublicKey, DigitalSignatureAndKey>, but bluntly not of much value. Use List as main API checking of 'sufficient' signatures is done by us during verify/recordTransaction

mnesbit: Remember that for RSA and ECDSA signatures repeat signing will produce different values as the algorithm includes entropy. (This isn't true for EdDSA that uses a cryptographic trick to remove need for the entropy and be deterministic, which is lovely but we can't generalise)

ledger/ledger-utxo/src/main/kotlin/net/corda/v5/ledger/utxo/transaction/UtxoTransactionBuilder.kt
mnesbit: Needs a short design meeting I think.

mnesbit: @rick-r3 and I drilled into C4 and we have concluded that the versions which pass in strings are possibly unsafe allowing people to hijack states with other contracts. The annotation based linkage between states and the contract class is possible to make safe. I would like to hear if there are any reasons people use the string approach that I haven't heard of, but we should probably kill all the forms taking strings.

ledger/ledger-utxo/src/main/kotlin/net/corda/v5/ledger/utxo/transaction/UtxoTransactionBuilder.kt
```diff
@@ -0,0 +1,228 @@
...
+
+    /**
+     * Adds an output state to the current [UtxoTransactionBuilder].
+     *
+     * @param contractState The [ContractState] to add to the current [UtxoTransactionBuilder].
+     * @param contractId The class name of the [Contract] associated with the transaction state.
+     * @param notary The notary that will be used to notarise the specified [ContractState].
+     * @return Returns a [UtxoTransactionBuilder] including the additional output state.
+     */
+    fun addOutputState(contractState: ContractState, contractId: String, notary: Party): UtxoTransactionBuilder
```
blsemo: Why do we need a function that adds a notary along with the output? I thought the notary has to be the same for all inputs and outputs?

ledger/ledger-utxo/src/main/kotlin/net/corda/v5/ledger/utxo/transaction/UtxoTransactionBuilder.kt
```diff
@@ -0,0 +1,228 @@
...
+     */
+    fun addOutputState(contractState: ContractState, contractId: String, notary: Party, encumbrance: Int?): UtxoTransactionBuilder
+
+    /**
+     * Sets the transaction time window to be valid from the specified [Instant], tending towards positive infinity.
+     *
+     * @param from The [Instant] from which the transaction time window is valid.
+     * @return Returns a [UtxoTransactionBuilder] including the specified time window.
+     */
+    fun setTimeWindowFrom(from: Instant): UtxoTransactionBuilder
```
blsemo, line 124: I don't think the notary would be happy with that. 
In order to be able to archive (consumed) states, we always need a finite end of the time window for every tx. That can be months in the future, but it needs to be defined.

ledger/ledger-utxo/src/main/kotlin/net/corda/v5/ledger/utxo/transaction/UtxoTransactionBuilder.kt
```diff
@@ -0,0 +1,228 @@
...
+     * @return Returns a [UtxoSignedTransaction] with signatures for the specified signatory keys.
+     */
+    fun verifyAndSign(vararg signatories: PublicKey): UtxoSignedTransaction
+
+    /**
+     * Builds the current transaction into a [UtxoWireTransaction].
+     *
+     * @return Returns a [UtxoWireTransaction] from the current [UtxoTransactionBuilder].
+     */
+    fun toWireTransaction(): UtxoWireTransaction
```
blsemo: Not sure that needs to be on the public API - users should only have to deal in signed and ledger tx.

ledger/ledger-utxo/src/main/kotlin/net/corda/v5/ledger/utxo/transaction/UtxoTransactionBuilder.kt
```diff
@@ -0,0 +1,228 @@
...
+     * Verifies the current transaction.
+     */
+    fun verify()
+
+    /**
+     * Verifies and signs the transaction with any required signatories that belong to the current node.
+     *
+     * @return Returns a [UtxoSignedTransaction] with signatures for any required signatories that belong to the current node.
+     */
+    fun verifyAndSign(): UtxoSignedTransaction
```
blsemo, line 185: Do we logically need a verify and sign method or is that just a convenience shortcut?

ledger/ledger-utxo/src/main/kotlin/net/corda/v5/ledger/utxo/transaction/UtxoTransactionBuilder.kt
```diff
@@ -0,0 +1,228 @@
...
+    fun setTimeWindowBetween(from: Instant, until: Instant): UtxoTransactionBuilder
+
+    /**
+     * Sets the transaction time window to be valid at the specified [Instant] with the specified tolerance.
+     *
+     * @param midpoint The [Instant] representing the midpoint of the time window.
+     * @param tolerance The [Duration] representing the tolerance of the time window, which is applied before and after the midpoint.
+     * @return Returns a [UtxoTransactionBuilder] including the specified time window.
+     */
+    fun setTimeWindowBetween(midpoint: Instant, tolerance: Duration): UtxoTransactionBuilder
```
blsemo, line 150: Not sure we need to support this function?

ledger/ledger-utxo/src/main/kotlin/net/corda/v5/ledger/utxo/Attachment.kt
```diff
@@ -0,0 +1,65 @@
...
+    val signatories: Set<PublicKey>
+
+    /**
+     * Finds the specified file (case insensitively) within the attachment and copies it to the specified output stream.
+     *
+     * @param path The path of the file to find within the attachment.
+     * @param output The output stream where the file should be copied to.
+     * @throws [FileNotFoundException] if the specified path does not exist in the attachment.
+     */
+    fun extractFile(path: String, output: OutputStream)
```
yarivus, line 43: I think the interface should not limit the implementation to files.
It will help with future enhancement and mocking of tests ;)

ledger/ledger-utxo/src/main/kotlin/net/corda/v5/ledger/utxo/StateAndRef.kt
```diff
@@ -0,0 +1,17 @@
...
+
+/**
+ * Represents a composition of a [TransactionState] and a [StateRef].
+ *
+ * @property state The [TransactionState] component of the current [StateAndRef].
+ * @property ref The [StateRef] component of the current [StateAndRef].
+ */
+@DoNotImplement
+@CordaSerializable
+interface StateAndRef<out T : ContractState> {
```
yarivus, line 14: We should define a better name and get that issue closed and decided.

ledger/ledger-utxo/src/main/kotlin/net/corda/v5/ledger/utxo/TimeWindow.kt
```diff
@@ -0,0 +1,35 @@
...
+ * and may be fully bounded, or unbounded towards negative or positive infinity.
+ *
+ * @property from The boundary at which the time window begins, or null if the time window is unbounded towards negative infinity.
+ * @property until The boundary at which the time window ends, or null if the time window is unbounded towards positive infinity.
+ * @property midpoint The midpoint between a fully bounded time window, or null if the time window is unbounded towards positive or negative infinity.
+ * @property duration The duration of a fully bounded time window, or null if the time window is unbounded towards positive or negative infinity.
+ */
+@DoNotImplement
+@CordaSerializable
+interface TimeWindow {
```
yarivus, line 22: I think this is redundant ...
We can use builtins such as Duration to represent most of our TimeWindow s (for midpoint it's very easy to calc in place).

yarivus (commented):
@MrMatthewLayton - please see my comments

ledger/ledger-utxo/src/main/kotlin/net/corda/v5/ledger/utxo/Attachment.kt
```diff
@@ -0,0 +1,65 @@
...
+     * @return Returns the [InputStream] for the current [Attachment].
+     */
+    fun open(): InputStream
+
+    /**
+     * Opens the current [Attachment] as a JAR (Java Archive).
+     *
+     * @return Returns the [JarInputStream] for the current [Attachment].
+     */
+    fun openAsJar(): JarInputStream {
```
blsemo, line 57: I don't think we want to do this - no executable JAR attachments in Corda 5. Contracts are distributed as CPKs/CPIs, not JARs.

ledger/ledger-utxo/src/main/kotlin/net/corda/v5/ledger/utxo/Attachment.kt
mnesbit: Attachments are zips/jars so they are necessarily file based, though I agree this should be URL rather than path

ledger/ledger-utxo/src/main/kotlin/net/corda/v5/ledger/utxo/TimeWindow.kt
mnesbit: @yarivus there are various reasons why we have this interface and also some of these foundation types are familiar to Corda 4 users and thus we are preserving naming to make transition easier. Please stop redesigning without fuller discussions.

ledger/ledger-utxo/src/main/kotlin/net/corda/v5/ledger/utxo/TimeWindow.kt
```diff
@@ -0,0 +1,35 @@
...
+    val from: Instant?
+    val until: Instant?
```
mnesbit, line 24: This needs to be non-nullable now for Notary changes

ledger/ledger-utxo/src/main/kotlin/net/corda/v5/ledger/utxo/TimeWindow.kt
```diff
@@ -0,0 +1,35 @@
...
+    val midpoint: Instant?
```
mnesbit, line 25: midpoint and duration are not part of the fundamental API so it should be an extension method

ledger/ledger-utxo/src/main/kotlin/net/corda/v5/ledger/utxo/StateAndRef.kt
mnesbit: This is a familiar type for C4 developers so we should leave alone, though needs better doc comments

ledger/ledger-utxo/src/main/kotlin/net/corda/v5/ledger/utxo/transaction/UtxoLedgerTransaction.kt
```diff
@@ -0,0 +1,189 @@
...
+    fun <T : Command> getCommands(type: Class<T>): List<T>
+
+    /**
+     * Obtains all ledger transaction [StateAndRef] inputs that match the specified [ContractState] type.
+     *
+     * @param T The underlying type of the [ContractState].
+     * @param type The type of the [ContractState].
+     * @return Returns all ledger transaction [StateAndRef] inputs that match the specified [ContractState] type.
+     */
+    fun <T : ContractState> getInputStateAndRefs(type: Class<T>): List<StateAndRef<T>>
```
blsemo, line 83: Do we need all these methods that resolve e.g. a StateRef to a StateAndRef on the LedgerTx?

ledger/ledger-utxo/src/main/kotlin/net/corda/v5/ledger/utxo/transaction/UtxoTransactionBuilder.kt
```diff
@@ -0,0 +1,202 @@
...
+interface UtxoTransactionBuilder {
+
+    val notary: Party
+    val timeWindow: TimeWindow
+    val attachments: List<SecureHash>
+    val commands: List<Command>
+    val signatories: Set<PublicKey>
+    val inputStateAndRefs: List<StateAndRef<*>>
+    val referenceInputStateAndRefs: List<StateAndRef<*>>
+    val outputTransactionStates: List<TransactionState<*>>
```
blsemo, line 39: Do these need to be public properties? I.e. is there a use case for getting e.g. the notary back from TX builder?

ledger/ledger-utxo/build.gradle
```diff
@@ -14,6 +14,8 @@ dependencies {
     api project(':application')
     api project(':base')
     api project(':crypto')
+    api project(':cipher-suite')
```
mnesbit, line 17: I am not sure if we should be referencing cipher-suite at all from any flow api modules. @alexey-kadyrov-r3 needs to clarify a bit.

lankydan: I agree, suggests that there is a type that should exist in `crypto` instead?

ledger/ledger-utxo/src/main/kotlin/net/corda/v5/ledger/utxo/StateRef.kt
```diff
@@ -0,0 +1,18 @@
...
+
+/**
+ * Defines a reference to a [ContractState].
+ *
+ * @property index The index of the state in the transaction's outputs in which the referenced state was created.
+ * @property transactionHash The hash of the transaction in which the referenced state was created.
+ */
+@DoNotImplement
+@CordaSerializable
+interface StateRef {
```
mnesbit, line 15: This may need to be a POJO, because it is used frequently by users. We also need to/from string functions as that is very common. Certainly we need a clear factory if we don't make this a POJO.

Can we also re-order to txhash then index and I suggest we don't rename the fields vs C4 as it is a heavily used class in existing code. The C4 version is
`@CordaSerializable
data class StateRef(val txhash: SecureHash, val index: Int) {
    override fun toString() = "$txhash($index)"
}`

ledger/ledger-utxo/src/main/kotlin/net/corda/v5/ledger/utxo/transaction/UtxoTransactionBuilder.kt
MrMatthewLayton: Convenient shortcut.

ledger/ledger-utxo/src/main/kotlin/net/corda/v5/ledger/utxo/transaction/UtxoTransactionBuilder.kt
MrMatthewLayton: Notary based functions have been removed.

ledger/ledger-utxo/src/main/kotlin/net/corda/v5/ledger/utxo/transaction/UtxoTransactionBuilder.kt
MrMatthewLayton: @mnesbit TimeWindow itself on UtxoTransactionBuilder is non-nullable now, but it's components (From and Until) are nullable. What are your thoughts?

ledger/ledger-utxo/src/main/kotlin/net/corda/v5/ledger/utxo/TimeWindow.kt
MrMatthewLayton: @mnesbit Ignore comments above - answered here.

ledger/ledger-utxo/src/main/kotlin/net/corda/v5/ledger/utxo/transaction/UtxoLedgerTransaction.kt
MrMatthewLayton: AFAIK no such methods on the original `LedgerTransaction` and no equivalent here. We only have the bare minimum APIs we need.

ledger/ledger-utxo/src/main/kotlin/net/corda/v5/ledger/utxo/transaction/UtxoTransactionBuilder.kt
MrMatthewLayton: I suppose not, although this is something to think about. There are cases where having public access is useful, but perhaps not for everything.

ledger/ledger-utxo/src/main/kotlin/net/corda/v5/ledger/utxo/Attachment.kt
mnesbit: I agree we should probably open as zip

ledger/ledger-utxo/build.gradle
MrMatthewLayton: This might be left over from a dependency on the `DigestService`, and can probably be removed entirely.

ledger/ledger-utxo/src/main/kotlin/net/corda/v5/ledger/utxo/StateRef.kt
MrMatthewLayton: I feel like we need to nail down what can be a POJO and what can't, as I've heard different opinions. `StateRef`, `StateAndRef`, `TransactionState`, `TimeWindow`, etc were all POJOs in a previous PR, but were changed to interfaces with implementations in Corda Runtime as per a conversation I had with Christian (which went something along the lines of "If user code isn't expected to instantiate these things, then they can be interfaces, and let the runtime deal with it).

Yes we can reorder/rename, although I don't think the order matters so much on the interface as it does on the implementation. The reason it was changed from `txhash` to `transactionHash` was basically because I'm a stickler for naming, and "don't abbreviate" is high on my list of don't dos.

yarivus: I agree on the non abbreviated naming.
With modern IDEs that has become a non-problem.
Also, if you have a proper and non-abbreviated names - you can sometimes save yourself the comments (if the name is self-explanatory).

yarivus (commented):
LGTM - but there are remaining comments I believe

ledger/ledger-utxo/src/main/kotlin/net/corda/v5/ledger/utxo/transaction/UtxoTransactionBuilder.kt
MrMatthewLayton: The vararg overloads are useful when you only have to provide a single key. Calling a method with setOf(key) is annoying.

ledger/ledger-utxo/src/main/kotlin/net/corda/v5/ledger/utxo/StateRef.kt
blsemo: I stand by that anything users don't need to instantiate themselves should be an interface - `StateRef` might not be one of those things, though.

PR #617, opened 2022-09-22
CORE-4166 flow messaging apis

LWogan:
https://github.com/corda/corda-runtime-os/pull/2209


gradle.properties
```diff
@@ -7,11 +7,11 @@ cordaProductVersion = 5.0.0
 # NOTE: update this each time this module contains a breaking change
 ## NOTE: currently this is a top level revision, so all API versions will line up, but this could be moved to
 ##   a per module property in which case module versions can change independently.
-cordaApiRevision = 413
+cordaApiRevision = 414
 
 # Main
 kotlinVersion = 1.7.10
-kotlin.stdlib.default.dependency = false
+kotlin.stdlib.default.dependency = falsea
```
JamesHR3: I think you've added a character here accidentally.

application/src/main/kotlin/net/corda/v5/application/messaging/FlowMessaging.kt
```diff
@@ -118,4 +118,56 @@ interface FlowMessaging {
      */
     @Suspendable
     fun initiateFlow(x500Name: MemberX500Name, flowContextPropertiesBuilder: FlowContextPropertiesBuilder): FlowSession
+
+    /** Suspends until a message has been received for each session in the specified [sessions].
```
JamesHR3: I've just noticed this first sentence is on the wrong line as well.

application/src/main/kotlin/net/corda/v5/application/messaging/FlowMessaging.kt
```diff
@@ -118,4 +118,56 @@ interface FlowMessaging {
...
+     * Consider [receiveAll(receiveType: Class<R>, sessions: Set<FlowSession>): List<UntrustworthyData<R>>] when the same type is expected from all sessions.
+     *
+     * Remember that when receiving data from other parties the data should not be trusted until it's been thoroughly
+     * verified for consistency and that all expectations are satisfied, as a malicious peer may send you subtly
+     * corrupted data in order to exploit your code.
+     *
+     * @returns a [Map] containing the objects received, wrapped in an [UntrustworthyData], by the [FlowSession]s who sent them.
+     */
+    @Suspendable
+    fun receiveAllMap(sessions: Map<FlowSession, Class<out Any>>): Map<FlowSession, Any>
```
JamesHR3: We should probably be consistent with declaration ordering in this file, so I'd prefer to see `receiveAll` before `receiveAllMap` here (to line up with the send case).

application/src/main/kotlin/net/corda/v5/application/messaging/FlowMessaging.kt
```diff
@@ -118,4 +118,56 @@ interface FlowMessaging {
...
     @Suspendable
     fun initiateFlow(x500Name: MemberX500Name, flowContextPropertiesBuilder: FlowContextPropertiesBuilder): FlowSession
+
+    /** Suspends until a message has been received for each session in the specified [sessions].
+     *
+     * Consider [receiveAll(receiveType: Class<R>, sessions: Set<FlowSession>): List<UntrustworthyData<R>>] when the same type is expected from all sessions.
+     *
+     * Remember that when receiving data from other parties the data should not be trusted until it's been thoroughly
+     * verified for consistency and that all expectations are satisfied, as a malicious peer may send you subtly
+     * corrupted data in order to exploit your code.
```
JamesHR3: We no longer have `UntrustworthyData` so you'll need to edit these comments to remove references to this. Further down the type is referenced directly.

Following our API description work in DP2, it's probably worth looking at what the comments look like for `send` and `receive` on the `FlowSession` and editing these to give the same level of detail about what the API does (otherwise we'll have to come back and do it later anyway).

JamesHR3 (commented):
A few bits and pieces that I may have previously missed.

application/src/main/kotlin/net/corda/v5/application/messaging/FlowMessaging.kt
LWogan: send and receive are pretty sparse

application/src/main/kotlin/net/corda/v5/application/messaging/FlowMessaging.kt
```diff
@@ -118,4 +119,69 @@ interface FlowMessaging {
...
+     * Consider [receiveAllMap(sessions: Map<FlowSession, Class<out Any>>): Map<FlowSession, Any>] when sessions are
+     * expected to receive different types.
+     *
+     * Remember that when receiving data from other parties the data should not be trusted until it's been thoroughly
+     * verified for consistency and that all expectations are satisfied, as a malicious peer may send you subtly
+     * corrupted data in order to exploit your code.
+     *
+     * @param receiveType type of object to be received for all [sessions].
+     * @param sessions Set of sessions to receive from.
+     * @returns a [List] containing the objects received with the same order of [sessions].
```
JamesHR3: Sorry, minor nitpick. Is the list ordering defined? The problem is that we ask for a `Set` on input, so I don't think we know anything about the session input ordering?

application/src/main/kotlin/net/corda/v5/application/messaging/FlowMessaging.kt
JamesHR3: That's ok, just looking for parity. I think the highlighted paragraph about not trusting received data from other parties needs to go though, as we've removed the mechanism for doing this.

application/src/main/kotlin/net/corda/v5/application/messaging/FlowMessaging.kt
```diff
@@ -118,4 +119,69 @@ interface FlowMessaging {
...
+     *
+     * @throws [CordaRuntimeException] if the session is closed or in a failed state.
```
JamesHR3: Think this should say if any session is closed or in a failed state (rather than one session).

application/src/main/kotlin/net/corda/v5/application/messaging/FlowMessaging.kt
```diff
@@ -118,4 +119,69 @@ interface FlowMessaging {
...
+     */
+    @Suspendable
+    fun receiveAllMap(sessions: Map<FlowSession, Class<out Any>>): Map<FlowSession, Any>
+
+    /**
+     * Queues the given [payload] for sending to the provided [sessions] and continues without waiting for a response.
+     *
+     * Note that the other parties may receive the message at some arbitrary later point or not at all: if one of the provided [sessions]
+     * is offline then message delivery will be retried until the corresponding node comes back or until the message is older than the
+     * network's event horizon time.
```
JamesHR3: Sorry, more documentation issues - the network event horizon time is an old concept that doesn't apply to the new session layer. It's now going to depend on session acknowledgements and the heartbeating mechanism. We'll need to rewrite this paragraph (and for the `sendAllMap`) accordingly.

JamesHR3 (approved):

PR #618, opened 2022-09-23
CORE-6364 Cleaning up merkle tree API

alexey-kadyrov-r3:
* Separated MerkleTreeHashDigestFactory to two interface - one marker for use in application (MerkleTreeHashDigestFactory) and another the actual (MerkleTreeHashDigestProvider) that needs to be implemented as extension and moved the latter to crypto-extensions
* Moved MerkleTreeFactory to application and created MerkleTreeProvider in the cipher-suite, (in runtime-os MerkleTreeFactoryImpl should proxy for the MerkleTreeProviderImlp)
* Moved constants to HashDigestConstants

### DIFF
--- a/application/src/main/kotlin/net/corda/v5/application/crypto/MerkleTreeFactory.kt
+++ b/application/src/main/kotlin/net/corda/v5/application/crypto/MerkleTreeFactory.kt
@@ -4,10 +4,10 @@ import net.corda.v5.crypto.DigestAlgorithmName
 import net.corda.v5.base.annotations.DoNotImplement
 import net.corda.v5.base.annotations.Suspendable
 import net.corda.v5.crypto.merkle.MerkleTree
-import net.corda.v5.crypto.merkle.MerkleTreeHashDigestFactory
+import net.corda.v5.crypto.merkle.MerkleTreeHashDigest
 
 /**
- * [MerkleTreeFactory] creates [MerkleTree]s and [MerkleTreeHashDigestFactory]s.
+ * [MerkleTreeFactory] creates [MerkleTree]s and [MerkleTreeHashDigest]s.
  */
 @DoNotImplement
 interface MerkleTreeFactory {
@@ -22,22 +22,22 @@ interface MerkleTreeFactory {
     @Suspendable
     fun createTree(
         leaves: List<ByteArray>,
-        digestProvider: MerkleTreeHashDigestFactory
+        digestProvider: MerkleTreeHashDigest
     ) : MerkleTree
 
     /**
-     * Creates a [MerkleTreeHashDigestFactory].
+     * Creates a [MerkleTreeHashDigest].
      *
      * @param merkleTreeHashDigestProviderName name of the Hash Digest Provider class
      * @param digestAlgorithmName name of the base Hash algorithm
      * @param options Hash Digest provider specific options
      *
-     * @return A new [MerkleTreeHashDigestFactory] instance.
+     * @return A new [MerkleTreeHashDigest] instance.
      */
     @Suspendable
     fun createHashDigestProvider(
         merkleTreeHashDigestProviderName: String,
         digestAlgorithmName: DigestAlgorithmName,
         options: Map<String, Any> = emptyMap(),
-    ) : MerkleTreeHashDigestFactory
+    ) : MerkleTreeHashDigest
 }
--- a/application/src/main/kotlin/net/corda/v5/application/crypto/MerkleTreeFactory.kt
+++ b/application/src/main/kotlin/net/corda/v5/application/crypto/MerkleTreeFactory.kt
@@ -1,0 +1,43 @@
+package net.corda.v5.application.crypto
+
+import net.corda.v5.crypto.DigestAlgorithmName
+import net.corda.v5.base.annotations.DoNotImplement
+import net.corda.v5.base.annotations.Suspendable
+import net.corda.v5.crypto.merkle.MerkleTree
+import net.corda.v5.crypto.merkle.MerkleTreeHashDigestFactory
+
+/**
+ * [MerkleTreeFactory] creates [MerkleTree]s and [MerkleTreeHashDigestFactory]s.
+ */
+@DoNotImplement
+interface MerkleTreeFactory {
+    /**
+     * Creates a [MerkleTree]
+     *
+     * @param leaves The leaves of the tree.
+     * @param digestProvider Merkle Tree Hash digest provider used to construct the tree's node and leaf hashes.
+     *
+     * @return A new [MerkleTree] instance.
+     */
+    @Suspendable
+    fun createTree(
+        leaves: List<ByteArray>,
+        digestProvider: MerkleTreeHashDigestFactory
+    ) : MerkleTree
+
+    /**
+     * Creates a [MerkleTreeHashDigestFactory].
+     *
+     * @param merkleTreeHashDigestProviderName name of the Hash Digest Provider class
+     * @param digestAlgorithmName name of the base Hash algorithm
+     * @param options Hash Digest provider specific options
+     *
+     * @return A new [MerkleTreeHashDigestFactory] instance.
+     */
+    @Suspendable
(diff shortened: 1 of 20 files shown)


crypto/src/main/kotlin/net/corda/v5/crypto/merkle/MerkleTreeHashDigestFactory.kt
```diff
@@ -0,0 +1,14 @@
...
+import net.corda.v5.crypto.DigestAlgorithmName
+
+/**
+ * Different use cases require different [MerkleTree] calculations.
+ * [MerkleTreeHashDigestFactory]s let us specify such implementations.
+ *
+ *  @property digestAlgorithmName Specifies the digest algorithm.
+ */
+
+interface MerkleTreeHashDigestFactory {
```
mnesbit: Can we rename to MerkleTreeHashDigest

mnesbit (dismissed):
OK

omfR3:
> * Separated MerkleTreeHashDigestFactory to two interface - one marker for use in application (MerkleTreeHashDigestFactory) and another the actual (MerkleTreeHashDigestProvider) that needs to be implemented as extension and moved the latter to crypto-extensions

Is the purpose of that extra decoupling to reduce the risk of Cordapp code breakage when something in the "real" interface changes?

alexey-kadyrov-r3:
> Is the purpose of that extra decoupling to reduce the risk of Cordapp code breakage when something in the "real" interface changes?

Yes, as the CorDapp developer doesn't care about the actual implementation and should never call those methods directly so the decoupling lets achieve it

crypto/src/main/kotlin/net/corda/v5/crypto/merkle/MerkleTreeHashDigestFactory.kt
omfR3: Why though? I'm a bit confused about the purpose of the interface now.

alexey-kadyrov-r3: The implementation of that interface would be a proxy to the MerkleTreeHashDigestProvider. That's again separation of concern and controlling what is actually available to the CorDApp developer, the contract of that interface suites the app developer as the actual implementation may have stuff which is of no interest to the app developer. If you look the methods are actually different

crypto/src/main/kotlin/net/corda/v5/crypto/merkle/MerkleProof.kt
```diff
@@ -31,6 +31,6 @@ interface MerkleProof {
      */
     fun verify(
         root: SecureHash,
-        digestProvider: MerkleTreeHashDigestProvider
+        digestProvider: MerkleTreeHashDigest
```
omfR3: @alexey-kadyrov-r3 in corda-runtime-os, this is a bit of an issue because it needs to call methods which don't exist in the marker interface. Is this where the proxying you mentioned needs to happen? The analogy for this was `DigitalSignatureVerificationServiceImpl`, but there we inject the proxy service, which doesn't seem possible here.

omfR3: Update: just had a quick chat with @mnesbit who explained that it's ok to just cast in this situation, since we can expect that it's safe to do so.

cipher-suite/src/main/kotlin/net/corda/v5/cipher/suite/merkle/MerkleTreeProvider.kt
```diff
@@ -0,0 +1,44 @@
...
+     * @param leaves The leaves of the tree.
+     * @param digestProvider Merkle Tree Hash digest provider used to construct the tree's node and leaf hashes.
+     *
+     * @return A new [MerkleTree] instance.
+     */
+    @Suspendable
+    fun createTree(
+        leaves: List<ByteArray>,
+        digestProvider: MerkleTreeHashDigestProvider
+    ) : MerkleTree
```
omfR3, line 27: What's the benefit of having `MerkleTreeProvider` as well as `MerkleTreeFactory`? We can create `MerkleTreeProviderImpl` and proxy call into it from `MerkleTreeFactoryImpl`, but it will still return `MerkleTree` which relies on the new marker interface `MerkleTreeHashDigest` rather than `MerkleTreeHashDigestProvider`.

alexey-kadyrov-r3: application implantations act as a mediator between the CorDapp code and the platform, that way the contracts may diverge and the application mediator may do some extra stuff, and yes in some cases it looks just like a code duplication

omfR3: Alright, so to stick with that pattern, we should create `MerkleTreeProviderImpl` in corda-runtime-os which will basically do what `MerkleTreeFactoryImpl` did? And have `MerkleTreeFactoryImpl` just delegate to the provider impl?

alexey-kadyrov-r3: Yep

application/src/main/kotlin/net/corda/v5/application/crypto/MerkleTreeFactory.kt
```diff
@@ -0,0 +1,43 @@
...
+     * Creates a [MerkleTreeHashDigest].
+     *
+     * @param merkleTreeHashDigestProviderName name of the Hash Digest Provider class
+     * @param digestAlgorithmName name of the base Hash algorithm
+     * @param options Hash Digest provider specific options
+     *
+     * @return A new [MerkleTreeHashDigest] instance.
+     */
+    @Suspendable
+    fun createHashDigestProvider(
```
alexey-kadyrov-r3: @omfR3 can you rename it to `createHashDigest`?

crypto/src/main/kotlin/net/corda/v5/crypto/merkle/MerkleProof.kt
alexey-kadyrov-r3: @omfR3 Can you rename it to `digest`, and generally go through the contracts and check that the names are matching after the renaming?

crypto/src/main/kotlin/net/corda/v5/crypto/merkle/MerkleTree.kt
```diff
@@ -9,13 +9,13 @@ import net.corda.v5.crypto.SecureHash
...
  * @property leaves The input data elements. Usually something deterministically serialized.
- * @property digestProvider The [MerkleTreeHashDigestProvider] used to construct the tree's node and leaf hashes.
+ * @property digestProvider The [MerkleTreeHashDigest] used to construct the tree's node and leaf hashes.
  * @property root The root element of the tree which is essentially the fingerprint of the whole tree/data set.
  *
  */
 interface MerkleTree {
     val leaves: List<ByteArray>
-    val digestProvider: MerkleTreeHashDigestProvider
+    val digestProvider: MerkleTreeHashDigest
```
alexey-kadyrov-r3: @omfR3 Can you rename it to digest?

application/src/main/kotlin/net/corda/v5/application/crypto/MerkleTreeFactory.kt
omfR3: Yes. I still don't quite understand whether that interface is supposed to represent a hash digest or a hash digest provider, but I'll make that change.

omfR3:
(See corresponding [corda-runtime-os PR](https://github.com/corda/corda-runtime-os/pull/2214))

mcgovc (approved):
LGTM - build changes approved

blsemo (approved):
Looks fine to me

PR #621, opened 2022-09-23
CORE-6871 Consensual Finality flow first version

lankydan:
Added `ConsensualFinalityFlow` and `ConsensualReceiveFinalityFlow` which are both accessed by `ConsensualLedgerService`.

### DIFF
--- a/ledger/ledger-consensual/src/main/kotlin/net/corda/v5/ledger/consensual/ConsensualLedgerService.kt
+++ b/ledger/ledger-consensual/src/main/kotlin/net/corda/v5/ledger/consensual/ConsensualLedgerService.kt
@@ -1,7 +1,10 @@
 package net.corda.v5.ledger.consensual
 
+import net.corda.v5.application.messaging.FlowSession
 import net.corda.v5.base.annotations.DoNotImplement
 import net.corda.v5.base.annotations.Suspendable
+import net.corda.v5.ledger.consensual.transaction.ConsensualSignedTransaction
+import net.corda.v5.ledger.consensual.transaction.ConsensualSignedTransactionVerifier
 import net.corda.v5.ledger.consensual.transaction.ConsensualTransactionBuilder
 
 /**
@@ -21,4 +24,33 @@ interface ConsensualLedgerService {
     fun receiveFinality( ()-> ... )
     ... Vault ...
     */
+
+    /**
+     * Collects signatures, records and broadcasts to involved peers a [ConsensualSignedTransaction].
+     *
+     * @param signedTransaction The [ConsensualSignedTransaction] to finalise and recorded locally and with peer [sessions].
+     * @param sessions The [FlowSession]s of the peers involved in the transaction.
+     *
+     * @return The fully signed [ConsensualSignedTransaction] that was recorded.
+     */
+    @Suspendable
+    fun finality(
+        signedTransaction: ConsensualSignedTransaction,
+        sessions: List<FlowSession>
+    ): ConsensualSignedTransaction
+
+
+    /**
+     * Verifies, signs and records a [ConsensualSignedTransaction].
+     *
+     * @param session The [FlowSession] to receive the [ConsensualSignedTransaction] from.
+     * @param verifier Verifies the received [ConsensualSignedTransaction].
+     *
+     * @return The fully signed [ConsensualSignedTransaction] that was received and recorded.
+     */
+    @Suspendable
+    fun receiveFinality(
+        session: FlowSession,
+        verifier: ConsensualSignedTransactionVerifier
+    ): ConsensualSignedTransaction
 }
--- a/ledger/ledger-consensual/src/main/kotlin/net/corda/v5/ledger/consensual/transaction/ConsensualSignedTransaction.kt
+++ b/ledger/ledger-consensual/src/main/kotlin/net/corda/v5/ledger/consensual/transaction/ConsensualSignedTransaction.kt
@@ -24,10 +24,6 @@ import net.corda.v5.crypto.SecureHash
  * Thus adding or removing a signature does not change it.
  */
 @DoNotImplement
-// Why do we make it illegal to have a custom serializer for a corda serializable annotated class/interface?
-// This seems very dangerous moving forward if we need to provide custom serializers for existing code
-// net.corda.internal.serialization.amqp.CachingCustomSerializerRegistry - Illegal custom serializer detected for interface net.corda.v5.ledger.consensual.transaction.ConsensualSignedTransaction: net.corda.internal.serialization.amqp.standard.CustomSerializer.Proxy
-//@CordaSerializable
 interface ConsensualSignedTransaction {
     /**
      * @property id The ID of the transaction.
--- a/ledger/ledger-consensual/src/main/kotlin/net/corda/v5/ledger/consensual/transaction/ConsensualSignedTransaction.kt
+++ b/ledger/ledger-consensual/src/main/kotlin/net/corda/v5/ledger/consensual/transaction/ConsensualSignedTransaction.kt
@@ -25,7 +25,10 @@ import java.security.PublicKey
  * Thus adding or removing a signature does not change it.
  */
 @DoNotImplement
-@CordaSerializable
+// Why do we make it illegal to have a custom serializer for a corda serializable annotated class/interface?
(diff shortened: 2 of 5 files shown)


ledger/ledger-consensual/src/main/kotlin/net/corda/v5/ledger/consensual/ConsensualLedgerService.kt
```diff
@@ -21,4 +24,33 @@ interface ConsensualLedgerService {
...
+     *
+     * @param signedTransaction The [ConsensualSignedTransaction] to finalise and recorded locally and with peer [sessions].
+     * @param sessions The [FlowSession]s of the peers involved in the transaction.
+     *
+     * @return The fully signed [ConsensualSignedTransaction] that was recorded.
+     */
+    @Suspendable
+    fun finality(
+        signedTransaction: ConsensualSignedTransaction,
+        sessions: List<FlowSession>
```
mnesbit, line 39: We need to clarify a few things about this when we get to UTXO, because we have confusing sources of participants we need in finality. Ultimately we need 1. Those who need to sign (plus the implicit presence of the Notary) 2. Those who need to receive a copy reliably on completion.  3. Those who can see the states on vault insert. The transaction has a notary (PublicKey) and probably a service MemberX500Name. The transaction also has CommandAndSignatories with their required signers. FInally, ContractState has participants that controls visibility. However, note that the transaction is in terms of PublicKeys and FlowSessions is in terms of MemberX500Names (and not sure if you can get that from structure)

So we probably should have a chat about who and when to use these things

mnesbit: I guess in C4 this works by each session participant evaluating what they feel OK with signing for i.e. identifying their keys from required signers and signing. Challenge there is that lookup of ones own keys by PublicKey/hash of PublicKey is not scalable when there are many accounts/CI keys. We might need to consider how during the negotiation phase peers make a note of what keys they intend to use with the transaction

JamesHR3: Just as an aside, I was idly wondering yesterday about a scheme where we work out what sessions we need internally in the finality flow, rather than ask the user to provide them. The advantage is that we should be able to verify up front if we're able to address the right network members, as we can either get the member X500 name from the key for well known keys, or come up with a secondary system flow for dealing with confidential keys further down the line. However, I think that might cause problems when thinking about inline vs non-inline responder flows, as that sort of idea would require non-inline. Perhaps some mixture of the two could work, where it's a pure verification step rather than actually making the sessions?

ledger/ledger-consensual/src/main/kotlin/net/corda/v5/ledger/consensual/transaction/ConsensualSignedTransactionVerifier.kt
```diff
@@ -0,0 +1,13 @@
+package net.corda.v5.ledger.consensual.transaction
+
+import net.corda.v5.base.annotations.CordaSerializable
+import net.corda.v5.base.annotations.Suspendable
+
+@CordaSerializable
```
blsemo: Why does this need to be `@CordaSerializable`? Will we store that or send it over the wire? I thought the verification in the `receiveFinality` call would be up to the local code?

ledger/ledger-consensual/src/main/kotlin/net/corda/v5/ledger/consensual/transaction/ConsensualSignedTransactionVerifier.kt
```diff
@@ -0,0 +1,13 @@
...
+fun interface ConsensualSignedTransactionVerifier {
+
+    // Should this return a Try.Success/Error? Or control purely through exceptions, but this means that should not
+    // be used to determine behaviour in the peer flow.
```
blsemo: Good question - what is the desired behaviour to tell the initiating party "We're not going to sign this"?

ledger/ledger-consensual/src/main/kotlin/net/corda/v5/ledger/consensual/transaction/ConsensualSignedTransaction.kt
```diff
@@ -25,7 +24,10 @@ import java.security.PublicKey
  * Thus adding or removing a signature does not change it.
  */
 @DoNotImplement
-@CordaSerializable
+// Why do we make it illegal to have a custom serializer for a corda serializable annotated class/interface?
+// This seems very dangerous moving forward if we need to provide custom serializers for existing code
+// net.corda.internal.serialization.amqp.CachingCustomSerializerRegistry - Illegal custom serializer detected for interface net.corda.v5.ledger.consensual.transaction.ConsensualSignedTransaction: net.corda.internal.serialization.amqp.standard.CustomSerializer.Proxy
+//@CordaSerializable
```
blsemo: Is there a follow up ticket/thread for this? Clearly not an essential blocker on this PR, but needs to be sorted out in general.

ledger/ledger-consensual/src/main/kotlin/net/corda/v5/ledger/consensual/ConsensualLedgerService.kt
blsemo: Passing in the sessions is sticking with the Corda 4 approach - creating sessions under the hood is tricky for inline flows as mentioned.
I like the idea of checking that we can resolve the keys to names, and obviously we could then create the matching sessions which would give a nicer API.

In the Consensual ledger world, there is probably not much point to it - I would expect virtually all use cases to get X500 names for the required counterparties as input to the flow, and resolving those to keys for the signing, and sessions for the flow messaging.

I think from a UTOX user perspective, there are a couple of different use cases: 
- you might deal with known counterparties (think a simple cash on ledger model, where the only signatures required are mine and the receiver), and we'll have the X500 names as inputs tying together keys and sessions. In this case, resolving sessions from keys would be a waste of resources.
- But you might also deal with the case where the required signatories are defined by participants in an input state - there we need to resolve key -> X500 anyway, and it might be a lot cleaner to do it under the hood and not bother the user with this.

ledger/ledger-consensual/src/main/kotlin/net/corda/v5/ledger/consensual/transaction/ConsensualSignedTransaction.kt
lankydan: https://r3-cev.atlassian.net/browse/CORE-6971 was created a few days ago.

ledger/ledger-consensual/src/main/kotlin/net/corda/v5/ledger/consensual/transaction/ConsensualSignedTransactionVerifier.kt
lankydan: Can we have a discussion about this.

lankydan:
> Passing in the sessions is sticking with the Corda 4 approach - creating sessions under the hood is tricky for inline flows as mentioned.

This is currently strictly prevented, in that it throws an error if you create a session in a non `@InitiatingFlow`.

There is nothing stopping us from initiating new sessions by calling an `@InitiatingFlow` within `FinalityFlow` which will start new flows with peers. But this won't be connected to `ReceiveFinalityFlow` in any way. 

> But you might also deal with the case where the required signatories are defined by participants in an input state - there we need to resolve key -> X500 anyway, and it might be a lot cleaner to do it under the hood and not bother the user with this.

We can filter through the sessions given to us to link up to the existing flow being used for interaction. For missing parties, we can then start an `@InitiatingFlow` that collects their signatures etc and wait for recording if needed.

lankydan:
merge comment:

```
CORE-6871 Consensual Finality Flow

Initial implementation of `ConsensualFinalityFlow` and 
`ConsensualReceiveFinalityFlow`.

`ConsensualLedgerService` now include `finality` and `receiveFinality` 
to call the internal implementations of the finality flows.

`ConsensualTransactionVerifier` has been added so that it can be passed 
into `receiveFinality`. This verifier verifies the transaction when 
collecting signatures from the parties calling `receiveFinality`.
```

blsemo (approved):

PR #624, opened 2022-09-28
CORE-5552 Added new API to support Corda Services Token Selection.

owenstanford:
CORE-5552 Adds a new API to support Token Selection.


application/src/main/kotlin/net/corda/v5/application/services/TokenCache.kt
```diff
@@ -0,0 +1,100 @@
...
+ *     val criteria = TokenClaimCriteria(
+ *         "Currency",
+ *         getIssuerHash(),
+ *         getNotaryX500Name(),
+ *         "USD",
+ *         BigDecimal(100)
+ *     )
+ *
+ *     // Call the token selection API to try and claim the tokens.
+ *     val claim = tokenCache!!.tryClaim(criteria)
```
owenstanford: remove !!

application/src/main/kotlin/net/corda/v5/application/services/TokenCache.kt
```diff
@@ -0,0 +1,100 @@
...
+ * a target amount of a given state/token type it wishes to spend. If available a set of tokens will be selected that
+ * sum to at least the target amount specified. The tokens will be locked in the cache to prevent other flows from
+ * selecting them.
+ *
+ * The platform will provide an instance of [TokenCache] to flows via property injection.
+ *
+ * Example of use in Kotlin.
+ * ```Kotlin
+ * @CordaInject
+ * var tokenCache: TokenCache? = null
```
owenstanford: ```lateinit var tokenCache: TokenCache```

application/src/main/kotlin/net/corda/v5/application/services/ClaimedToken.kt
```diff
@@ -0,0 +1,29 @@
...
+ * @property stateRef Unique identifier of a state in the <TX_ID>:<STATE_NUM> format.
+ * @property tokenType The type of token.
+ * @property issuerHash The [SecureHash] of the issuer of the token.
+ * @property notaryX500Name The [MemberX500Name] of the notary for the token.
+ * @property symbol The symbol for the token.
+ * @property tag The user defined tag for the token.
+ * @property ownerHash The [SecureHash] for the owner of the token.
+ * @property amount The amount of the token.
+ */
+data class ClaimedToken (
```
lankydan: Are users meant to construct this?

If not potentially consider an interface, if they are, then consider including a builder due to the amount of properties.

application/src/main/kotlin/net/corda/v5/application/services/TokenCache.kt
```diff
@@ -0,0 +1,100 @@
...
+ * ```Java
+ * @CordaInject
+ * public TokenCache tokenCache;
+ *
+ * @Override
+ * public String call(RPCRequestData requestBody) {
+ *
+ *     // Create a criteria describing the tokens to be selected and
+ *     // the target amount to be claimed.
+ *     TokenClaimCriteria criteria = new TokenClaimCriteria
```
lankydan: Formatting is a bit off.

application/src/main/kotlin/net/corda/v5/application/services/TokenClaimCriteria.kt
```diff
@@ -0,0 +1,28 @@
+package net.corda.v5.application.services
```
lankydan: Is `application` the correct place to put this code?

Is it a `ledger` concern instead? Should it be in the `ledger-utxo` module?

application/src/main/kotlin/net/corda/v5/application/services/TokenClaimCriteria.kt
```diff
@@ -0,0 +1,28 @@
...
+ *
+ * @property tokenType The type of tokens to be selected.
+ * @property issuerHash The [SecureHash] of issuer of tokens to be selected.
+ * @property notaryX500Name The [MemberX500Name] of the notary of tokens to be selected.
+ * @property symbol The symbol of the notary of tokens to be selected.
+ * @property targetAmount The minimum value for the sum of [ClaimedToken.amount] for the selected tokens.
+ * @property tagRegex Optional regular expression to match against the [ClaimedToken.tag] field. Null matches all.
+ * @property ownerHash Optional owner [SecureHash] of the tokens to be selected . Null matches all.
+ */
+class TokenClaimCriteria(
```
lankydan: Should we have a builder to make it easier for java users?

Why are some properties not in the constructor? Even if they're optional they should be in the constructor with another constructor provided that sets them.

data/avro-schema/src/main/resources/avro/net/corda/data/services/data/TokenClaimQueryResult.avsc
```diff
@@ -0,0 +1,35 @@
+{
+  "type": "record",
+  "name": "TokenClaimQueryResult",
+  "namespace": "net.corda.data.services.data",
+  "doc": "Represents the result of a claim query, sent back to the flow",
+  "fields": [
+      {
+        "name": "requestContext",
```
lankydan: You shouldn't need the `ExternalEventContext` on your result object, unless you're using the information it contains when receiving the response?

owenstanford: Possibly for the heartbeat logic, but that is coming as later feature. Will remove for now.

application/src/main/kotlin/net/corda/v5/application/services/TokenClaimCriteria.kt
owenstanford: To set them in java/kotlin is pretty clean, would a builder add anything here?
```
        val criteria = TokenClaimCriteria(
            "Currency",
            getIssuerHash(),
            getNotaryX500Name(),
            "USD",
            BigDecimal(100)
        ).apply {
            // Set optional criteria
            ownerHash = getOwnerHash()
            tagRegex = "(test)"
        }
```
```
        TokenClaimCriteria criteria = new TokenClaimCriteria(
                "Currency",
                getIssuerHash(),
                getNotaryX500Name(),
                "USD",
                new BigDecimal(100)
        );

        // Set optional criteria
        criteria.setOwnerHash(getOwnerHash());
        criteria.setTagRegex("(test)");
```

application/src/main/kotlin/net/corda/v5/application/services/TokenClaimCriteria.kt
owenstanford: Interesting, I see token selection (and in the future the other corda services) as platform services for the flows to use, although the token/state data comes from the ledger I'm not sure the service lives there, but I don't have a strong opinion, do we have other flow facing injectable APIs that don't live in application?

lankydan: Yes, for example we're adding new services in the ledger code to be consumed by flows.

application/src/main/kotlin/net/corda/v5/application/services/TokenClaimCriteria.kt
lankydan: There are quite a few properties, and even in the kotlin version, without named arguments, it's not super easy to know what each input represents. I don't mind too much for kotlin since we can always used named args, but java doesn't have the ability.

lankydan: That being said, I'm really not keen on the optional args and would prefer another constructor that sets both and/or either of them.

application/src/main/kotlin/net/corda/v5/application/services/ClaimedToken.kt
mnesbit: Yes we should use interface for and @DoNotImplement objects. I think it is correct that this is not @CordaSerializable as you shouldn't be sending it to anybody.

application/src/main/kotlin/net/corda/v5/application/services/TokenCache.kt
```diff
@@ -0,0 +1,100 @@
...
+ * selecting them.
+ *
+ * The platform will provide an instance of [TokenCache] to flows via property injection.
+ *
+ * Example of use in Kotlin.
+ * ```Kotlin
+ * @CordaInject
+ * var tokenCache: TokenCache? = null
+ *
+ * override fun call(requestBody: RPCRequestData): String {
```
mnesbit: Missing @Suspendable

application/src/main/kotlin/net/corda/v5/application/services/TokenCache.kt
```diff
@@ -0,0 +1,100 @@
...
+
+    /**
+     * Attempts to claim a set of tokens from the cache that satisfies the supplied [TokenClaimCriteria]
+     *
+     * @param criteria The [TokenClaimCriteria] used to select tokens.
+     *
+     * @return [TokenClaim] if enough tokens were claimed to satisfy the [TokenClaimCriteria.targetAmount]. null If the
+     * [TokenClaimCriteria.targetAmount] could not be reached.
+     */
+    fun tryClaim(criteria: TokenClaimCriteria): TokenClaim?
```
mnesbit: Must be @Suspsendable

application/src/main/kotlin/net/corda/v5/application/services/TokenClaim.kt
```diff
@@ -0,0 +1,31 @@
...
+    val claimId: String
+
+    val claimedTokens: List<ClaimedToken>
+
+    /**
+     * Removes any used tokens from the cache and unlocks any remaining tokens for other flows to claim.
+     *
+     * @param usedTokensRefs List of state refs to mark as used in the <TX_ID>:<STATE_NUM> format.
+     */
+    fun useAndRelease(usedTokensRefs: List<String>)
```
mnesbit: must be @Suspendable

application/src/main/kotlin/net/corda/v5/application/services/TokenClaimCriteria.kt
mnesbit: I think it should be in utxo-ledger. The claim logic is really only relevant for states that can be consumed and that only occurs on utxo ledgers with notaries

application/src/main/kotlin/net/corda/v5/application/services/ClaimedToken.kt
```diff
@@ -0,0 +1,29 @@
+package net.corda.v5.application.services
```
mnesbit: I agree with @lankydan below this API is utxo ledger related

application/src/main/kotlin/net/corda/v5/application/services/ClaimedToken.kt
```diff
@@ -0,0 +1,29 @@
...
+ * @property tokenType The type of token.
+ * @property issuerHash The [SecureHash] of the issuer of the token.
+ * @property notaryX500Name The [MemberX500Name] of the notary for the token.
+ * @property symbol The symbol for the token.
+ * @property tag The user defined tag for the token.
+ * @property ownerHash The [SecureHash] for the owner of the token.
+ * @property amount The amount of the token.
+ */
+data class ClaimedToken (
+    val stateRef: String,
```
mnesbit: This should be a StateRef ready to be resolved and pushed into a transaction builder

mnesbit (commented):
We also need to work out where we are going to put the post-insert hook interface. It is similar in concept to the crypto-extensions module, which we had to move to a separate module so that it can be loaded slightly differently if we need. It is also similar to the JPA entities and 'named query' logic that we at some point have. Perhaps we will need a persistence-extensions module?

application/src/main/kotlin/net/corda/v5/application/services/ClaimedToken.kt
owenstanford: yes agreed it should have been an interface will change.

application/src/main/kotlin/net/corda/v5/application/services/TokenClaimCriteria.kt
owenstanford: Moved these and the avro types to be under the ledger.utxo namespace

lankydan (dismissed):

blsemo (dismissed):
The API looks fine to me - I just wonder how much of a token model the avro schema assumes?

ronanbrowne (dismissed):
LGTM (only gralde changes reviewed)

ledger/ledger-utxo/src/main/kotlin/net/corda/v5/ledger/utxo/token/selection/TokenClaimCriteria.kt
```diff
@@ -0,0 +1,28 @@
...
+ * @property tokenType The type of tokens to be selected.
+ * @property issuerHash The [SecureHash] of issuer of tokens to be selected.
+ * @property notaryX500Name The [MemberX500Name] of the notary of tokens to be selected.
+ * @property symbol The symbol of the notary of tokens to be selected.
+ * @property targetAmount The minimum value for the sum of [ClaimedToken.amount] for the selected tokens.
+ * @property tagRegex Optional regular expression to match against the [ClaimedToken.tag] field. Null matches all.
+ * @property ownerHash Optional owner [SecureHash] of the tokens to be selected . Null matches all.
+ */
+class TokenClaimCriteria(
+    val tokenType: String,
```
blsemo, line 19: Isn't that a problem to have the token type as a string here, if the type of tokens in the ledger might e.g. be a reference state? Or the symbol?

blsemo (commented):
Maybe we have to accept that the API dealing with the actual tokens is subject to change until we have figured out the tokens API on the ledger side?

I think the PR also needs another bump in the API version.

lankydan (commented):
Nothing jumps out as wrong to me.

lankydan (approved):

mnesbit (approved):
OK

ronanbrowne (approved):
LGTM (build)

PR #626, opened 2022-10-03
CORE-6874 Custom Json Serializer Interface

simon-johnson-r3:
* The serializer interface which custom serializers (platform or user code) must implement.
* The interface that can write JSON will forward all calls to Jackson in its implementation, so starting with [JsonGenerator (Jackson-core 2.6.0 API)](https://fasterxml.github.io/jackson-core/javadoc/2.6/com/fasterxml/jackson/core/JsonGenerator.html) writeXXXX methods was used to determine which methods are made available to users of this class.
* Tests to ensure the interfaces are Java compliant.
* These interfaces will be part of the Corda public application api, but there will be nothing anyone can do with them at the moment. As such this is not a breaking change.

### DIFF
--- a/application/src/main/kotlin/net/corda/v5/application/marshalling/JsonSerializedBase64Config.kt
+++ b/application/src/main/kotlin/net/corda/v5/application/marshalling/JsonSerializedBase64Config.kt
@@ -5,8 +5,24 @@ package net.corda.v5.application.marshalling
  * enum can be used to configure how that base64 is written into the Json.
  */
 enum class JsonSerializedBase64Config {
+    /**
+     * Standard base64 encoding.
+     */
     MIME,
+
+    /**
+     * Standard base64 encoding but with no line feeds.
+     */
     MIME_NO_LINEFEEDS,
+
+    /**
+     * Base64 encoding suitable for URLs. No line feeds, and characters that need quoting in URLs are replaced with ones
+     * that don't.
+     */
     MODIFIED_FOR_URL,
+
+    /**
+     * Standard base64 encoding but with shorter line lengths.
+     */
     PEM
 }
--- a/application/src/main/kotlin/net/corda/v5/application/marshalling/JsonSerializedBase64Config.kt
+++ b/application/src/main/kotlin/net/corda/v5/application/marshalling/JsonSerializedBase64Config.kt
@@ -1,0 +1,12 @@
+package net.corda.v5.application.marshalling
+
+/**
+ * When objects are serialized to Json using custom serializers and they require to be encoded into base64 encoded, this
+ * enum can be used to configure how that base64 is written into the Json.
+ */
+enum class JsonSerializedBase64Config {
+    MIME,
+    MIME_NO_LINEFEEDS,
+    MODIFIED_FOR_URL,
+    PEM
+}
--- a/application/src/main/kotlin/net/corda/v5/application/marshalling/JsonWriter.kt
+++ b/application/src/main/kotlin/net/corda/v5/application/marshalling/JsonWriter.kt
@@ -1,5 +1,6 @@
 package net.corda.v5.application.marshalling
 
+import net.corda.v5.base.annotations.DoNotImplement
 import java.io.InputStream
 import java.math.BigDecimal
 import java.math.BigInteger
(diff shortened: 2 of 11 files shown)


application/src/main/kotlin/net/corda/v5/application/marshalling/JsonSerializedBase64Config.kt
```diff
@@ -0,0 +1,12 @@
+package net.corda.v5.application.marshalling
+
+/**
+ * When objects are serialized to Json using custom serializers and they require to be encoded into base64 encoded, this
+ * enum can be used to configure how that base64 is written into the Json.
+ */
```
JamesHR3: I wouldn't mind a description of what each element means - even if that's forwarding to the underlying Jackson equivalents in `Base64Variant`.

What's the default? Feels like `MIME` is standard, but I wonder if `MODIFIED_FOR_URL` might be appropriate based on the docs at https://access.redhat.com/webassets/avalon/d/red-hat-jboss-enterprise-application-platform/7.0.0/javadocs/com/fasterxml/jackson/core/class-use/Base64Variant.html?

application/src/main/kotlin/net/corda/v5/application/marshalling/JsonSerializedBase64Config.kt
```diff
@@ -0,0 +1,12 @@
+package net.corda.v5.application.marshalling
```
JamesHR3: I wonder if this is the right place for this? I think at a minimum we should put it under its own package under `marshalling` to indicate that it's specific to JSON custom serializers. Possibly though it should be in its own module, as we might need to declare a dependency on this from the contract/data part of the cordapp, and that shouldn't really need the rest of the `application` API. Thoughts @mnesbit?

application/src/main/kotlin/net/corda/v5/application/marshalling/JsonWriter.kt
```diff
@@ -0,0 +1,482 @@
...
+ * fields. Objects must contain fields and values. Separators between objects, arrays, and array items are automatically
+ * added.
+ *
+ * When fields and values are written with either two method calls or the single methods which take a field name and
+ * value parameter, the separators and enclosing double quotation marks are added automatically. The [JsonWriter] will
+ * throw if you attempt to write values without field names, the exception being the [writeRaw] methods which dump
+ * whatever you pass directly into the Json output unchanged. To write raw content with validation instead call one of
+ * the [writeRawValue] methods, which will assume whatever is passed is a valid Json value.
+ */
+interface JsonWriter {
```
JamesHR3: Might be worth sticking a `@DoNotImplement` on here to signal that we give you one of these and you're not expected to provide one yourself.

application/src/main/kotlin/net/corda/v5/application/marshalling/JsonSerializedBase64Config.kt
JamesHR3: Thinking about it a little more, I'd expect to only actually need the serializer when thinking about flows, so probably this is the right module for it at least.

simon-johnson-r3: It's at the same level as JsonMarshallingService now. I can move it all of course. I wonder if we are going a bit package happy though, even Marshalling service mentions that its only subclass is JsonMarshallingService right now. I don't mind moving it all and removing that reference (this is a breaking change but that doesn't really matter) if that is what the consensus is?

application/src/main/kotlin/net/corda/v5/application/marshalling/JsonSerializedBase64Config.kt
simon-johnson-r3: I agree on the point about describing each, I will action that.

The default is described at the point of method overload, at the moment it's MIME_NO_LINEFEEDS as per stock Jackson. I'm not sure moving it here makes sense, because users who are using the overloads that don't take an enum are exactly the audience who wont come to this enum to read the docs (aside from the fact it could technically be different in different places - i.e. the enum declaration itself is not the decider of the default).

application/src/main/kotlin/net/corda/v5/application/marshalling/JsonSerializedBase64Config.kt
simon-johnson-r3: I've added @mnesbit to the review as well, probably should have done that anyway.

JamesHR3 (approved):

PR #627, opened 2022-10-03
CORE-5088 Consensual ledger persistence

mbrkic-r3:
- Persistence messages moved to new schema for persistence messages.
- Added field "signature.pub_key_hash" in consensual ledger DB schema.


data/db-schema/src/main/resources/net/corda/db/schema/vnode-vault/migration/ledger-consensual-creation-v1.0.xml
```diff
@@ -39,7 +39,10 @@
             <column name="signature_idx" type="INT">
                 <constraints nullable="false"/>
             </column>
-            <column name="signature" type="VARBINARY(64)">
+            <column name="signature" type="VARBINARY(1048576)">
```
omfR3, line 42: What exactly is the signature? I initially confused it with a hash, but I wouldn't expect a digital signature to be more than 1 or 2kb long. Although I don't think this makes much difference to Postgres since it's pretty clever about storage optimisation.

omfR3:
:+1:

data/db-schema/src/main/resources/net/corda/db/schema/vnode-vault/migration/ledger-consensual-creation-v1.0.xml
mbrkic-r3: This column will contain serialized `DigitalSignatureAndMetadata`, which also contains public key related to signature, context passed to signing operation  and `DigitalSignatureMetadata`. I'm not sure what would be expected size of all that, this could be adjusted later. But as you said it's not important for Postgres as it ends up using type `bytea`.

blsemo (dismissed):
Looks ok to me

blsemo (approved):

PR #628, opened 2022-10-03
CORE-6316 - Add configuration for max request size

dimosr:
Adding a new configuration attribute for the max request size acceptable by the p2p gateway.

Used by: https://github.com/corda/corda-runtime-os/pull/2227


data/config-schema/src/main/resources/net/corda/schema/configuration/p2p.gateway/1.0/corda.p2p.gateway.json
```diff
@@ -41,6 +41,12 @@
         "additionalProperties": false
       }
     },
+    "maxRequestSize": {
+      "description": "The maximum request size accepted by the gateway, in bytes. Any requests crossing that threshold will be rejected.",
```
yift-r3, line 45: Maybe we should use kb or mb (or at least set the default and minimum using a scientific notation (`500000000` -> `5e8`))?

dimosr: I think kb/mb etc. is not natively supported and I would be relucant to start adding custom parsing code. But, scientific notation seems to be supported. So, I can give that a try.

yift-r3: I meant to change the description to: `... request size accepted by the gateway, in megabytes ...` (But I'm OK either way)

yift-r3 (dismissed):

yift-r3 (approved):

PR #629, opened 2022-10-05
CORE-6875 Json deserializer interfaces

simon-johnson-r3:
Comprises of:
* The deserializer interface which custom Json deserializers implement.
* The interface that can parse Json. The implementation will forward all calls to Jackson, this is part of a later story.

Notes:
* Json can of course model a tree of nodes, where each node can have properties and other nodes and a relationship like this is exposed by Jackson’s deserializer interface. This interface does the same albeit in a simplified fashion, offering a decent set of reading functions without adding too much complexity at this stage.
* Tests to ensure the interfaces are Java compliant.
* These interfaces will be part of the Corda public application api, but there will be nothing anyone can do with them at the moment.
* This is not a breaking change.


application/src/main/kotlin/net/corda/v5/application/marshalling/json/JsonNodeReader.kt
```diff
@@ -0,0 +1,387 @@
...
+    fun binaryValue(): ByteArray?
+
+    /**
+     * Returns true if this node represents the value of "null" in the underlying Json.
+     *
+     * @return true if null, otherwise false.
+     */
+    fun isNull(): Boolean
+
+    /**
```
simon-johnson-r3, line 309: @mnesbit This is the functionality you were enquiring about in the standup.

application/src/main/kotlin/net/corda/v5/application/marshalling/json/JsonWriter.kt
```diff
@@ -63,12 +63,10 @@ interface JsonWriter {
      * Writes a string value.
```
simon-johnson-r3, line 63: The changes in this file reflect tweaks required to the JsonWriter interface when work was started on the implementation of that in the corda-runtime-os repo. It seems easier to push them in here than start a new PR with the same reviewers.

Although some of these changes would normally constitute breaking changes, this interface is not yet used anywhere in Corda, so nothing can break as a result of it anymore than it could when they were first added.

application/src/main/kotlin/net/corda/v5/application/marshalling/json/JsonWriter.kt
```diff
@@ -63,12 +63,10 @@ interface JsonWriter {
...
      *
      * @param text The string to write.
-     * @param offset The offset into the text to begin writing from.
-     * @param len The length of data to write.
      *
      * @throws IOException if there is a writing error.
      */
```
simon-johnson-r3, line 68: This method and the next were not supported in their original form by Jackson when I tried to plumb it in.

application/src/main/kotlin/net/corda/v5/application/marshalling/json/JsonWriter.kt
```diff
@@ -401,6 +398,13 @@ interface JsonWriter {
      * data is intended to be a field or value, it does not add any separators or escaping, nor does it require that the
      * Json output that results is valid Json.
      *
```
simon-johnson-r3, line 400: Documentation tweaks after I saw the Json output in tests to make it clearer to the end user what the result of certain actions are.

application/src/test/java/net/corda/v5/application/marshalling/json/JsonWriterJavaApiTest.java
```diff
@@ -43,7 +43,7 @@ void callEveryMethod() {
         jw.writeFieldName(TEST_STRING);
 
         jw.writeString(TEST_CHAR_ARRAY, 1, 1);
```
simon-johnson-r3, line 45: Test fixes due to the api changes.

fowlerrr (dismissed):
Have you had a chance to implement this yet?  Just wondering if you'll run into the same problem as with the serializers.  Though, I agree that we can update them as necessary (and not worry about version change) since they're new.

simon-johnson-r3:
> Have you had a chance to implement this yet? Just wondering if you'll run into the same problem as with the serializers. Though, I agree that we can update them as necessary (and not worry about version change) since they're new.

I've implemented them locally and not noticed any problems, but only tests use the new interfaces at the moment. In any case I've added the required OSGI export now.

fowlerrr (dismissed):

mnesbit (approved):
OK

PR #634, opened 2022-10-06
CORE-7034: draft new higher level signing interface for discussion

dickon:
Add higher level signing operations that work out SignatureSpec based on key and possibly digest algorith,.


application/src/main/kotlin/net/corda/v5/application/crypto/SigningService.kt
```diff
@@ -20,10 +21,36 @@ import java.util.*
...
+     * primary identity, or previously generated via the freshKey method. If the [PublicKey] is actually
+     * a [CompositeKey], the first leaf signing key hosted by the node is used.
+     * @param digestAlgorithmName the message digest algorithm name; if null have the crypto service choose one.
+     * @return A [DigitalSignature.WithKey] representing the signed data and the [PublicKey] that belongs to the
+     * same [KeyPair] as the [PrivateKey] that signed the data. The signature type will be determined automatically.
+     *
+     * @throws [CordaRuntimeException] If the input key is not a member of [keys].
+     */
+    @Suspendable
+    fun sign(payload: ByteArray, metadata: Map<String, String>?, publicKey: PublicKey, digestAlgorithmName: DigestAlgorithmName?): DigitalSignature.WithKey
```
alexey-kadyrov-r3: The return values should be changed to `DigitalSignature.WithMetadata` where it's defined as

```kotlin
/**
 * A wrapper around a digital signature.
 *
 * @property bytes The signature.
 * @param bytes The signature.
 */
@CordaSerializable
open class DigitalSignature(
    bytes: ByteArray
) : OpaqueBytes(bytes) {
    /**
     * A digital signature that identifies who the public key is owned by.
     *
     * @param by The public key of the corresponding private key used to sign the data (as if an instance
     * of the [CompositeKey] is passed to the sign operation it may contain keys which are not actually owned by
     * the member).
     * @param bytes The signature.
     * @param context The context which was passed to the signing operation, note that this context is not signed over.
     */
    open class WithKey(
        /**
         * Public key which corresponding private key was used to sign the data (as if an instance
         * of the [CompositeKey] is passed to the sign operation it may contain keys which are not actually owned by
         * the member).
         */
        val by: PublicKey,
        bytes: ByteArray,
        /**
         * The context which was passed to the signing operation, note that this context is not signed over.
         */
        val context: Map<String, String>
    ) : DigitalSignature(bytes)

    class WithMetadata(
        /**
         * Public key which corresponding private key was used to sign the data (as if an instance
         * of the [CompositeKey] is passed to the sign operation it may contain keys which are not actually owned by
         * the member).
         */
        by: PublicKey,
        bytes: ByteArray,
        /**
         * The context which was passed to the signing operation, note that this context is not signed over.
         */
        context: Map<String, String>,
        /**
         * The serialised metadata which was used by the signing operation, note that the metadata **is signed over**.
         * The actual signing data is concatenation of the serialised data and the user data.
         */
        metadata: ByteArray,
    ) : WithKey(by, bytes, context)
}
```

application/src/main/kotlin/net/corda/v5/application/crypto/SigningService.kt
```diff
@@ -35,5 +62,17 @@ interface SigningService {
...
+     * Work out the best [SignatureSpec] to use for a given public key and possibly a specific digest algorithm
+     *
+     * @param publicKey The [PublicKey], inspected for its type.
+     * @param digestAlgorithmName Optionally, the digest algorithm. If not specified, pick the most appropriate
+     * digest algorithm
+     *
+     * @return a [SignatureSpec] that is appropriate.
+     */
+
+    fun selectSignatureSpec(publicKey: PublicKey, digestAlgorithmName: DigestAlgorithmName?): SignatureSpec
```
alexey-kadyrov-r3: That one would be better to be moved to a separate service

application/src/main/kotlin/net/corda/v5/application/crypto/DigitalSignatureVerificationService.kt
```diff
@@ -13,17 +15,35 @@ import java.security.PublicKey
...
+     * @param payload The payload/message that may be in clear data that was signed (e.g. the Merkle root).
+     * @param signatureData The signature data and public key, which should have been earlier produced by [SigningService.sign].
+     * @param signatureSpec The expected specific signature spec.
      *
      * @throws CryptoSignatureException If verification of the digital signature fails.
      * @throws IllegalArgumentException If the signature scheme is not supported or if any of the clear or signature
      * data is empty.
      */
-    fun verify(publicKey: PublicKey, signatureSpec: SignatureSpec, signatureData: ByteArray, clearData: ByteArray)
+    fun verify(payload: ByteArray, signatureData: DigitalSignature.WithKey, signatureSpec: SignatureSpec): Map<String, String>
```
alexey-kadyrov-r3: Really is should be
```kotlin
fun verify(payload: ByteArray, metadata: ByteArray, publicKey: PublicKey): Map<String, String>
```

alexey-kadyrov-r3: Also I would prefer string typed return value

crypto/src/main/kotlin/net/corda/v5/crypto/DigitalSignature.kt
```diff
@@ -36,4 +36,19 @@ open class DigitalSignature(
...
+         * of the [CompositeKey] is passed to the sign operation it may contain keys which are not actually owned by
+         * the member).
+         */
+        by: PublicKey,
+        bytes: ByteArray,
+        /**
+         * The metadata which was used by the signing operation, note that the metadata **is signed over**.
+         * The format will be stable and consistent. The actual signing data is concatenation of the serialised metadata and the  payload data.
+         */
+        metadata: Map<String, String>,
```
alexey-kadyrov-r3: I steel feel that must be the byte array as we will be signing over it so it must be predicable on the other side to verify as well. Plus that metadata will contain additional data compared to the metadata which was passed to the sign function - namely the signature spec in some form

alexey-kadyrov-r3: We can drop the context though and add `val` to metadata and WithKey could be deleted

application/src/main/kotlin/net/corda/v5/application/crypto/DigitalSignatureAndMetadata.kt
```diff
@@ -13,6 +13,8 @@ import net.corda.v5.crypto.DigitalSignature
  *
  * @constructor Creates a [DigitalSignatureAndMetadata].
  */
+// TODO - remove this since public key should be handled separately from the signature, and not simply extracted from it
```
mnesbit: We do need an identifier of who signed. I would be fine with it being hash of publickey, so you are forced to do lookup, but we need something to correlate when we have multiple signatures.

application/src/main/kotlin/net/corda/v5/application/crypto/SigningService.kt
mnesbit: Yes, probably should be separate service. I think we would also need a `List<SignatureSpec>` returning version. Also, this one can fail so we have to decide if it return null, or it throws. Behaviour around `CompositeKey `should also be explicit (probably throws)

crypto/src/main/kotlin/net/corda/v5/crypto/DigitalSignature.kt
```diff
@@ -36,4 +36,19 @@ open class DigitalSignature(
          */
         val context: Map<String, String>
```
mnesbit: Why is the class open?

mnesbit: The context is also really vague and either we need to say what you can expect in it, or remove it

alexey-kadyrov-r3: I say remove it, the class was left open for no reason other than oversight

crypto/src/main/kotlin/net/corda/v5/crypto/DigitalSignature.kt
mnesbit: I agree it has to be predictably serialized. This is where the old SignableData was clearer in use. You put your data's SecureHash and metadata into it and we AMQP serialized it and signed that. It therefore acted as a public specification of the approach. The Map<String, String> begs questions such as do we sort by key and of course how do we serialize (AMQP?, Simply concatenate sorted UTF8 pairs?)

alexey-kadyrov-r3: Yes, before the map is serialised it must be sorted by the keys and on the deserialisation side the order should be checked - similar how MGM behaves with the layered map

application/src/main/kotlin/net/corda/v5/application/crypto/SigningService.kt
dickon: Done

crypto/src/main/kotlin/net/corda/v5/crypto/DigitalSignature.kt
dickon: Done

application/src/main/kotlin/net/corda/v5/application/crypto/SigningService.kt
```diff
@@ -20,10 +21,36 @@ import java.util.*
...
+     * looks up the matching [PrivateKey] and signs the data, selecting the most appropriate signing type.
+     * This overload is preferred; see the other overload if you need to control the signing type.
+     *
+     * @param payload The data to sign over using the chosen key.
+     * @param metadata Optional map of string configuration metadata to be included in the signature. If the metadata
+     *                 is null, or if there are no know entries in the map, that's equivalent to calling `sign` without
+     *                 metadata.
+     * @param publicKey The [PublicKey] partner to an internally held [PrivateKey], either derived from the node's
+     * primary identity, or previously generated via the freshKey method. If the [PublicKey] is actually
+     * a [CompositeKey], the first leaf signing key hosted by the node is used.
```
mnesbit: This could be expensive to do, but it does make sense from a usability perspective.

application/src/main/kotlin/net/corda/v5/application/crypto/SigningService.kt
```diff
@@ -20,10 +21,36 @@ import java.util.*
...
+     * @param digestAlgorithmName the message digest algorithm name; if null have the crypto service choose one.
+     * @return A [DigitalSignature.WithKey] representing the signed data and the [PublicKey] that belongs to the
+     * same [KeyPair] as the [PrivateKey] that signed the data. The signature type will be determined automatically.
+     *
+     * @throws [CordaRuntimeException] If the input key is not a member of [keys].
+     */
+    @Suspendable
+    fun sign(payload: ByteArray, metadata: Map<String, String>?, publicKey: PublicKey, digestAlgorithmName: DigestAlgorithmName?): DigitalSignature.WithMetadata
```
mnesbit: In practice we will tend to be signing over a SecureHash from some merkle tree. Should we provide a version that takes SecureHash for payload? Perhaps that should be the only version? We would have to decide if we can avoid a second round of hashing if the input is of same type as the digest and there is a compatible signaturespec. I think we also need to document precisely how we form the bytes that we sign at the raw level for interop purposes.

crypto/src/main/kotlin/net/corda/v5/crypto/DigitalSignature.kt
mnesbit: I do think the DigitalSignatureAndMetadata was clearer than this internal class and note that this isn't marked as CordaSerializable, so it isn't in fact going to work. We should also make the metadata SortedMap<String,String> so that the ordering is explicit (and perhaps to some degree automatic?)

application/src/main/kotlin/net/corda/v5/application/crypto/DigitalSignatureVerificationService.kt
```diff
@@ -13,17 +15,36 @@ import java.security.PublicKey
...
+     * @param payload The payload/message that may be in clear data that was signed (e.g. the Merkle root).
+     * @param signatureData The signature data and public key, which should have been earlier produced by [SigningService.sign].
+     * @param publicKey A verified public key aginst which to verify the signature.
+     * @return The metadata associated with the signature, passed in to [SigningService.sign]
+     *
+     * @throws CryptoSignatureException If verification of the digital signature fails.
+     * @throws IllegalArgumentException If the signature scheme is not supported or if any of the clear or signature
+     * data is empty.
+     */
+    fun verify(payload: ByteArray, signatureData: DigitalSignature.WithMetadata, publicKey: PublicKey, digestAlgorithmName: DigestAlgorithmName?): Map<String, String>
```
mnesbit: If we put the SignatureSpec onto the metadata before signing then there is no need to pass in the digest name and the return is not needed either as it is in the object

mnesbit:
My own playing around lead to:

```
@CordaSerializable
data class DigitalSignatureMetadata(
    val timestamp: Instant,
    val signatureSpec: SignatureSpec,
    val properties: Map<String, String>
)

@CordaSerializable
data class DigitalSignatureAndMetadata(
    val signature: DigitalSignature.WithKeyHash,
    val metadata: DigitalSignatureMetadata
)

interface SignatureSpecService {
    @Suspendable
    fun getDefaultSignatureSpec(publicKey: PublicKey, digestAlgorithm: DigestAlgorithmName): SignatureSpec?

    @Suspendable
    fun getCompatibleSignatureSpecs(publicKey: PublicKey, digestAlgorithm: DigestAlgorithmName): List<SignatureSpec>
}

interface SigningService {
    @Suspendable
    fun sign(
        bytes: ByteArray,
        publicKey: PublicKey,
        signatureSpec: SignatureSpec
    ): DigitalSignature.WithKeyHash

    @Suspendable
    fun sign(
        bytes: ByteArray,
        metadataProperties: Map<String, String>,
        digestAlgorithm: DigestAlgorithmName,
        publicKey: PublicKey
    ): DigitalSignatureAndMetadata

}

@DoNotImplement
interface DigitalSignatureVerificationService {
    @Suspendable
    fun verify(
        bytes: ByteArray,
        expectedPublicKey: PublicKey,
        signatureSpec: SignatureSpec,
        signature: DigitalSignature
    )

    @Suspendable
    fun verify(
        bytes: ByteArray,
        expectedPublicKey: PublicKey,
        signature: DigitalSignatureAndMetadata
    )

}
```

mnesbit:
I also think this PR needs to include re-adding a HashDigestFactory to application/crypto, because it is currently sending people off to wrong module to make SecureHash

dickon:
> My own playing around lead to:
> 
> ```
> @CordaSerializable
> data class DigitalSignatureMetadata(
>     val timestamp: Instant,
>     val signatureSpec: SignatureSpec,
>     val properties: Map<String, String>
> )

I'd want to make this named `metadataProperties`,  to be consistent with 
> 
> @CordaSerializable
> data class DigitalSignatureAndMetadata(
>     val signature: DigitalSignature.WithKeyHash,
>     val metadata: DigitalSignatureMetadata
> )
> 
> interface SignatureSpecService {
>     @Suspendable
>     fun getDefaultSignatureSpec(publicKey: PublicKey, digestAlgorithm: DigestAlgorithmName): SignatureSpec?
> 
>     @Suspendable
>     fun getCompatibleSignatureSpecs(publicKey: PublicKey, digestAlgorithm: DigestAlgorithmName): List<SignatureSpec>
> }
> 
> interface SigningService {
>     @Suspendable
>     fun sign(
>         bytes: ByteArray,
>         publicKey: PublicKey,
>         signatureSpec: SignatureSpec
>     ): DigitalSignature.WithKeyHash
> 

Any reason not to put `metadataProperites` on this overload as well as the next one?

The caller still has to cope with consistently putting the DiigitalSignture together with the payload and signed-over metadata, in such a way that verify will take it.

So I think we either returned a new `SignedMessage` data class, which has a toBytes() method, or return the bytes directly if we don't want people poking around inside the signed message.
>     @Suspendable
>     fun sign(
>         bytes: ByteArray,
>         metadataProperties: Map<String, String>,
>         digestAlgorithm: DigestAlgorithmName,
>         publicKey: PublicKey
>     ): DigitalSignatureAndMetadata

So, you're requiring people here to get any non-string  metadata down to strings, ideally in a standardised way? I imagine this could be a source of errors, so we may want to support a small number of additional metadata arguments, e.g. secureHashMetadataProperites of tpe Map<String, SecureHash> with the requirement that the keys do not conflict, so that the caller doesn't need to figure out how to do that.
 
> }
> 
> @DoNotImplement
> interface DigitalSignatureVerificationService {
>     @Suspendable
>     fun verify(
>         bytes: ByteArray,
>         expectedPublicKey: PublicKey,
>         signatureSpec: SignatureSpec,
>         signature: DigitalSignature
>     )
>     @Suspendable
>     fun verify(
>         bytes: ByteArray,
>         expectedPublicKey: PublicKey,
>         signature: DigitalSignatureAndMetadata
>     )
> 
> }

Should this return the metadata, so that a caller can pass in the full `bytes` of the signed document, and `verify` would pull out the metadata, and return it deserialised?  
> ```

mnesbit:
The intent was that the
```
@Suspendable
fun sign(
bytes: ByteArray,
publicKey: PublicKey,
signatureSpec: SignatureSpec
): DigitalSignature.WithKeyHash
```
is a raw signing facility it signs precisely the bytes you give it.
the 
```
@Suspendable
fun sign(
    bytes: ByteArray,
    metadataProperties: Map<String, String>,
    digestAlgorithm: DigestAlgorithmName,
    publicKey: PublicKey
): DigitalSignatureAndMetadata
```
version forms a `DigitalSignatureMetadata` with the chosen `SignatureSpec` set correctly and then concatenates the users's `ByteArray` and the Serialized DigitalSignatureMetadata to sign. returning the signature itself plus metadata in the DigitalSignatureAndMetadata. So this is a safer to use form, but it does include some implicit logic around precisely how the bytes to sign are formed. I would be fine with making public (as we used to) a
```
@CordaSerializable
data class SignableData(val bytes: ByteArray, val metadata: DigitalSignatureMetadata)
```
to document this serialization better and allow you to independently construct a signature from the raw API.

Also, it is becoming clear that we need a version of sign that can identify which keys in a set are owned by you and return those  signatures. This is needed for ledger logic, though we definitely do need the internal service in Flow Engine to store key mappings so that we can reduce the number of expensive queries required to locate our own keys. A method signature would probably be:

```
@Suspendable
fun signWithOwnedKeys(
    bytes: ByteArray,
    metadataProperties: Map<String, String>,
    digestAlgorithm: DigestAlgorithmName,
    publicKey: Set<PublicKey>
): List<DigitalSignatureAndMetadata>
```
see https://r3-cev.atlassian.net/browse/CORE-7031 for some context

application/src/main/kotlin/net/corda/v5/application/crypto/DigitalSigningService.kt
```diff
@@ -0,0 +1,52 @@
...
+     * @param metadata Additional data to sign over. Pass in Unit if unneeded. Must be Corda serializable.
+     * @param publicKeys A set of [PublicKey]. Some of these public keys must have private keys available in the crypto service.
+     *                   The result will include signatures for each field.
+     * @param digestAlgorithmName the message digest algorithm name; if null have the crypto service choose one.
+     * @return A [DigitalSignature.WithKey] representing the signed data and the [PublicKey] that belongs to the
+     * same [KeyPair] as the [PrivateKey] that signed the data. The signature type will be determined automatically.
+     *
+     */
+    @Suspendable
+    fun <T, M> sign(payload: T, metadata: M, publicKeys: Set<PublicKey>, digestAlgorithmName: DigestAlgorithmName?): DigitalSignedDocument<T, M>
```
alexey-kadyrov-r3, line 34: You will have to define and describe the serialisation/deserialisation process which is predictable, remember that some serialisation libraries may produce different ByteArray depending on the library and JVM versions.
We either define strict serialisation or use ByteArrays here.

application/src/main/kotlin/net/corda/v5/application/crypto/DigitalSigningService.kt
```diff
@@ -0,0 +1,53 @@
...
+import java.security.KeyPair
+import java.security.PrivateKey
+import java.security.PublicKey
+import java.util.*
+
+/**
+ * Responsible for storing and using private keys to sign things. An implementation of this may, for example, call out
+ * to a hardware security module that enforces various auditing and frequency-of-use requirements.
+ *
+ * Corda provides an instance of [DigitalSignatureVerificationService] to flows via property injection.
```
lankydan, line 15: References wrong class.

application/src/main/kotlin/net/corda/v5/application/crypto/DigitalSigningService.kt
```diff
@@ -0,0 +1,53 @@
...
+ */
+@DoNotImplement
+interface DigitalSigningService {
```
lankydan, line 18: Maybe its just me, but `DigitalSigningService` sounds a bit odd to me. I'd prefer `DigitalSingatureService` (rolls off the tongue better) but then that sounds like a more general service and might confuse things since we have `DigitalSignatureVerificationService`. Should we go as far as to name it `DigitalSingatureSigningService`? That would pair it up with the verification version more closely.

application/src/main/kotlin/net/corda/v5/application/crypto/DigitalSigningService.kt
```diff
@@ -0,0 +1,53 @@
...
+     * @param metadata Additional data to sign over. Pass in Unit if unneeded. Must be Corda serializable.
+     * @param publicKeys A set of [PublicKey]. Some of these public keys must have private keys available in the crypto service.
+     *                   The result will include signatures for each field.
+     * @param digestAlgorithmName the message digest algorithm name; if null have the crypto service choose one.
+     * @return A [DigitalSignature.WithKey] representing the signed data and the [PublicKey] that belongs to the
+     * same [KeyPair] as the [PrivateKey] that signed the data. The signature type will be determined automatically.
+     *
+     */
+    @Suspendable
+    fun <T, M> sign(payload: T, metadata: M, publicKeys: Set<PublicKey>, digestAlgorithmName: DigestAlgorithmName?): DigitalSignedDocument<T, M>
```
lankydan, line 34: Include `T` and `M` in the params.

crypto/src/main/kotlin/net/corda/v5/crypto/DigitalSignedDocument.kt
```diff
@@ -0,0 +1,16 @@
+package net.corda.v5.crypto
+
+import net.corda.v5.base.annotations.CordaSerializable
+
+
+@CordaSerializable
```
lankydan, line 6: Annotation should be after the kdoc

application/src/main/kotlin/net/corda/v5/application/crypto/DigitalSignatureVerificationService.kt
```diff
@@ -13,17 +15,18 @@ import java.security.PublicKey
...
+     * @param payload The payload/message that may be in clear data that was signed (e.g. the Merkle root).
+     * @param publicKeys A set of verified public keys to use for verification. Some of the signatures in signed document must have keys in this set.
+     * @return The metadata associated with the signature, passed in to [DigitalSigningService.sign]
      *
      * @throws CryptoSignatureException If verification of the digital signature fails.
      * @throws IllegalArgumentException If the signature scheme is not supported or if any of the clear or signature
      * data is empty.
      */
-    fun verify(publicKey: PublicKey, signatureSpec: SignatureSpec, signatureData: ByteArray, clearData: ByteArray)
+    fun <T, M> verify(payload: DigitalSignedDocument<T, M>, publicKeys: Set<PublicKey>): List<DigitalSignature>
```
lankydan, line 31: think this should be `document`, `signedDocument` or `digitalSignedDocument` instead of `payload`.

crypto/src/main/kotlin/net/corda/v5/crypto/DigitalSignedDocument.kt
```diff
@@ -0,0 +1,16 @@
...
+/** A representation of a signed document,
+ *
+ * @param payload the payload that is to be signed, which must be Corda serializable
+ * @param metadata additional records. Needs to be Corda serializable..
+ */
+class DigitalSignedDocument<T, M>(
+    val payload: T,
```
kyriathar, line 13: Based on this design proposal, `DigitalSignedDocument` is what `sign` produces and what `verify` consumes and that makes the two APIs symmetrical. However, my understanding is that our main use case for `sign` and `verify` APIs is:

A has a payload to be signed by B, then it sends it across the wire to B, and B signs it and sends back to A its signature to be verified by A who already possess the payload that was signed, so why B send back again the payload itself. This might also means if B starts acting maliciously, and A was lazy enough to not compare its payload with the payload that was sent back to be the same, B could have changed the payload over which they signed. 

Well, we could be recreating `DigitalSignedDocument` to be passed in on verify side with the payload A holds and then maybe B just sends over only the signature from this structure, but this structure pretty much implies to me at least that it gets returned from `sign` operation to be passed in to `verify` operation.

Less importantly, it also means overhead to the wire.

So maybe we should not be including the payload here? Or will there be a need for us to include the payload as well here? @mnesbit @lankydan what do you think?

mnesbit:
This seems to have diverged considerably from what I thought we had agreed! I thought we had agreed:
1. Keep signing and verifying as basic operations that sign bytes and verify bytes against signature object
2. Do not regard SignatureSpec as being inside signed materials as part of signing service API. Instead have separated service to get correct SignatureSpec for a specific public key and hashing algorithm name. Up to user to make byte array that does contain this info and then sign
3. Put in separate service for that SignatureSpec determination that can get 'default' SignatureSpec, but also list of all compatible SignatureSpecs.
4. Make output of signing DigitalSignature.WithId rather than DigitalSignature.WithKey to prevent mis-use of the PublicKey in DigitalSignature.WithKey.
5. Put back in a proper DigestService into application/crypto
6. Provide a method that will take a set of public keys and can sign with those keys (or leaves of any composite keys) that we own.

dickon:
Closing after slack discussion with @mnesbit's summarised in his comment this morning.

The core question is who does the serialisation; this design is the high level approach where the crypto code handles the serialisation. There's another approach which we'll cover in a separate draft design PR where the caller (ledger or flow) does the serialisation, and so the Corda low-level crypto implementation stays simple.

PR #637, opened 2022-10-07
CORE-6784 utxo api

blsemo:
I've picked up Matts PR https://github.com/corda/corda-api/pull/613 and simplified the API further according to the comments,


ledger/ledger-utxo/src/main/kotlin/net/corda/v5/ledger/utxo/TransactionState.kt
```diff
@@ -0,0 +1,25 @@
...
+ * @property contractStateType The [ContractState] type of the current [TransactionState] instance.
+ * @property contractType The [Contract] type of the current [TransactionState] instance.
+ * @property notary The notary of the current [TransactionState] instance.
+ * @property encumbrance The encumbrance of the current [TransactionState] instance.
+ */
+@DoNotImplement
+@CordaSerializable
+interface TransactionState<out T : ContractState> {
+    val contractState: T
+    val contractStateType: Class<out T>
```
lankydan, line 21: What do we think `contractStateType` would be used for?

ledger/ledger-utxo/src/main/kotlin/net/corda/v5/ledger/utxo/UtxoLedgerService.kt
```diff
@@ -0,0 +1,56 @@
...
+
+    /**
+     * Resolves the specified [StateRef] instances into [StateAndRef] instances of the specified [ContractState] type.
+     *
+     * @param T The underlying [ContractState] type.
+     * @param stateRefs The [StateRef] instances to resolve.
+     * @return Returns a [List] of [StateAndRef] of the specified [ContractState] type.
+     */
+    @Suspendable
+    fun <T : ContractState> resolve(stateRefs: Iterable<StateRef>): List<StateAndRef<T>>
```
lankydan, line 29: I'd be keen for some of these methods to be moved to another service at some point in the future.

ledger/ledger-utxo/src/main/kotlin/net/corda/v5/ledger/utxo/transaction/UtxoTransactionBuilder.kt
```diff
@@ -0,0 +1,152 @@
...
+     * @return Returns a [UtxoTransactionBuilder] including the specified time window.
+     */
+    fun setTimeWindowBetween(from: Instant, until: Instant): UtxoTransactionBuilder
+
+    /**
+     * Signs the transaction with any required signatories that belong to the current node.
+     *
+     * @return Returns a [UtxoSignedTransaction] with signatures for any required signatories that belong to the current node.
+     */
+    fun sign(): UtxoSignedTransaction
```
lankydan, line 130: It's called `signInitial` on `ConsensualTransactionBuilder`, we should plan on aligning them.

ledger/ledger-utxo/src/main/kotlin/net/corda/v5/ledger/utxo/StateRef.kt
```diff
@@ -0,0 +1,13 @@
...
+import net.corda.v5.crypto.SecureHash
+
+/**
+ * Defines a reference to a [ContractState].
+ *
+ * @property index The index of the state in the transaction's outputs in which the referenced state was created.
+ * @property transactionHash The hash of the transaction in which the referenced state was created.
+ */
+@CordaSerializable
+data class StateRef(val transactionHash: SecureHash, val index: Int)
```
MrMatthewLayton: @blsemo @mnesbit We keep moving the goalposts with this...

One argument is that if APIs are just POJO data containers, then just use data classes. This is what I originally did for `StateRef`, `StateAndRef`, `TransactionState`, `TimeWindow`, and `Party`.

Another argument is that if APIs are only meant to be instantiated by the runtime and not by the user, then make them interfaces with runtime implementations, hence the second iteration of this work was just so.

Now we have a mismatch of interfaces and a single data class, for which there is little justification, and I personally feel like it should be all one or the other for consistency.

Can you justify why only `StateRef` deserves to be a data class in the API, and why the others deserve no such amendment? (If so, perhaps this justification ought to be documented in the API)

ledger/ledger-utxo/src/main/kotlin/net/corda/v5/ledger/utxo/TimeWindow.kt
```diff
@@ -0,0 +1,30 @@
...
+ * issues such as network latency. A time window represents an approximation of an instant with a margin of tolerance,
+ * and may be fully bounded, or unbounded towards negative or positive infinity.
+ *
+ * @property from The boundary at which the time window begins, or null if the time window is unbounded towards negative infinity.
+ * @property until The boundary at which the time window ends, or null if the time window is unbounded towards positive infinity.
+ */
+@DoNotImplement
+@CordaSerializable
+interface TimeWindow {
+    val from: Instant?
```
MrMatthewLayton: Should these still be nullable?

ledger/ledger-utxo/src/main/kotlin/net/corda/v5/ledger/utxo/TimeWindow.kt
```diff
@@ -0,0 +1,30 @@
...
+    val until: Instant?
```
MrMatthewLayton: As above...

ledger/ledger-utxo/src/main/kotlin/net/corda/v5/ledger/utxo/UtxoLedgerService.kt
MrMatthewLayton: @lankydan What would be your rationale for moving ledger related functionality to another service, and which services did you have in mind?

ledger/ledger-utxo/src/main/kotlin/net/corda/v5/ledger/utxo/transaction/UtxoLedgerTransaction.kt
```diff
@@ -0,0 +1,189 @@
...
+ * @property referenceInputContractStates The input contract states to be referenced by the current [UtxoLedgerTransaction].
+ * @property outputStateAndRefs The output states and state references to be created by the current [UtxoLedgerTransaction].
+ * @property outputTransactionStates The output transaction states to be created by the current [UtxoLedgerTransaction].
+ * @property outputContractStates The output contract states to be created by the current [UtxoLedgerTransaction].
+ */
+@DoNotImplement
+@Suppress("TooManyFunctions")
+interface UtxoLedgerTransaction {
+
+    val timeWindow: TimeWindow?
```
MrMatthewLayton: Should not be nullable as I understand?

ledger/ledger-utxo/src/main/kotlin/net/corda/v5/ledger/utxo/transaction/UtxoTransactionBuilder.kt
MrMatthewLayton: @lankydan @blsemo @mnesbit I agree that we should align, although I feel like "initial" is redundant, since we rightly call it a `TransactionBuilder`, not an `InitialTransaction`.

ledger/ledger-utxo/src/main/kotlin/net/corda/v5/ledger/utxo/ContractState.kt
```diff
@@ -0,0 +1,21 @@
...
+ *
+ * States are immutable. Once created they are never updated, instead, any changes must generate a new successor state.
+ * States can be updated (consumed) only once. The notary is responsible for ensuring there is no "double spending" by
+ * only signing a transaction if the input states are all free.
+ *
+ * @property participants The public keys of any participants associated with the current contract state.
+ */
+@CordaSerializable
+interface ContractState {
+    val participants: List<PublicKey>
```
MrMatthewLayton, line 20: This has been a bugbear of mine for a long time in C4 - `Set` semantics make more sense here than `List` semantics.

Consider...
```kotlin
data class Obligation(val obligee: PublicKey, val obligor: PublicKey) : ContractState {
    override val participants: List<PublicKey> get() = listOf(obligee, obligor)
}
```

The correct usage of this would be:
```kotlin
val obligation = Obligation(aliceKey, bobKey)
obligation.participants.size == 2 // true
```

But you could easily go wrong:
```kotlin
val obligation = Obligation(aliceKey, aliceKey)
obligation.participants.size == 2 // true
```
**Why are there two participants in this case...it's just Alice?**
Okay...it's a computed property; serializers should just ignore it and it won't bloat the ledger, but even so, there is only **one** participant! Maybe in this case, you could implement a contract constraint to check that `obligor` and `obligee` aren't the same, but that's domain specific logic and won't apply to all contract state implementations.

You could do `listOf(obligor, obligee).distinct()` which in this case is what I would do (IIRC, this just calls `.toSet().toList()` under the hood)

In the event that the `participants` property for whatever reason must maintain order, then let the implementor decide that for themselves:
```kotlin
data class Obligation(val obligor: PublicKey, val obligee: PublicKey) {
    override val participants: Set<PublicKey> get() = linkedSetOf(obligor, obligee)
}
```

I certainly see no reason why this should be a ledger level behaviour.

In the event that you are using a key whose entropy is subject to change, or simpler, two keys owned by the same _identity_ then just treat them as different keys, which I think is okay, but I definitely don't believe that the semantics of this property should allow duplicate keys!

P.S. The same applies elsewhere in the codebase for `signatories` and `signature` properties.

ledger/ledger-common/src/main/kotlin/net/corda/v5/ledger/common/Party.kt
```diff
@@ -0,0 +1,17 @@
...
+import java.security.PublicKey
+
+/**
+ * Defines a well-known identity.
+ *
+ * @property name The well-known [MemberX500Name] that represents the current identity.
+ * @property owningKey The [PublicKey] that represents the current identity.
+ */
+@CordaSerializable
+interface Party {
```
MrMatthewLayton: See comments on `StateRef`.

ledger/ledger-utxo/src/main/kotlin/net/corda/v5/ledger/utxo/Attachment.kt
```diff
@@ -0,0 +1,65 @@
...
+ * @property size The size of the attachment in bytes.
+ * @property signatories The keys that have been used to sign the attachment, or empty if the attachment does not require signing.
+ */
+@CordaSerializable
+@DoNotImplement
+interface Attachment {
+
+    val id: SecureHash
+    val size: Int
+    val signatories: List<PublicKey>
```
MrMatthewLayton, line 34: See comments on `ContractState`.

ledger/ledger-utxo/src/main/kotlin/net/corda/v5/ledger/utxo/StateAndRef.kt
```diff
@@ -0,0 +1,17 @@
...
+
+/**
+ * Represents a composition of a [TransactionState] and a [StateRef].
+ *
+ * @property state The [TransactionState] component of the current [StateAndRef].
+ * @property ref The [StateRef] component of the current [StateAndRef].
+ */
+@DoNotImplement
+@CordaSerializable
+interface StateAndRef<out T : ContractState> {
```
MrMatthewLayton, line 14: See comments on `StateRef`.

ledger/ledger-utxo/src/main/kotlin/net/corda/v5/ledger/utxo/TimeWindow.kt
```diff
@@ -0,0 +1,30 @@
...
+ * There is no such thing as "exact" time in distributed systems, due to the underlying physics involved, and other
+ * issues such as network latency. A time window represents an approximation of an instant with a margin of tolerance,
+ * and may be fully bounded, or unbounded towards negative or positive infinity.
+ *
+ * @property from The boundary at which the time window begins, or null if the time window is unbounded towards negative infinity.
+ * @property until The boundary at which the time window ends, or null if the time window is unbounded towards positive infinity.
+ */
+@DoNotImplement
+@CordaSerializable
+interface TimeWindow {
```
MrMatthewLayton, line 19: See comments on `StateRef`.

ledger/ledger-utxo/src/main/kotlin/net/corda/v5/ledger/utxo/TransactionState.kt
```diff
@@ -0,0 +1,25 @@
...
+ * @param T The underlying type of the [ContractState] instance.
+ * @property contractState The [ContractState] of the current [TransactionState] instance.
+ * @property contractStateType The [ContractState] type of the current [TransactionState] instance.
+ * @property contractType The [Contract] type of the current [TransactionState] instance.
+ * @property notary The notary of the current [TransactionState] instance.
+ * @property encumbrance The encumbrance of the current [TransactionState] instance.
+ */
+@DoNotImplement
+@CordaSerializable
+interface TransactionState<out T : ContractState> {
```
MrMatthewLayton, line 19: See comments on `StateRef`.

ledger/ledger-utxo/src/main/kotlin/net/corda/v5/ledger/utxo/transaction/UtxoSignedTransaction.kt
```diff
@@ -0,0 +1,51 @@
...
+     * @return Returns a new [UtxoSignedTransaction] containing the applied signature.
+     */
+    fun addSignatures(vararg signatures: DigitalSignatureAndMetadata): UtxoSignedTransaction
+
+    /**
+     * Gets the missing signatories from the current [UtxoSignedTransaction].
+     *
+     * @return Returns a [List] of [PublicKey] representing the missing signatories from the current [UtxoSignedTransaction].
+     */
+    fun getMissingSignatories(): List<PublicKey>
```
MrMatthewLayton, line 43: See comments on `ContractState`.

MrMatthewLayton (commented):
I have made a couple of amendments to this branch in order to fix the failing tests.

ledger/ledger-utxo/src/main/kotlin/net/corda/v5/ledger/utxo/TransactionState.kt
MrMatthewLayton: @mnesbit You mentioned that we needed to know the specific type; presumably because of type information loss at serialization, and we need to cast it back?

ledger/ledger-utxo/src/main/kotlin/net/corda/v5/ledger/utxo/StateRef.kt
blsemo: Most objects on the API never need to be created by users directly, and can therefore interfaces where the implementation is returned from API function calls. Some objects need factory methods as there initialisation is non-trivial, but users need to create them. And there are *some* objects that users do need to create and they can be plain POJOs.
Specifically, a `StateRef` is your key to get more information about a state that you might construct yourself from a transaction id and an index. If it's a interface, we need to provide a factory method or an overload of the `resolve` function that takes a transaction id and an index.

ledger/ledger-utxo/src/main/kotlin/net/corda/v5/ledger/utxo/TimeWindow.kt
blsemo: Probably not.

ledger/ledger-utxo/src/main/kotlin/net/corda/v5/ledger/utxo/UtxoLedgerService.kt
lankydan: Eventually this service is going to be huge if everything gets put onto it. So splitting things out which do not make up the primary workflow of the ledger sounds like a good idea to me. E.g. retrieving states or transactions can be done on other services. Whereas the primary workflow of, creating a tx builder, signing it, collecting signatures and finality should all be grouped together.

ledger/ledger-utxo/src/main/kotlin/net/corda/v5/ledger/utxo/transaction/UtxoTransactionBuilder.kt
lankydan: Ye I think `sign` itself sounds right.

ledger/ledger-common/src/main/kotlin/net/corda/v5/ledger/common/Party.kt
MrMatthewLayton: Candidate for a data class.

ledger/ledger-utxo/src/main/kotlin/net/corda/v5/ledger/utxo/ContractState.kt
MrMatthewLayton: Issue relating to set serialization - lists are safer.

ledger/ledger-utxo/src/main/kotlin/net/corda/v5/ledger/utxo/transaction/UtxoTransactionBuilder.kt
```diff
@@ -0,0 +1,152 @@
...
+    fun addOutputState(contractState: ContractState): UtxoTransactionBuilder
+
+    /**
+     * Adds an output state to the current [UtxoTransactionBuilder].
+     *
+     * @param contractState The [ContractState] to add to the current [UtxoTransactionBuilder].
+     * @param encumbrance The index of an associated, encumbered state, or null if no encumbrance applies to the associated transaction state.
+     * @return Returns a [UtxoTransactionBuilder] including the additional output state.
+     */
+    fun addOutputState(contractState: ContractState, encumbrance: Int?): UtxoTransactionBuilder
```
MrMatthewLayton, line 106: We need to consider alternative mechanisms for encumbrance, but not in this PR.

ledger/ledger-utxo/src/main/kotlin/net/corda/v5/ledger/utxo/StateRef.kt
mnesbit: We need a clear toString that can reversibly be parsed back, because people are doing this quite a lot. I would suggest delegating to JsonMarshallingService, but not sure if notary would handle that.

ledger/ledger-utxo/src/main/kotlin/net/corda/v5/ledger/utxo/TransactionState.kt
mnesbit: To prove that you have revealed all cash output states to an oracle for instance. The more important use for this data is in the under the hood info component group, which we have to use to handle validation and deserialization

ledger/ledger-utxo/src/main/kotlin/net/corda/v5/ledger/utxo/transaction/UtxoTransactionBuilder.kt
mnesbit: So long as people understand that this is a partially signed transaction, not a fully complete one

ledger/ledger-utxo/src/main/kotlin/net/corda/v5/ledger/utxo/transaction/UtxoTransactionBuilder.kt
```diff
@@ -0,0 +1,152 @@
...
+     *
+     * @param signatories The signatories expected to sign the current transaction.
+     * @return Returns a [UtxoSignedTransaction] with signatures for the specified signatory keys.
+     */
+    fun sign(vararg signatories: PublicKey): UtxoSignedTransaction
+
+    /**
+     * Verifies the current transaction.
+     */
+    fun verify()
```
mnesbit: we can remove, because sign will verify and then sign

MrMatthewLayton: nice!

mnesbit (approved):
OK

ronanbrowne (approved):
LGTM (Build)

PR #638, opened 2022-10-07
CORE-1261 Replaced the cipher suite by the one from the PoC

alexey-kadyrov-r3:
Checking whenever it can be replaced by the PoC;

See https://github.com/corda/corda-runtime-os/tree/akadyrov/POC-cipher_suite/components/crypto/crypto-poc

More comprehensive version than https://github.com/corda/corda-runtime-os/pull/2276


vkolomeyko:
Unlikely to be merged.

PR #639, opened 2022-10-10
CORE-6873: Add a hook for JsonSerializer and JsonDeserializer classes to be passed to the marshaller

fowlerrr:
These will be the custom CorDapp serializers that can be declared for users to encode their objects into Json.


simon-johnson-r3 (dismissed):

JamesHR3 (commented):
Do you need the API revision change? It might compile without it.

fowlerrr:
It failed the CI build.  Though it looks like someone else has already got there

simon-johnson-r3 (approved):

PR #642, opened 2022-10-10
CORE-4442 Update HostedIdentityEntry so we can optionally add a session certificate

williamvigorr3:
Add the session certificate to the encrypted extensions of both `Initiator Handshake` and `Responder Handshake` (see https://github.com/corda/platform-eng-design/blob/299b53fcd634cd9dcacea9cc0504121d5fabe28d/core/corda-5/corda-5.0/p2p-customer-facing-doc/doc.md#handshake).  Add a revocation option to the LinkManager config. So the revocation check can be controlled in the Link Manager in the same way as the Gateway.  Add the session certificates to the locally hosted identity.


dimosr (dismissed):
Looks ok

yift-r3 (dismissed):

dimosr (approved):

PR #648, opened 2022-10-13
CORE-6935 Expose `ledger-common` as api of `ledger-consensual`

lankydan:


blsemo (approved):
LGTM - should be mergeable without requiring a merge lock.

ronanbrowne (approved):
LGTM

PR #650, opened 2022-10-17
CORE-7071: Constants to be used in RBAC permission strings

vkolomeyko:
This is a non-breaking change which introduces few string constants.

The change is used by: https://github.com/corda/corda-runtime-os/pull/2329


ronanbrowne (approved):
LGTM (build)

ronanbrowne (changes requested):
actually sorry on second look, i think your build issue relates to the fact you have not added this new module to https://github.com/corda/corda-api/blob/vkolomeyko/CORE-7071-rbac-const/corda-api/build.gradle 

therefore in runtime it does not know how to resolve it when you just declare "implementation 'net.corda:corda-rbac-schema'" without the version 

as the version comes from this corda-api/corda-api module which is a [java-platform 
](https://docs.gradle.org/current/userguide/java_platform_plugin.html)

ronanbrowne (approved):
LGTM (build)

PR #651, opened 2022-10-17
CORE-7001: Define notary lookup service

yift-r3:
Runtime PR in https://github.com/corda/corda-runtime-os/pull/2348

### DIFF
--- a/application/src/main/kotlin/net/corda/v5/application/membership/NotaryLookup.kt
+++ b/application/src/main/kotlin/net/corda/v5/application/membership/NotaryLookup.kt
@@ -1,0 +1,69 @@
+package net.corda.v5.application.membership
+
+import net.corda.v5.base.annotations.DoNotImplement
+import net.corda.v5.base.annotations.Suspendable
+import net.corda.v5.base.types.MemberX500Name
+import net.corda.v5.membership.NotaryInfo
+import java.security.PublicKey
+
+/**
+ * [NotaryLookup] allows flows to retrieve the [NotaryInfo] in the network.
+ *
+ * The platform will provide an instance of [NotaryLookup] to flows via property injection.
+ */
+@DoNotImplement
+interface NotaryLookup {
+
+    /**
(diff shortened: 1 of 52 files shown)


application/src/main/kotlin/net/corda/v5/application/membership/NotaryLookup.kt
```diff
@@ -0,0 +1,69 @@
...
+     * ```
+     * list.joinToString("|")
+     * list.map { "[$it]" }.toString()
+     * ```
+     * This allows to distinguish between each element of a map, as
+     * [MemberX500Name.toString] returns comma separated values of [MemberX500Name].
+     *
+     * @return A list of the network's Notary services.
+     */
+    @Suspendable
```
charlieR3: I'm curious why is this required?

relyafi (commented):
LGTM

application/src/main/kotlin/net/corda/v5/application/membership/NotaryLookup.kt
charlieR3: It seem like this is required on all functions based on the conversations in the runtime os PR

charlieR3: Added

charlieR3:
Discusses with Matthew at today's standup. This may now be a ledger concern that we do not need to work on. Waiting until Christian is back from PTO on Thursday to confirm.

charlieR3:
> Discusses with Matthew at today's standup. This may now be a ledger concern that we do not need to work on. Waiting until Christian is back from PTO on Thursday to confirm.

Agreed that we would continue this work. `NotaryInfo` and `NotaryLookup` can be moved to the ledger common module. This post seems relevant https://r3holdco.slack.com/archives/C03A3TQMBHR/p1667380816659219

charlieR3 (dismissed):

charlieR3 (approved):

PR #654, opened 2022-10-18
CORE-7169 Use public keys instead of parties

vlajos:
Use PublicKeys instead of Parties in ConsensualStates

Runtime-os: https://github.com/corda/corda-runtime-os/pull/2343


mnesbit (approved):
OK

blsemo (approved):
LGTM

PR #655, opened 2022-10-18
CORE-5375: Add changeset_id to changelog and changelog audit

benmcmahon100:
Change UUIDs now serve as a on upload record of uniqueness between change sets. Two identical CPIs will have the same migrations, but different change UUIDS. This lets us tag migrations we run with the changeset UUID, which in turn makes it possible to associate a database state, with the changelogs that created that state in the audit table.

`corda-runtime-os` PR: https://github.com/corda/corda-runtime-os/pull/2371

### DIFF
--- a/data/db-schema/src/main/resources/net/corda/db/schema/config/migration/cpx-creation-v1.0.xml
+++ b/data/db-schema/src/main/resources/net/corda/db/schema/config/migration/cpx-creation-v1.0.xml
@@ -154,7 +154,7 @@
             <column name="content" type="TEXT">
                 <constraints nullable="false"/>
             </column>
-            <column name="change_uuid" type="VARCHAR(255)">
+            <column name="change_uuid" type="UUID">
                 <constraints nullable="false"/>
             </column>
             <!-- audit -->
@@ -198,7 +198,7 @@
             <column name="content" type="TEXT">
                 <constraints nullable="false"/>
             </column>
-            <column name="change_uuid" type="VARCHAR(255)">
+            <column name="change_uuid" type="UUID">
                 <constraints nullable="false"/>
             </column>
             <column name="is_deleted" type="BOOLEAN">
--- a/data/db-schema/src/main/resources/net/corda/db/schema/config/migration/cpx-creation-v1.0.xml
+++ b/data/db-schema/src/main/resources/net/corda/db/schema/config/migration/cpx-creation-v1.0.xml
@@ -201,6 +201,9 @@
             <column name="changeset_id" type="UUID">
                 <constraints nullable="false"/>
             </column>
+            <column name="entity_version" type="INT">
+                <constraints nullable="false"/>
+            </column>
             <column name="is_deleted" type="BOOLEAN">
                 <constraints nullable="false"/>
             </column>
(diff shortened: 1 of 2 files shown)


data/db-schema/src/main/resources/net/corda/db/schema/config/migration/cpx-creation-v1.0.xml
```diff
@@ -195,7 +198,7 @@
             <column name="content" type="TEXT">
                 <constraints nullable="false"/>
             </column>
-            <column name="entity_version" type="INT">
```
vkolomeyko, line 198: @benmcmahon100 - could you please put a description on this PR why this is necessary?
(I do not think Jira description would have any information on that)

benmcmahon100: 👍 Just working on that now

vkolomeyko:
We probably need `corda-runtime-os` PR (different to https://github.com/corda/corda-runtime-os/pull/2226) which reflects this change on the existing codebase.

data/db-schema/src/main/resources/net/corda/db/schema/config/migration/cpx-creation-v1.0.xml
```diff
@@ -154,6 +154,9 @@
             <column name="content" type="TEXT">
                 <constraints nullable="false"/>
             </column>
+            <column name="change_uuid" type="VARCHAR(255)">
```
driessamyn: The type can be `UUID` if this is a UUID.

But do we really need this?
Assuming this will change for each row, can we not use the PK?

`change_uuid` doesn't really explain what this is used for.

benmcmahon100: It won't be for each row no. It's for each set of migrations

driessamyn: Gotcha.
So it's on the CPI level?
Can we use the CPI ID in that case?

benmcmahon100: Also I didn't think UUID was a supported type across all DBMS so I suck with stringifying it

benmcmahon100: Can't use CPI ID as upon force upload you'll reuse the CPIID, but won't know which version of the CPI is related to which stet of migrations (they'll all look the same)

driessamyn: We use UUID elsewhere.

benmcmahon100: I can change it no problem, just seemed upon googling that it'd break on different DBMS and was worried

driessamyn: > Can't use CPI ID as upon force upload you'll reuse the CPIID, but won't know which version of the CPI is related to which stet of migrations (they'll all look the same)

CPI ID + entity version maybe?
It just feels a bit like we're introducing a surrogate key, which may be justifiable, but then we should use it as a surrogate key. If we use the composite key elsewhere, we should use that I think.

benmcmahon100: Entity version won't work either in the case of a new migration or a complete change. The smoke tests in the other PR has this issues as it changes from the dogs migration to the fish migration. It used to work based on the CPI Id + entityVersion until this bug came up

benmcmahon100: Just to follow up on this - agreed on a call to call it changeset_id instead for clarity

vkolomeyko (approved):
LGTM

PR #656, opened 2022-10-19
CORE-6130: Remove cluster certificate table to the config schema

yift-r3:
Runtime PR in https://github.com/corda/corda-runtime-os/pull/2353

This PR:
1. Adds the cluster-level certificates table to the config database.
2. Remove the certificate table.
3. Replace the `tenantID` with `usage` in the certificate RPC request.
4. Define CertificateType enum.


charlieR3 (dismissed):

data/avro-schema/src/main/resources/avro/net/corda/data/certificates/CertificateType.avsc
```diff
@@ -0,0 +1,11 @@
+{
+  "type": "enum",
+  "name": "CertificateType",
+  "namespace": "net.corda.data.certificates",
+  "doc": "A cluster level certificate type.",
```
dimosr: Is this going to be used generally or just for cluster-level certificates?

data/avro-schema/src/main/resources/avro/net/corda/data/certificates/rpc/request/CertificateRpcRequest.avsc
```diff
@@ -5,9 +5,9 @@
   "doc": "RPC request regarding certificates.",
   "fields": [
     {
-      "name": "tenantId",
-      "doc": "'p2p', 'rpc-api', or holding identity identity ID.",
-      "type": "string"
+      "name": "usage",
+      "doc": "The certificate usage. Either CertificateType or holding identity ID.",
+      "type": ["net.corda.data.certificates.CertificateType", "string"]
```
dimosr: Seems like we have two different pieces of information (the certificate type and whether the corresponding key is a cluster-level or vnode-level one) mushed up into one field/type here? Wouldn't it be cleaner to have separate fields to model each thing?

data/db-schema/src/main/resources/net/corda/db/schema/config/migration/config-creation-v1.0.xml
```diff
@@ -311,5 +311,19 @@
                                  constraintName="FK_vnode_instance_cpi"
                                  baseTableSchemaName="${schema.name}"
                                  referencedTableSchemaName="${schema.name}"/>
+        <createTable tableName="cluster_certificate" schemaName="${schema.name}">
+            <column name="type" type="VARCHAR(255)">
```
dimosr: related to my comment above.. does this mean this value can either contain a certificate type or a holding identity value? Should we create separate columns for each piece of information?

data/avro-schema/src/main/resources/avro/net/corda/data/certificates/CertificateType.avsc
yift-r3: Just for cluster-level certificates.

data/db-schema/src/main/resources/net/corda/db/schema/config/migration/config-creation-v1.0.xml
dimosr: Also, if I understand correctly in case it's a vnode-level key/cert, we expect the user to specify the holding id short hash here and does this mean that we won't store the certificate type/usage in the DB at all?

data/avro-schema/src/main/resources/avro/net/corda/data/certificates/rpc/request/CertificateRpcRequest.avsc
yift-r3: Certificate can either be a cluster level certificate (with a type) or a virtual node certificate.

data/db-schema/src/main/resources/net/corda/db/schema/config/migration/config-creation-v1.0.xml
yift-r3: This is just for the cluster-level certificate (so, only type).  The virtual node certificates are in another table in the virtual node database.

dimosr: ah right, forgot about the other table. Thanks

data/avro-schema/src/main/resources/avro/net/corda/data/certificates/rpc/request/CertificateRpcRequest.avsc
dimosr: but, isn't the concept of a certificate type useful generally regardless of whether it's a cluster-evel or vnode-level one i.e. shouldn't we keep hold of that information for vnode-level certificates too?

dimosr: (what I was thinking is having a separate field that indicates the type of the certificate and a separate field that indicates whether it's a cluster-level or a vnode-level one)

data/avro-schema/src/main/resources/avro/net/corda/data/certificates/CertificateType.avsc
```diff
@@ -0,0 +1,11 @@
...
+  "symbols": [
+    "P2P",
```
dimosr: what do we want this type to reflect exactly? Is it how the certificate is used or from which part of the system it's used?

Asking because if it's the former, should we have `P2P_SESSION` and `P2P_TLS` here? or something along those lines

dimosr (dismissed):

charlieR3 (dismissed):
LGTM

dimosr (dismissed):

charlieR3 (dismissed):

charlieR3 (approved):

PR #657, opened 2022-10-20
CORE-6768 - Add reason behind not submitted status

nikinagy:
https://r3-cev.atlassian.net/browse/CORE-6768

### DIFF
--- a/data/avro-schema/src/main/resources/avro/net/corda/data/membership/rpc/response/RegistrationRpcResponse.avsc
+++ b/data/avro-schema/src/main/resources/avro/net/corda/data/membership/rpc/response/RegistrationRpcResponse.avsc
@@ -27,6 +27,11 @@
       }
     },
     {
+      "name": "reason",
+      "doc": "Reason why the request has not submitted status. Null, if the status is submitted.",
+      "type": ["null", "string"]
+    },
+    {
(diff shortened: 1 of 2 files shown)


charlieR3 (dismissed):

data/avro-schema/src/main/resources/avro/net/corda/data/membership/rpc/response/RegistrationRpcResponse.avsc
```diff
@@ -26,6 +26,11 @@
         "symbols": ["SUBMITTED", "NOT_SUBMITTED"]
       }
     },
+    {
+      "name": "reason",
+      "doc": "Reason why the request has not submitted status. Null, if the status is submitted.",
+      "type": ["null", "string"]
```
charlieR3, line 32: I approved this before looking at the runtime-os PR. I see there is a reason given for the success case so does this still need to be nullable?

nikinagy: Yes, because this is created on the service side sent over on RPC. We are defaulting on the client side if this was null.

charlieR3 (approved):

PR #660, opened 2022-10-24
CORE-4733 Initial API for the notary plugin architecture

nkovacsx:
## Overview

This PR adds the required interfaces for plugin creation. Some of these interfaces will be implemented in the corresponding runtime-os PR. However, some of the interfaces (provider, factory) will not be implemented for now because we don't have the architecture to load/select plugins for now.


ledger/notary-plugin/src/main/java/net/corda/v5/ledger/notary/pluggable/package-info.java
```diff
@@ -0,0 +1,4 @@
+@Export
+package net.corda.v5.ledger.notary.pluggable;
```
relyafi:
```suggestion
package net.corda.v5.ledger.notary.plugin;
```

Consistency with module name; notaries are pluggable by default in C5, so I assume what we're really saying is that this module / package is providing the plugin core / infrastructure? With that in mind I'd also be open to the module / package referring to `plugin-core` or the like if we thought that made sense.

(Comment applies to all files in this package)

ledger/notary-plugin/src/main/kotlin/net/corda/v5/ledger/notary/pluggable/NotarisationPayload.kt
```diff
@@ -0,0 +1,15 @@
+package net.corda.v5.ledger.notary.pluggable
+
+/**
+ * Container for the transaction and notarisation request signature. This is the payload that gets sent by a client to
+ * a notary service for committing the input states of the [transaction].
+ *
+ * This interface uses Any as for the [transaction] type. The [transaction] object will be validated as a concrete type
+ * (for example [net.corda.v5.ledger.utxo.transaction.UtxoSignedTransaction]) when a class inherits from this interface
+ * with a specified [validTypes] list.
```
relyafi:
```suggestion
 * This interface uses Any for the [transaction] type. The [transaction] object will be validated as a concrete type
 * (for example [UtxoSignedTransaction][net.corda.v5.ledger.utxo.transaction.UtxoSignedTransaction]) against the specified [validTypes] list.
 *
```

- I feel this is clearer that we are validating the transaction against the type, rather than the inheriting class.
- Doc will display `UtxoSignedTransaction` (neater) but still have a valid link using the fully qualified name

ledger/notary-plugin/build.gradle
```diff
@@ -0,0 +1,23 @@
+plugins {
+    id 'corda-api.common-library'
+    id 'corda.common-publishing'
+    id 'org.jetbrains.kotlin.jvm'
+}
+
+description 'Corda Ledger Common'
```
relyafi: Looks like a copy / paste, and should be updated for this module.

ledger/notary-plugin/src/main/kotlin/net/corda/v5/ledger/notary/pluggable/NotarisationRequest.kt
```diff
@@ -0,0 +1,28 @@
...
+ * Reference inputs states are not included as a separate property in the [NotarisationRequest] as they are not
+ * consumed.
+ *
+ * In case there is a need to prove that a party spent a particular state, the notary will:
+ * 1) Locate the consuming transaction id in the index, along with all other states consumed in the same transaction.
+ * 2) Build a [NotarisationRequest].
+ * 3) Locate the [NotarisationRequestSignature] for the transaction id. The signature will contain the signing public key.
+ * 4) Demonstrate the signature verifies against the serialized request. The provided states are always sorted internally,
+ *    to ensure the serialization does not get affected by the order.
+ */
```
relyafi: This is a really descriptive comment, but is going into a lot of (assumed) implementation detail - who's to say what an implementation will use, or what checks it will do?

These comments may still be valid in the context of an implementation class, but at the API level I would review this and make sure that this only refers to things that an API needs to worry about.

ledger/notary-plugin/src/main/kotlin/net/corda/v5/ledger/notary/pluggable/NotarisationResponse.kt
```diff
@@ -0,0 +1,13 @@
+package net.corda.v5.ledger.notary.pluggable
+
+import net.corda.v5.application.crypto.DigitalSignatureAndMetadata
+import net.corda.v5.application.uniqueness.model.UniquenessCheckError
+
+/**
+ * An interface that represents the payload returned by the notary server flow to the client flow. Contains the
+ * signatures from the server side and the error, in case there's any.
```
relyafi:
```suggestion
 * signatures from the server side and the error, if applicable.
```

ledger/notary-plugin/src/main/kotlin/net/corda/v5/ledger/notary/pluggable/PluggableNotaryClientFlow.kt
```diff
@@ -0,0 +1,24 @@
...
+ * Further, the [call] method must specify the [net.corda.v5.base.annotations.Suspendable] annotation.
+ *
+ * The server side will not have an interface like the client, it will only implement the
+ * [net.corda.v5.application.flows.ResponderFlow] interface.
+ *
+ * For an example client and server implementation please refer to the non-validating notary plugin under
+ * the `notary-plugins/notary-plugin-non-validating` module in the `corda-runtime-os` repository,
+ * or the quick starter guide.
+ */
+interface PluggableNotaryClientFlow : SubFlow<List<DigitalSignatureAndMetadata>>
```
relyafi: Observation - I think it makes sense for this set of interfaces to refer to pluggable (as they currently do), because they are directly used by implementing plugins, as opposed to being plugin "core" / "infrastructure"

ledger/notary-plugin/src/main/kotlin/net/corda/v5/ledger/notary/pluggable/PluggableNotaryClientFlow.kt
```diff
@@ -0,0 +1,24 @@
...
+ * This interface has a single method called [call] which contains the main logic for the client.
+ * The client flow must implement this interface in order for it to be instantiated in the provider.
+ * If the client flow doesn't implement this interface the plugin selection will not work.
+ * To learn more about plugin selection and types see [PluggableNotaryType].
+ *
+ * Implementations must specify the [net.corda.v5.application.flows.InitiatingFlow] annotation.
+ * Further, the [call] method must specify the [net.corda.v5.base.annotations.Suspendable] annotation.
+ *
+ * The server side will not have an interface like the client, it will only implement the
+ * [net.corda.v5.application.flows.ResponderFlow] interface.
```
relyafi:
```suggestion
 * Implementations must specify the [InitiatingFlow][net.corda.v5.application.flows.InitiatingFlow] annotation.
 * Further, the [call] method must specify the [Suspendable][net.corda.v5.base.annotations.Suspendable] annotation.
 *
 * The server side will not have an interface like the client, it will only implement the
 * [ResponderFlow][net.corda.v5.application.flows.ResponderFlow] interface.
```

ledger/notary-plugin/src/main/kotlin/net/corda/v5/ledger/notary/pluggable/PluggableNotaryClientFlow.kt
```diff
@@ -0,0 +1,24 @@
...
+ *
+ * For an example client and server implementation please refer to the non-validating notary plugin under
+ * the `notary-plugins/notary-plugin-non-validating` module in the `corda-runtime-os` repository,
+ * or the quick starter guide.
```
relyafi:
```suggestion
 * or the quick start guide.
```

ledger/notary-plugin/src/main/kotlin/net/corda/v5/ledger/notary/pluggable/PluggableNotaryClientFlowProvider.kt
```diff
@@ -0,0 +1,18 @@
...
+
+/**
+ * This interface is used to instantiate the client-side logic (flow) of the plugin.
+ * This is needed because the plugins will be installed on node startup.
+ *
+ * Implementers of this interface must have the [PluggableNotaryType] annotation,
+ * otherwise an exception will be thrown on startup.
+ *
+ * The client-side of the plugin must have the same constructor parameters as the
+ * [create] function (notary, stx, context).
```
relyafi: I'm not sure where `context` comes in as this isn't actually in the `create` function signature, but isn't this comment redundant since a class must implement the `create` function as defined in the interface in order to compile?

relyafi (changes requested):
Broadly looks good, a number of minor comments, mostly relating to KDocs

ledger/notary-plugin/src/main/java/net/corda/v5/ledger/notary/pluggable/package-info.java
nkovacsx: All these interfaces in this package are needed to *create a plugin*, not related to an actual plugin. I'd argue it's better to be in `pluggable` since they are required to create a plugin thus "pluggable". 

A `plugin-core` module name would make sense to me.

That's just my opinion though, if you still want to stick with `plugin` I can modify it.

ledger/notary-plugin/src/main/kotlin/net/corda/v5/ledger/notary/pluggable/PluggableNotaryClientFlowProvider.kt
nkovacsx: `context` used to be in DP1 but it's not a thing here, so will remove this.

nkovacsx: Yeah good point, will remove this comment altogether.

ledger/notary-plugin/src/main/kotlin/net/corda/v5/ledger/notary/pluggable/PluggableNotaryClientFlow.kt
nkovacsx: I'm not familiar with this syntax in KDocs, what will be the output of this double square bracket structure `[][]`?

ledger/notary-plugin/src/main/java/net/corda/v5/ledger/notary/pluggable/package-info.java
nkovacsx: Discussed this offline, will stick to having `plugin` as the root package and `api`, `core` as sub-packages.

ledger/notary-plugin/src/main/kotlin/net/corda/v5/ledger/notary/pluggable/NotarisationRequest.kt
nkovacsx: Good point, I'll only keep the first paragraph, I think that's enough description for this interface.

relyafi (dismissed):
LGTM

ledger/notary-plugin/src/main/java/net/corda/v5/ledger/notary/plugin/package-info.java
```diff
@@ -0,0 +1,4 @@
+@Export
+package net.corda.v5.ledger.notary.plugin;
+
+import org.osgi.annotation.bundle.Export;
```
relyafi: Is this file necessary, given there's no source files at this level (only under api / core)?

nkovacsx: Good point!

ronanbrowne (dismissed):
LGTM ( only build changes reviewed ) - corda-runtime-os side will need to update it's composite build [logic](https://github.com/corda/corda-runtime-os/blob/release/os/5.0/settings.gradle#L414) in the runtime-os settings.gradle also for this new module

relyafi (changes requested):
As discussed verbally, I think this needs reviewing to ensure that we are following the rules specified in the [design](https://github.com/corda/platform-eng-design/blob/646e922751c592a0e995eaeea56554de014a6d6f/core/corda-5/corda-5.0/notary/pluggable-notary-placement-packaging/index.md), i.e. that `corda-api` only contains things necessary for a developer to create a plugin protocol. A good chunk of these interfaces / classes as they stand appear to be either things that are commonly useful to plugin implementations or are only used by our own internal implementations. Both these categories of things should be in `corda-runtime-os`

ledger/notary-plugin/src/main/kotlin/net/corda/v5/ledger/notary/plugin/core/NotaryError.kt
```diff
@@ -0,0 +1,3 @@
+package net.corda.v5.ledger.notary.plugin.core
+
+interface NotaryError
```
relyafi, line 12: This should probably have a kdoc comment to explain what it is / how it is used, given that someone looking just at the `corda-api` repo will have no further context beyond this file.

relyafi (commented):
Looks reassuringly minimal now. One minor comment regarding docs, otherwise ready to approve

relyafi (dismissed):
LGTM

relyafi (approved):

ronanbrowne (approved):
LGTM (build)

PR #661, opened 2022-10-25
CORE-6083 - Add URL path as a config option for gateway

dimosr:
Add configuration option to adjust the URL path the gateway server is listening to for requests.

To keep things simple, I've added a validation pattern that allows relatively simple paths for now (i.e. alphanumerics separated by slashes). The root path was selected as default, so that it works out of the box with URLs that do not specify any custom path.

Corresponding runtime PR: https://github.com/corda/corda-runtime-os/pull/2378

### DIFF
--- a/data/config-schema/src/main/resources/net/corda/schema/configuration/p2p.gateway/1.0/corda.p2p.gateway.json
+++ b/data/config-schema/src/main/resources/net/corda/schema/configuration/p2p.gateway/1.0/corda.p2p.gateway.json
@@ -21,7 +21,8 @@
     "urlPath": {
       "description": "The url path the gateway server will be listening to.",
       "type": "string",
-      "default": "/"
+      "default": "/",
+      "pattern": "^\\/[\\/.a-zA-Z0-9-]*$"
     },
     "sslConfig": {
       "description": "Used for one-way TLS between Gateways.",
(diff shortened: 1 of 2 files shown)


data/config-schema/src/main/resources/net/corda/schema/configuration/p2p.gateway/1.0/corda.p2p.gateway.json
```diff
@@ -18,6 +18,12 @@
       "minimum": 0,
       "maximum": 65535
     },
+    "urlPath": {
```
williamvigorr3, line 21: We could consider combining this with the `hostAddress` and `hostPort` as a URI.

data/config-schema/src/main/resources/net/corda/schema/configuration/p2p.gateway/1.0/corda.p2p.gateway.json
```diff
@@ -18,6 +18,12 @@
...
+      "description": "The url path the gateway server will be listening to.",
+      "type": "string",
+      "default": "/",
+      "pattern": "^\\/[\\/.a-zA-Z0-9-]*$"
```
williamvigorr3, line 25: ~Did you get this pattern from anywhere specific (like an RFC)?~ I should have read the PR description.

data/config-schema/src/main/resources/net/corda/schema/configuration/p2p.gateway/1.0/corda.p2p.gateway.json
dimosr: [Discussed in the standup] Agreed to keep them as separate fields for now to prevent confusion between binding address/port and connection URL. And to make it a bit more easy for people to only adjust one part without having to specify a full URL. We can revisit later on depending on user feedback.

williamvigorr3 (dismissed):
LGTM

williamvigorr3 (approved):
LGTM

PR #662, opened 2022-10-26
CORE-7128 Sign operation for single key that picks placeholder default signature spec

kyriathar:
WIP

runtime-os PR: [#2387](https://github.com/corda/corda-runtime-os/pull/2387)

### DIFF
--- a/application/src/main/kotlin/net/corda/v5/application/crypto/SigningService.kt
+++ b/application/src/main/kotlin/net/corda/v5/application/crypto/SigningService.kt
@@ -10,7 +10,7 @@ import java.security.KeyPair
 import java.security.PrivateKey
 import java.security.PublicKey
 import java.util.*
-import net.corda.v5.crypto.DigitalSignatureWithSpec
+import net.corda.v5.crypto.DigitalSignatureWithSignatureSpec
 
 /**
  * Responsible for storing and using private keys to sign things. An implementation of this may, for example, call out
@@ -39,5 +39,5 @@ interface SigningService {
     fun sign(bytes: ByteArray, publicKey: PublicKey, signatureSpec: SignatureSpec): DigitalSignature.WithKey
 
     @Suspendable
-    fun sign(bytes: ByteArray, publicKey: PublicKey): DigitalSignatureWithSpec
+    fun sign(bytes: ByteArray, publicKey: PublicKey): DigitalSignatureWithSignatureSpec
 }
--- a/application/src/main/kotlin/net/corda/v5/application/crypto/SigningService.kt
+++ b/application/src/main/kotlin/net/corda/v5/application/crypto/SigningService.kt
@@ -10,7 +10,7 @@ import java.security.KeyPair
 import java.security.PrivateKey
 import java.security.PublicKey
 import java.util.*
-import net.corda.v5.crypto.DigitalSignatureWithSignatureSpec
+import net.corda.v5.crypto.DigitalSignatureWithSpec
 
 /**
  * Responsible for storing and using private keys to sign things. An implementation of this may, for example, call out
@@ -39,5 +39,5 @@ interface SigningService {
     fun sign(bytes: ByteArray, publicKey: PublicKey, signatureSpec: SignatureSpec): DigitalSignature.WithKey
 
     @Suspendable
-    fun sign(bytes: ByteArray, publicKey: PublicKey): DigitalSignatureWithSignatureSpec
+    fun sign(bytes: ByteArray, publicKey: PublicKey): DigitalSignatureWithSpec
 }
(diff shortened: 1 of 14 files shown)


cipher-suite/src/main/kotlin/net/corda/v5/cipher/suite/CipherSchemeMetadata.kt
```diff
@@ -82,13 +82,16 @@ interface CipherSchemeMetadata : KeyEncodingService, AlgorithmParameterSpecEncod
...
      * public key types as the digest is integral part of the signing/verification, e.g. if the [publicKey] is 'EdDSA'
-     * hen the [digest] is set to "EdDSA" by the platform's default implementation.
+     * then the [digest] is set to "EdDSA" by the platform's default implementation.
      *
      * @return [SignatureSpec] with the signatureName formatted like "SHA256withECDSA" if that can be inferred or
      * otherwise null.
      */
     fun inferSignatureSpec(publicKey: PublicKey, digest: DigestAlgorithmName): SignatureSpec?
 
+    // TODO kdoc
```
dickon: Please add the kdoc

crypto/src/main/kotlin/net/corda/v5/crypto/DigitalSignature.kt
```diff
@@ -36,4 +36,14 @@ open class DigitalSignature(
          */
         val context: Map<String, String>
     ) : DigitalSignature(bytes)
+
+    class WithId(
+        /**
+         * SHA-256 digest of the public key. TODO add reasoning for hashing public hey
```
dickon: To be added. Suggest "SHA-256 digest of the encoded public key. We propagate hash rather than public key itself since these values should only be used for comparison. The actual public keys need to be propagated by some other mechanism before the signature can be validated."

crypto/src/main/kotlin/net/corda/v5/crypto/DigitalSignature.kt
```diff
@@ -36,4 +36,14 @@ open class DigitalSignature(
...
+         */
+        val by: SecureHash,
+        bytes: ByteArray
+        // TODO do we need to include the context map?
```
dickon: Suggest we just delete this and the commented out context map for now.

dickon (changes requested):

crypto/src/main/kotlin/net/corda/v5/crypto/DigitalSignature.kt
kyriathar: Done

crypto/src/main/kotlin/net/corda/v5/crypto/DigitalSignature.kt
kyriathar: Added

cipher-suite/src/main/kotlin/net/corda/v5/cipher/suite/CipherSchemeMetadata.kt
kyriathar: Added with a more internal TODO that will be addressed with [CORE-7132](https://r3-cev.atlassian.net/browse/CORE-7132).

application/src/main/kotlin/net/corda/v5/application/crypto/SigningService.kt
```diff
@@ -36,4 +36,7 @@ interface SigningService {
      */
     @Suspendable
     fun sign(bytes: ByteArray, publicKey: PublicKey, signatureSpec: SignatureSpec): DigitalSignature.WithKey
+
+    @Suspendable
+    fun sign(bytes: ByteArray, publicKey: PublicKey): DigitalSignatureWithSpec
```
dickon, line 41: There was some discussion on this today. There  are some tradeoffs with this new operation:

For safe operation, the trouble is that the signed payload must contain the signature spec, but the caller doesn't know the signature spec until we get the result. The implementation _could_ just choose the signature spec, then append it to the bytes, then sign that. But then it's no longer true that we are signing the `bytes` argument, and the recipient wil need to split out the payload from the signature spec. 

Consider a caller that wants to send messages as JSON. It wouldn't be valid JSON to just append a serialised signtature spec to the end of the payload.

This also means that some kind of serialization - of the signature spec, and how we combined it with the bytes payload, must be done by the crypto code with no control from the caller.

Further, this means that the signature for a given keypiar is not going to match what other implementations would produce, when signing `bytes` itself. It means that no other implementation could exist without assuming the internal serialisation details of the crypto worker, e.g. appended AMQP for the signing spec.

dickon (changes requested):

PR #666, opened 2022-10-27
CORE-7186: Add regex constants for RBAC arguments

vkolomeyko:
Putting those in `corda-api` as they wll be needed at multiple modules in `corda-runtime-os` repo.

Corresponding `corda-runtime-os` PR: https://github.com/corda/corda-runtime-os/pull/2397

### DIFF
--- a/data/rbac-schema/src/main/kotlin/net/corda/rbac/schema/RbacKeys.kt
+++ b/data/rbac-schema/src/main/kotlin/net/corda/rbac/schema/RbacKeys.kt
@@ -15,4 +15,35 @@ object RbacKeys {
      * Separator for permission string which sits between prefix and the wildcard expression
      */
     const val PREFIX_SEPARATOR = ":"
+
+    /**
+     * Regular expressions to validate common data structures as part of RBAC checks
+     */
+    private const val UUID_CHARS = "[a-fA-F0-9]"
+    const val UUID_REGEX = "$UUID_CHARS{8}-$UUID_CHARS{4}-$UUID_CHARS{4}-$UUID_CHARS{4}-$UUID_CHARS{12}"
+
+    /**
+     * vNode short hash
+     */
+    const val VNODE_SHORT_HASH_REGEX = "$UUID_CHARS{12}"
+
+    /**
+     * RBAC user name regex
+     */
+    const val USER_REGEX = "[-._@a-zA-Z0-9]{3,255}"
+
+    // first.last@company.com is a valid username, however when encoded in the URL it will be shown as
+    // first.last%40company.com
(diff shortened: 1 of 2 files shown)


data/rbac-schema/src/main/kotlin/net/corda/rbac/schema/RbacKeys.kt
```diff
@@ -15,4 +15,35 @@ object RbacKeys {
...
+
+    /**
+     * vNode short hash
+     */
+    const val VNODE_SHORT_HASH_REGEX = "$UUID_CHARS{12}"
+
+    /**
+     * RBAC user name regex
+     */
+    const val USER_REGEX = "[-._@a-zA-Z0-9]{3,255}"
```
mnesbit, line 33: Regex's for valid user names are very difficult. There are lots of people with unusual characters in their name e.g. ' or French names with accents. I think we need to consider alternative escaping strategies here

data/rbac-schema/src/main/kotlin/net/corda/rbac/schema/RbacKeys.kt
```diff
@@ -15,4 +15,35 @@ object RbacKeys {
      * Separator for permission string which sits between prefix and the wildcard expression
      */
     const val PREFIX_SEPARATOR = ":"
+
+    /**
+     * Regular expressions to validate common data structures as part of RBAC checks
+     */
+    private const val UUID_CHARS = "[a-fA-F0-9]"
+    const val UUID_REGEX = "$UUID_CHARS{8}-$UUID_CHARS{4}-$UUID_CHARS{4}-$UUID_CHARS{4}-$UUID_CHARS{12}"
```
mnesbit: Use "^[a-fA-F0-9]{8}-[a-fA-F0-9]{4}-[1-5][a-fA-F0-9]{3}-[89aAbB][a-fA-F0-9]{3}-[a-fA-F0-9]{12}$" as per https://github.com/corda/corda-api/blob/release/os/5.0/data/membership-schema/src/main/resources/net/corda/schema/membership/group/policy/1.0/corda.group.policy.json#L22 as UUID actually has more structure than this

data/rbac-schema/src/main/kotlin/net/corda/rbac/schema/RbacKeys.kt
vkolomeyko: I have raised a follow-up Jira to that effect: https://r3-cev.atlassian.net/browse/CORE-7305

mnesbit (approved):
ok subject to follow up on user names

PR #667, opened 2022-10-28
CORE-7296 - Add schema and topic for group params

nikinagy:
https://r3-cev.atlassian.net/browse/CORE-7296

Added schema and topic for the reconciler writer.


data/topic-schema/src/main/kotlin/net/corda/schema/Schemas.kt
```diff
@@ -101,6 +101,8 @@ class Schemas {
      */
     class Membership {
         companion object {
+            const val GROUP_PARAMETERS_TOPIC = "membership.group.params"
```
charlieR3, line 104: The topic needs to be configured in Membership.yaml also.

data/avro-schema/src/main/resources/avro/net/corda/data/membership/GroupParameters.avsc
```diff
@@ -0,0 +1,18 @@
+{
+  "type": "record",
+  "name": "GroupParameters",
+  "namespace": "net.corda.data.membership",
+  "doc": "Avro representation of GroupParameters.",
+  "fields": [
+    {
+      "name": "viewOwner",
+      "doc": "For dynamic networks it is the short hash of member's holding identity or for static registration the name of the owning service.",
+      "type": ["string"]
```
charlieR3: I think we'll need something else here. 

For the static registration that Yash is looking at we'll need a service identifier and a group ID. For every other case we'll need a way to connect the record to a holding identity; i.e. holding identity hash, holding identity, or just name and group identifier. 

I'd like to change `viewOwner` to either:
1) Two new fields; group identifier and owner identifier (which would be service name or x500 name)
2) One new type called something like `GroupParametersOwner` which contained group identifier and owner identifier.

What I like about `GroupParametersOwner` is that it's clearly specifically for group parameters and it is clear that it is intentionally different to `HoldingIdentity`.

data/topic-schema/src/main/resources/net/corda/schema/Membership.yaml
```diff
@@ -1,4 +1,9 @@
 topics:
+  GroupParametersTopic:
+    name: membership.group.params
+    consumers:
+    producers:
+    config:
```
charlieR3, line 6: This should have the config to make it a compacted topic like the member list.

data/avro-schema/src/main/resources/avro/net/corda/data/membership/GroupParametersOwner.avsc
```diff
@@ -0,0 +1,18 @@
+{
+  "type": "record",
+  "name": "GroupParametersOwner",
```
charlieR3: Sorry, I'm going to be a pain and throw another suggestion into the mix.
While talking to Yash I realised that if we use a static string for the case of static, we'll need to copy that static string everywhere, but if it was a strong type we'll just need to check the class type. 

For example, if `viewOwner` was either `HoldingIdentity` or `Service` where `Service` was basically the same as `GroupParametersOwner` (ID and group ID), then in other areas of the code we can just check that the view owner is a holding identity and not have to be aware of the static service record. i.e. `if(viewOwner is HoldingIdentity) do something else do nothing`

yift-r3 (dismissed):

charlieR3 (commented):
Sorry, I left comments on this last week but just realised I didn't hit submit 😥

charlieR3 (approved):

PR #668, opened 2022-10-28
CORE-7304: Service to determine default signature spec

dickon:
Adds `SignatureSpecService` application service to provide users with default signature spec and compatible signature specs APIs for specified public key and (optionally) digest algorithm name.

complement runtime-os PR: [#2437](https://github.com/corda/corda-runtime-os/pull/2437)

### DIFF
--- a/application/src/main/kotlin/net/corda/v5/application/crypto/DefaultSignatureSpecService.kt
+++ b/application/src/main/kotlin/net/corda/v5/application/crypto/DefaultSignatureSpecService.kt
@@ -1,0 +1,31 @@
+package net.corda.v5.application.crypto
+
+import net.corda.v5.base.annotations.CordaSerializable
+import net.corda.v5.base.annotations.Suspendable
+import net.corda.v5.crypto.DigestAlgorithmName
+import net.corda.v5.crypto.SignatureSpec
+import java.security.PublicKey
+
+interface DefaultSignatureSpecService {
+    /**
+     * Work out a signature and hash algorithm to use for a given public key, given current security policies.
+     *
+     * @param publicKey the public key to be used for signing
+     *
+     * @return An appropriate [SignatureSpec], or null if nothing is available for the key type.
+     */
+    @Suspendable
+    fun suggest(publicKey: PublicKey): SignatureSpec?
+
+    /**
+     * Work out a signature algorithm given current security policies and a hash algorithm
+     *
+     * @param publicKey the public key to be used for signing
+     * @param digestAlgorithm the digest algorithm to use, e.g. []DigestAlgorithmName.SHA2_256]
+     *
+     * @return An appropriate [SignatureSpec], or null if nothing is available for the key type
+     */
+
+    @Suspendable
+    fun suggest(publicKey: PublicKey, digestAlgorithm: DigestAlgorithmName: DigestAlgorithmName): SignatureSpec?
+}
--- a/application/src/main/kotlin/net/corda/v5/application/crypto/SignatureSpecService.kt
+++ b/application/src/main/kotlin/net/corda/v5/application/crypto/SignatureSpecService.kt
@@ -1,12 +1,13 @@
 package net.corda.v5.application.crypto
 
-import net.corda.v5.base.annotations.CordaSerializable
+import net.corda.v5.base.annotations.DoNotImplement
 import net.corda.v5.base.annotations.Suspendable
 import net.corda.v5.crypto.DigestAlgorithmName
 import net.corda.v5.crypto.SignatureSpec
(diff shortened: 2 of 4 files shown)


application/src/main/kotlin/net/corda/v5/application/crypto/DefaultSignatureSpecService.kt
```diff
@@ -0,0 +1,26 @@
...
+import java.security.PublicKey
+
+interface DefaultSignatureSpecService {
+    /**
+     * Work out a signature and hash algorithm to use for a given public key, given current security policies.
+     *
+     * @param publicKey the public key to be used for signing
+     *
+     * @return An appropriate [SignatureSpec], or null if nothing is available for the key type.
+     */
```
mnesbit: Needs @Suspendable

application/src/main/kotlin/net/corda/v5/application/crypto/DefaultSignatureSpecService.kt
```diff
@@ -0,0 +1,26 @@
...
+    /**
+     * Work out a signature algorithm given current security policies and a hash algorithm
+     *
+     * @param publicKey the public key to be used for signing
+     * @param hash the hash algortithm e.g. "SHA256" - TODO: use an interface or enum here?
+     *
+     * @return An appropriate [SignatureSpec], or null if nothing is available for the key type
+     */
+
+    fun suggest(publicKey: PublicKey, hash: String): SignatureSpec?
```
mnesbit: Needs @Suspendable. hash parameters should be `DigestAlgorithmName`

application/src/main/kotlin/net/corda/v5/application/crypto/DefaultSignatureSpecService.kt
```diff
@@ -0,0 +1,26 @@
...
+}
```
mnesbit: We also need to add a `fun compatible(publicKey: PublicKey, hash: DigestAlgorithmName): List<SignatureSpec>` for either people who don't want to use our defaults, or for situations where older data used the default and we later changed that default

application/src/main/kotlin/net/corda/v5/application/crypto/DefaultSignatureSpecService.kt
dickon: Okay. It may be that we don't need to suspend the fiber to implement the API, but it may also be that we do since we want to go the crypto service for configuration.

application/src/main/kotlin/net/corda/v5/application/crypto/DefaultSignatureSpecService.kt
dickon: Done

application/src/main/kotlin/net/corda/v5/application/crypto/DefaultSignatureSpecService.kt
dickon: That could potentially be done in a later step. I'm also not sure if the output should be a `Set<SignatureSpec>`, of if keep it as a list is the order significant, e.g. most secure first? @kyriathar any thoughts?

dickon: `listCompatible` might be a better name?

kyriathar: I would think that `fun compatible` should only take `PublicKey` because with `PublicKey` and `DigestAlgorithm` that can lead to a single `SignatureSpec` ?

application/src/main/kotlin/net/corda/v5/application/crypto/SignatureSpecService.kt
```diff
@@ -0,0 +1,38 @@
...
+     *
+     * @param publicKey the public key to be used for signing
+     * @param digestAlgorithm the digest algorithm to use, e.g. []DigestAlgorithmName.SHA2_256]
+     *
+     * @return An appropriate [SignatureSpec], or null if nothing is available for the key type
+     */
+
+    @Suspendable
+    fun defaultSignatureSpec(publicKey: PublicKey, digestAlgorithm: DigestAlgorithmName): SignatureSpec?
+
```
mnesbit, line 31: Needs some doc comments of course

kyriathar: Added

mnesbit (dismissed):
OK

mnesbit (approved):
OK

PR #672, opened 2022-11-01
CORE-7326: Upgrade to Kotlin 1.7.20.

chrisr3:
Kotlin 1.7.20 has been released, as a minor update to Kotlin 1.7.10.

Contains small tweak to Kotlin generic types to resolve new compiler warning.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -7,10 +7,10 @@ cordaProductVersion = 5.0.0
 # NOTE: update this each time this module contains a breaking change
 ## NOTE: currently this is a top level revision, so all API versions will line up, but this could be moved to
 ##   a per module property in which case module versions can change independently.
-cordaApiRevision = 424
+cordaApiRevision = 425
 
 # Main
-kotlinVersion = 1.7.10
+kotlinVersion = 1.7.20
 kotlin.stdlib.default.dependency = false
 
 # These are the same annotations that Kotlin uses.


application/src/main/kotlin/net/corda/v5/application/messaging/FlowMessaging.kt
```diff
@@ -146,7 +146,7 @@ interface FlowMessaging {
      * @throws [CordaRuntimeException] if any session is closed or in a failed state.
      */
     @Suspendable
-    fun receiveAllMap(sessions: Map<FlowSession, Class<out Any>>): Map<FlowSession, Any>
+    fun receiveAllMap(sessions: Map<FlowSession, Class<out Any>>): Map<out FlowSession, Any>
```
mnesbit: Should the out be added to the value not the key type?

driessamyn (dismissed):
LGTM

application/src/main/kotlin/net/corda/v5/application/messaging/FlowMessaging.kt
mnesbit: I think this is due to the internals of the implementation and shouldn't be changing

chrisr3: The bottom line is that our generic types are wrong, but we typically paper over such errors using `uncheckedCast()`. Unfortunately, `uncheckedCast()` is a kludge at best, and the problems it tries to hide frequently generate other warnings whenever we upgrade Kotlin.

At some point, we are going to trip over something that future Kotlin will just _refuse_ to compile :crystal_ball:.

JamesHR3 (approved):
I assume you do need to bump the API revision because the runtime and API should be against the same Kotlin version? On that assumption, LGTM.

PR #681, opened 2022-11-07
CORE-7317 Closing non initiated sessions

lankydan:
Add `FlowStackItemSession` to maintain whether a session has been initiated or not, so that we do not executing closing or errorring logic upon them when a flow or subflow finishes or completes.


LWogan (dismissed):
LGTM

blsemo (approved):

PR #684, opened 2022-11-08
CORE-6674 - Identifiable States

MrMatthewLayton:
This PR adds the `IdentifiableState` API to Corda 5, and includes Java API Compatibility tests.

`IdentifiableState` replaces `LinearState` from Corda 4 and no longer requires the `UniqueIdentifier` type to be implemented. Instead, it just provides an `id` property of type `UUID`.

The `externalId` in Corda 4 was of little value and should be implemented by the CorDapp developer, if/when required.

### DIFF
--- a/ledger/ledger-utxo/src/main/kotlin/net/corda/v5/ledger/utxo/IdentifiableState.kt
+++ b/ledger/ledger-utxo/src/main/kotlin/net/corda/v5/ledger/utxo/IdentifiableState.kt
@@ -1,0 +1,15 @@
+package net.corda.v5.ledger.utxo
+
+import java.util.*
+
+/**
+ * Defines a contract state that is identifiable.
+ *
+ * The expectation is that identifiable state implementations will evolve by superseding previous versions of
+ * themselves whilst sharing a common identifier (id).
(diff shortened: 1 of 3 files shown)


ledger/ledger-utxo/src/main/kotlin/net/corda/v5/ledger/utxo/IdentifiableState.kt
```diff
@@ -0,0 +1,15 @@
+package net.corda.v5.ledger.utxo
+
+import java.util.*
```
vlajos, line 3: I think detekt does not like `*`imports.

MrMatthewLayton: Detekt is passing locally (as is the build). I think detekt is okay with wildcard imports for java libraries, but fails for stuff we add to the API.

MrMatthewLayton:
@vkolomeyko I'll close this as it's all been implemented here: https://github.com/corda/corda-utxo-ledger-extensions

PR #688, opened 2022-11-10
CORE-6996 - Expose group params

nikinagy:
Expose notaries via GroupParameters. Move GroupParameters and NotaryInfo into membership module to avoid circular dependency issue caused by `NotaryInfo`.
Renamed `party` to `name` because that's a more suitable field for notary service.


ledger/ledger-common/src/main/kotlin/net/corda/v5/ledger/common/GroupParametersExtension.kt
```diff
@@ -0,0 +1,15 @@
...
+
+import net.corda.v5.membership.GroupParameters
+
+private const val NOTARIES_KEY = "corda.notaries"
+
+/**
+ * A list of all available notary services in the group.
+ * This will parse the available notary services information from the group parameters internal property map.
+ */
+val GroupParameters.notaries: List<NotaryInfo>
```
charlieR3: Could you add a java test to make sure this API is callable from java?

ledger/ledger-common/src/main/kotlin/net/corda/v5/ledger/common/GroupParametersExtension.kt
```diff
@@ -0,0 +1,17 @@
...
```
charlieR3: Thinking about this again. I'm not sure we can do this. The coding standards advise against it:
https://r3-cev.atlassian.net/wiki/spaces/CB/pages/4082761773/Public+API+coding+standards#Extension-functions-should-add-to-existing-APIs

We may need to rethink this. Either, we move `NotaryInfo` back to membership, which will mix ledger information into the membership module, or we split the group parameters interface so we have:

```
package net.corda.v5.membership

interface MembershipGroupParameters : LayeredPropertyMap {
    val minimumPlatformVersion: Int
    val modifiedTime: Instant
    val epoch: Int
}
```

```
package net.corda.v5.ledger.common

interface GroupParameters : MembershipGroupParameters {
    val notaries: List<NotaryInfo>
}
```

Open to other suggestions also.

nikinagy: I I think that having 2 interfaces can be also confusing to people.

nikinagy: Moved notaries under membership module. Having 2 interfaces caused some issues around the `GroupParametersReader`, because that is expected to provide the notary information too, while it must be under membership module in runtime-os.

membership/src/main/kotlin/net/corda/v5/membership/GroupParameters.kt
```diff
@@ -19,23 +19,27 @@ import java.time.Instant
...
  * @property epoch An [Int] representing the version of the group parameters. This is incremented on each modification to the
  * group parameters.
+ * @property notaries A list of all available notary services in the group.
  */
 @CordaSerializable
 interface GroupParameters : LayeredPropertyMap {
     val minimumPlatformVersion: Int
     val modifiedTime: Instant
     val epoch: Int
+    val notaries: List<NotaryInfo>
```
yift-r3: Do we care about the order? (If not, can we use Collection instead of List)

membership/src/test/java/net/corda/v5/membership/GroupParametersJavaApiTest.java
```diff
@@ -66,4 +69,16 @@ public void epoch() {
         Assertions.assertThat(epoch).isNotNull();
         Assertions.assertThat(epoch).isEqualTo(669);
     }
+
+    @Test
+    public void notaries() {
+        List<NotaryInfo> notaryInfoList = new LinkedList<>();
+        notaryInfoList.add(notary);
+        when(groupParameters.getNotaries()).thenReturn(notaryInfoList);
```
yift-r3: I think you can also do:
```
when(groupParameters.getNotaries()).thenReturn(List.of(notary));
```
and remove the first two lines...

membership/src/test/java/net/corda/v5/membership/GroupParametersJavaApiTest.java
```diff
@@ -66,4 +69,16 @@ public void epoch() {
...
+
+        final List<NotaryInfo> result = groupParameters.getNotaries();
+
+        Assertions.assertThat(result).isNotEmpty();
```
yift-r3: Why not something like `assertThat(result).containsExactly(notary)`?

membership/src/main/kotlin/net/corda/v5/membership/GroupParameters.kt
nikinagy: But we will get a list anyway because we have parseList() on layered property map.

nikinagy: Or we can use parseSet().

nikinagy: This would also mean that i need to modify NotaryLookUp interface to return a collection of notaries or call.toList() on the collection there

yift-r3: I think that `parseList` is OK, but it's an implementation detail. We don't want to commit ourselves not to change it in the future unless we have to...

nikinagy: OK then changing everything including NotaryLookup to collection

charlieR3 (approved):

charlieR3:
Description needs to be updates (not using extension functions anymore)

PR #689, opened 2022-11-10
CORE-5610 included dokkahtml into serialization and produce multi module Dokka Zip

E1P3:
- Ensure Dokka formatted docs are produced by `serialization` submodule.  This required a rework also of `corda.java-only.gradle` to ensure the `org.jetbrains.kotlin.jvm` plugin is applied here as this is a requirement for Dokka Generation. This changes also required that we re work how we detect if a module is configured to compile Kotlin files.

- Update build.gradle so we can produce a [multi module](https://kotlin.github.io/dokka/1.7.10/user_guide/gradle/usage/#multi-module-projects) dokka zip, only publish this in a CI environment, add check for `publishMultiModulekDocs` when publishing here. As this is time consuming task we don't want to trigger it on `publishToMavenLocal `during local workflows 

- Accompany Jenkins changes https://github.com/corda/corda-shared-build-pipeline-steps/pull/422


serialization/build.gradle
```diff
@@ -2,6 +2,8 @@ plugins {
     id 'corda-api.common-library'
     id 'corda.common-publishing'
     id 'corda.java-only'
+    id 'org.jetbrains.dokka'
+    id 'kotlin'
```
ronanbrowne: @chrisr3  thoughts on this ?  - I want to insure no other impacts. 

Context Java docs generated by this module are not styled in the dokka format, which means our docs website had incorrect rendering for javadoc related to this module. -Docs team has raised this as a issue which needs resolving prior to beta. 

It seems however Dokka module will _only_  work on a java only module when kotlin plugin is also applied even if not used. 

see [here ](https://github.com/Kotlin/dokka/issues/2421)

chrisr3: I was hoping to exclude the possibility of anyone adding more Kotlin code to that module. In fact, didn't I create a `corda.java-only` plugin fo prevent people apply the Kotlin plugin again?

I also thought I'd ported the KDoc to Javadoc.

chrisr3: You can remove these two lines.

buildSrc/src/main/groovy/corda.java-only.gradle
```diff
@@ -1,9 +1,3 @@
...
+if (project.sourceSets.main.output.files.toString().contains("\\classes\\kotlin\\")) {
     throw new StopExecutionException("Module '${project.path}' should only contain Java classes")
-}
-
-pluginManager.withPlugin('java') {
-    java {
-        withJavadocJar()
-    }
-}
+}
```
chrisr3: Maybe rewrite more along lines of:
```gradle
plugins {
    id 'org.jetbrains.kotlin.jvm'
}

tasks.matching { it.name ==~ 'compile.*Kotlin' }.configureEach {
    doFirst {
        throw new InvalidUserCodeException("Module '${project.path}' should only contain Java classes")
    }
}
```
The "root" `build.gradle` file should already be applying the `dokka` plugin to anything that applies `org.jetbrains.kotlin.jvm`, so it's just a matter of preventing anyone from compiling Kotlin classes.

This is simpler than I thought - the Kotlin compile tasks don't execute if Gradle doesn't find any source files, and we don't want this to fail silently.

chrisr3 (changes requested):
We should only need to change the `compile.java-only` plugin.

ronanbrowne:
Putting PR back in draft while finalizing some testing - taking this over from Jan.

Jenkinsfile
```diff
@@ -1,13 +1,11 @@
-@Library('corda-shared-build-pipeline-steps@5.0') _
```
ronanbrowne: Changes in this fie to be reverted just while testing changes

buildSrc/src/main/groovy/corda.java-only.gradle
ronanbrowne: Ah nice, updated thanks

chrisr3 (approved):
LGTM.

yarivus (approved):
LGTM

PR #691, opened 2022-11-11
CORE-7288 Add digest service for flows

kyriathar:
Purpose of this PR is to distinct sandbox digest service from platform digest service. Platform digest service should be a platform singleton to be used for our internal needs and also used in sandbox digest service. Sandbox digest service should be a singleton per sandbox which should reference the platform digest service and a list of custom digest algorithms found in CorDapps. So, sandbox digest service should look up in both the platform digest service and its custom digest algorithms.

- adds digest service (sandbox singleton) to be used by all sandbox types.
- updates name of platform digest service (platform singleton).  

runtime-os PR: [#2502](https://github.com/corda/corda-runtime-os/pull/2502)


kyriathar:
Registered [CORE-8151](https://r3-cev.atlassian.net/browse/CORE-8151) to provide APIs for getting: 
* default digest algorithm 
* supported digest algorithms

cipher-suite/src/main/kotlin/net/corda/v5/cipher/suite/PlatformDigestService.kt
```diff
@@ -0,0 +1,41 @@
+package net.corda.v5.cipher.suite
+
+import net.corda.v5.crypto.DigestAlgorithmName
+import net.corda.v5.crypto.SecureHash
+import java.io.InputStream
+
+/**
+ * The platform digest calculation service. Holds a list of supported digest algorithms to be used for hashing.
+ */
+interface PlatformDigestService {
```
dickon, line 10: As discussed, and I think you said @vlajos also suggseted, if it works it'd be better to have a base DigestService interface and then PlatformDigestService and and FlowDigestService which simply extend the interface while adding nothing, so we still have 2 different type to depend on.

dickon: However I'd like to get this merged before you go on holiday so if we get everything working with the 2 separate interface strucutre you have here then let's merge that and we can revisit this question.

dickon (dismissed):

cipher-suite/src/main/kotlin/net/corda/v5/cipher/suite/PlatformDigestService.kt
kyriathar: I think I agree with having a base `DigestService` and then platform one becomes a internal marker interface to extend for OSGi to distinguish between the two because it seems like there could be constructs to utilize both (currently they both have the exact same API). But we can consider doing it at a later point when we internalize it (i.e. when move cipher-suite module to corda-runtime-os).

blsemo: Would a platform digest service even need to be on the public API? What use case would we have where a user of the public API would get hold of the non-sandboxed version?

kyriathar: `cipher-suite` api module will be moved to corda-runtime-os with [CORE-7976](https://r3-cev.atlassian.net/browse/CORE-7976). Is to be done right next this gets merged since we will be now offering digest service to flows and other sandboxes actually.

mnesbit: I would rather not link the public api and the internal platform one, so separate interfaces is correct. Sharing is also likely to produce issues with OSGi. Fundamentally the stuff in corda-api is for public consumption and we might end up with several versions of it, where as the platform one is the infrastructure to be used internally

mnesbit (approved):
OK

PR #693, opened 2022-11-11
CORE-6969 Add SignatureSpec.equals(), hashcode()

vlajos:
CORE-6969 Add SignatureSpec.equals(), hashcode()


kyriathar (approved):

lankydan (approved):
Approving it based on @kyriathar's approval.

PR #700, opened 2022-11-16
CORE-2368: Add certificate read APIs

yift-r3:
The runtime Pull request can be found at https://github.com/corda/corda-runtime-os/pull/2556

This pull request adds a command to list the certificates aliases.


charlieR3 (dismissed):
LGTM. Although I think this is technically not a breaking change so the API bump might not be necessary.

charlieR3 (dismissed):

charlieR3 (approved):

PR #701, opened 2022-11-16
CORE-8145: Add missing entry to PK

benmcmahon100:
The [CpiCpkEntity](https://github.com/corda/corda-runtime-os/blob/799d93fa95e253217e8a74b3199b6a8b0ee707f6/libs/virtual-node/cpi-datamodel/src/main/kotlin/net/corda/libs/cpi/datamodel/CpiCpkEntity.kt#L59-L73) in the runtime os repo expects the `cpk_signer_summary_hash` to be part of the pk. This change aligns this with the migration here in API

### DIFF
--- a/data/db-schema/src/main/resources/net/corda/db/schema/config/migration/cpx-creation-v1.0.xml
+++ b/data/db-schema/src/main/resources/net/corda/db/schema/config/migration/cpx-creation-v1.0.xml
@@ -254,7 +254,7 @@
             </column>
         </createTable>
 
-        <addPrimaryKey tableName="cpi_cpk" columnNames="cpi_name, cpi_version, cpi_signer_summary_hash, cpk_file_checksum"
+        <addPrimaryKey tableName="cpi_cpk" columnNames="cpi_name, cpi_version, cpi_signer_summary_hash, cpk_name, cpk_version, cpk_signer_summary_hash, cpk_file_checksum"
                        constraintName="cpi_cpk_pk" schemaName="${schema.name}"/>
 
         <addForeignKeyConstraint baseTableName="cpi_cpk" baseColumnNames="cpi_name, cpi_version, cpi_signer_summary_hash"
--- a/gradle.properties
+++ b/gradle.properties
@@ -7,7 +7,7 @@ cordaProductVersion = 5.0.0
 # NOTE: update this each time this module contains a breaking change
 ## NOTE: currently this is a top level revision, so all API versions will line up, but this could be moved to
 ##   a per module property in which case module versions can change independently.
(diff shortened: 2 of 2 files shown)


vkolomeyko (approved):

gradle.properties
```diff
@@ -7,7 +7,7 @@ cordaProductVersion = 5.0.0
 # NOTE: update this each time this module contains a breaking change
 ## NOTE: currently this is a top level revision, so all API versions will line up, but this could be moved to
 ##   a per module property in which case module versions can change independently.
-cordaApiRevision = 500
```
vkolomeyko, line 10: Is it a breaking change, though?

driessamyn: It shouldn't be. But I don't think the change is right.

data/db-schema/src/main/resources/net/corda/db/schema/config/migration/cpx-creation-v1.0.xml
```diff
@@ -254,7 +254,7 @@
             </column>
         </createTable>
 
-        <addPrimaryKey tableName="cpi_cpk" columnNames="cpi_name, cpi_version, cpi_signer_summary_hash, cpk_file_checksum"
+        <addPrimaryKey tableName="cpi_cpk" columnNames="cpi_name, cpi_version, cpi_signer_summary_hash, cpk_name, cpk_version, cpk_signer_summary_hash, cpk_file_checksum"
```
driessamyn, line 257: I'm not sure this makes sense. `cpk_file_checksum is unique`. It shouldn't be possible to have an identical file that is signed by a different key.
(Name and Version could technically be, but maybe instead we should do a validation that this matches what's in the manifest so that you can't have 2 identical files with a different name and version.
@josephzunigadaly may be able to add more detail.

benmcmahon100: This was from @kyriathar suspicion., I see what you mean though about the cpk signer summary. Let me stick this back in draft and try get the integration tests to reproduce the error (I thought I had, but was running into another issue), and I can swap this to do the opposite and remove the cpk signer from the EntityKey

driessamyn: I'm not saying it's not possible that the issue is related to this, but I would rather fix something that we know is broken.

One thing @josephzunigadaly mentioned yesterday is that the signer has is still nullable in the code (left-over from V1). It is possible something somewhere isn't wiring up. I think @josephzunigadaly was going to look at making that change.

PR #702, opened 2022-11-16
CORE-7627 Data model and storage code for the UTXO ledger

mbrkic-r3:
- Added DB table utxo_transaction_status

### DIFF
--- a/data/db-schema/src/main/resources/net/corda/db/schema/vnode-vault/migration/ledger-utxo-creation-v1.0.xml
+++ b/data/db-schema/src/main/resources/net/corda/db/schema/vnode-vault/migration/ledger-utxo-creation-v1.0.xml
@@ -20,6 +20,18 @@
             </column>
         </createTable>
 
+        <createTable tableName="utxo_transaction_status">
+            <column name="transaction_id" type="VARCHAR(160)">
+                <constraints nullable="false" foreignKeyName="fk_utxo_transaction_status" references="utxo_transaction(id)"/>
+            </column>
+            <column name="status" type="VARCHAR(255)">
+                <constraints nullable="false"/>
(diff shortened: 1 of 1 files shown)


mbrkic-r3:
DB table `utxo_transaction_status` was initially left out until we are clear whether status should be an updatable field or we need a dedicated table. Since persisting status is important for flow code, I've implemented it in the same way as it is done for Consensual Ledger and we can modify it later.

data/db-schema/src/main/resources/net/corda/db/schema/vnode-vault/migration/ledger-utxo-creation-v1.0.xml
```diff
@@ -20,6 +20,18 @@
             </column>
         </createTable>
 
+        <createTable tableName="utxo_transaction_status">
```
owenstanford, line 23: @mnesbit @blsemo 
I can't remember what we agreed in the end on this vs a field in on the tx table, if the purpose is an audit trail of status changes, then I think we should consider either:
- adding a version # to the tx and the status table to make the join 1-1
- adding the current status to the tx table and keep the status table as a pure audit table.

lankydan (approved):

data/db-schema/src/main/resources/net/corda/db/schema/vnode-vault/migration/ledger-utxo-creation-v1.0.xml
mbrkic-r3: We agreed to left it out until we are clear what we need, but Dan said he needs it now in some form so I've added a table that is same as we have for consensual ledger, and we can modify it later.

PR #703, opened 2022-11-16
NOTICK Remove not needed OSGi uses from api modules

kyriathar:
- removes trying to load resource going through `Bundle`. It should not be needed because current class is already loaded whether it is through OSGi or not. Going through the `Bundle` should again ending up calling the same class loader. 
- makes modules OSGi free.


chrisr3 (approved):
LGTM.

JamesHR3 (approved):

mag1c-48 (approved):
LGTM (build)

PR #713, opened 2022-11-29
CORE-8431 Interfaces and functions for getting data out of signed TX

blsemo:
Add functionality to the UTXO signed transaction to get some of the data out without transforming to a ledger transaction (which would be costly because it requires backchain resolution).
Currently limited to the properties that will be required for the notary.

### DIFF
--- a/ledger/ledger-common/src/main/kotlin/net/corda/v5/ledger/common/transaction/TransactionMetadata.kt
+++ b/ledger/ledger-common/src/main/kotlin/net/corda/v5/ledger/common/transaction/TransactionMetadata.kt
@@ -1,0 +1,59 @@
+package net.corda.v5.ledger.common.transaction
+
+/**
+ * This interface represents the metadata governing a transaction, capturing a snapshot
+ * of the system and software versions as the transaction was created.
+ */
+interface TransactionMetadata {
+
+    /**
+     * Returns the ledger model this transaction belongs to.
+     *
+     * @return the class name of the transaction implementation
+     */
+    fun getLedgerModel(): String
+
+    /**
+     * Returns the version of the ledger this transaction was created with
+     *
+     * @return the ledger version at creation time
+     */
+    fun getLedgerVersion(): Int
+
+    /**
+     * Returns the transaction subtype. This is ledger specific, check with the documentation of the
+     * ledger model you are using.
+     *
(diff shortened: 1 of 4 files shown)


mnesbit (dismissed):
OK

MrMatthewLayton (dismissed):
LGTM

ledger/ledger-utxo/src/main/kotlin/net/corda/v5/ledger/utxo/transaction/UtxoSignedTransaction.kt
```diff
@@ -34,6 +41,46 @@ interface UtxoSignedTransaction {
...
+
+    /**
+     * @property timeWindow The validity time window for completing/notarising this transaction
+     */
+    val timeWindow: TimeWindow
+
+    /**
+     * @property metadata The metadata for this transaction
+     */
+    val metadata: TransactionMetadata
```
lankydan, line 72: Why do we need to make the metadata public?

ledger/ledger-common/src/main/kotlin/net/corda/v5/ledger/common/transaction/TransactionMetadata.kt
```diff
@@ -0,0 +1,59 @@
...
+     * @return Transaction subtype as string defined in the ledger model
+     */
+    fun getTransactionSubtype(): String?
+
+    /**
+     * Get information about the CPI running on the virtual node creating the transaction
+     *
+     * @return a summary of the CPI
+     */
+    fun getCpiMetadata(): CordaPackageSummary?
```
lankydan, line 36: If we're making `TransactionMetadata` public we should probably do something about the nullability of the metadata fields. Some of these are `?` and others aren't. I think they should either all be nullable or none be nullable and throw an error if its missing.

vlajos: AFAIK CPI should never be empty. (Although we only fill it with some fake data temporarily.)
`TransactionSubtype` is not really needed for Consensual ledger, but I think we agreed that we was going to put there a placeholder value ("normal"(?)), just to avoid diverging the two ledgers's metadata schemas just for this field.

lankydan (approved):

PR #716, opened 2022-12-01
CORE-5096 : Initial API implementation for encumbrance

MrMatthewLayton:
Initial API implementation for encumbrance
Refactored UTXO Transaction Builder APIs to better support encumbrance groups.


lankydan (dismissed):

blsemo:
Shouldn't there be a way to discover all states in an encumbrance group from a ledger tx, for example? That would seem useful to check that the whole group gets spend in one go?

blsemo (approved):
Looks ok to me

mcgovc (approved):
LGTM - build changes

PR #717, opened 2022-12-01
CORE-8618 One status row per transaction

mbrkic-r3:
- Set primary key for DB tables consensual_transaction_status and utxo_transaction_status


data/db-schema/src/main/resources/net/corda/db/schema/vnode-vault/migration/ledger-consensual-creation-v1.0.xml
```diff
@@ -22,7 +22,7 @@
 
         <createTable tableName="consensual_transaction_status">
             <column name="transaction_id" type="VARCHAR(160)">
-                <constraints nullable="false" foreignKeyName="fk_transaction_status" references="consensual_transaction(id)"/>
+                <constraints nullable="false" primaryKey="true" foreignKeyName="fk_transaction_status" references="consensual_transaction(id)"/>
             </column>
             <column name="status" type="VARCHAR(255)">
```
vlajos, line 27: Not related to this PR, but are not statuses 1 character long? (`U`, `V`) ?

mbrkic-r3: Yes, they are at the moment, as part of initial implementation to have it working, but status values in DB were not properly discussed, not sure whether these would be the final values.

vlajos (dismissed):

data/db-schema/src/main/resources/net/corda/db/schema/vnode-vault/migration/ledger-consensual-creation-v1.0.xml
lankydan: Ye 160 is probably too long but might not want to restrict to 1.

Do we have a tech debt to go through the schemas and make sure we're happy with everything? If not, we probably should so we can tick it off and move on at some point. Definitely a do later thing though.

vlajos: I added it here a bit earlier: https://r3-cev.atlassian.net/browse/CORE-7928

data/db-schema/src/main/resources/net/corda/db/schema/vnode-vault/migration/ledger-utxo-creation-v1.0.xml
```diff
@@ -22,12 +22,12 @@
...
         <createTable tableName="utxo_transaction_status">
             <column name="transaction_id" type="VARCHAR(160)">
-                <constraints nullable="false" foreignKeyName="fk_utxo_transaction_status" references="utxo_transaction(id)"/>
+                <constraints nullable="false" primaryKey="true" foreignKeyName="fk_utxo_transaction_status" references="utxo_transaction(id)"/>
             </column>
             <column name="status" type="VARCHAR(255)">
                 <constraints nullable="false"/>
             </column>
-            <column name="created" type="TIMESTAMP">
+            <column name="updated" type="TIMESTAMP">
```
lankydan, line 30: This is convenient for me since I had an `UPDATE` statement in my own PR so this means I don't need to change it too much. That being said, I need to make it update the `updated` column. I'll raise a ticket for that.

lankydan (dismissed):

lankydan (approved):

vlajos (approved):

PR #719, opened 2022-12-02
CORE-8432 filtered tx interface

blsemo:
Interface for a public UTXO filtered transaction - no version bump required at this point as this is all new interfaces.


ledger/ledger-utxo/src/main/kotlin/net/corda/v5/ledger/utxo/transaction/UtxoFilteredTransaction.kt
```diff
@@ -0,0 +1,69 @@
...
+
+    /**
+     * @param commands Potentially filtered list of commands
+     */
+    val commands: UtxoFilteredData<Command>
+
+    /**
+     * Verifies the current [UtxoFilteredTransaction].
+     *
+     * @throws TransactionVerificationException if the current [UtxoSignedTransaction] fails to verify correctly.
```
vlajos:
```suggestion
     * @throws TransactionVerificationException if the current [UtxoFilteredTransaction] fails to verify correctly.
```
?

ledger/ledger-utxo/src/main/kotlin/net/corda/v5/ledger/utxo/transaction/UtxoFilteredTransaction.kt
```diff
@@ -0,0 +1,69 @@
...
+import net.corda.v5.ledger.common.transaction.TransactionMetadata
+import net.corda.v5.ledger.common.transaction.TransactionVerificationException
+import net.corda.v5.ledger.utxo.Command
+import net.corda.v5.ledger.utxo.StateAndRef
+import net.corda.v5.ledger.utxo.StateRef
+import net.corda.v5.ledger.utxo.TimeWindow
+import java.security.PublicKey
+
+@DoNotImplement
+interface UtxoFilteredTransaction {
```
relyafi, line 35: Think this could do with a KDoc comment like the other interfaces

ledger/ledger-utxo/src/main/kotlin/net/corda/v5/ledger/utxo/transaction/UtxoFilteredTransaction.kt
```diff
@@ -0,0 +1,69 @@
...
+
+    /**
+     * @param timeWindow The validity time window for finalizing/notarising this transaction or null if filtered
+     */
+    val timeWindow: TimeWindow?
+
+    /**
+     * @param notary The notary party for this transaction or null if filtered
+     */
+    val notary: Party?
```
relyafi, line 54: Why are these null, as opposed to being a piece of filtered data using the "removed" interface?

This is particularly concerning for time windows, because we rely on the upper bound being part of the merkle tree to prevent state reissuance after being archived.

ledger/ledger-utxo/src/test/java/net/corda/v5/ledger/utxo/transaction/UtxoFilteredTransactionJavaApiTest.java
```diff
@@ -0,0 +1,66 @@
...
+
+    protected final SecureHash hash = SecureHash.parse("SHA256:0000000000000000000000000000000000000000000000000000000000000000");
+
+    @Test
+    void canReadFilteredTransaction() {
+        // Set up Filtered TX mock
+        UtxoFilteredTransaction filteredTx = Mockito.mock(UtxoFilteredTransaction.class);
+
+        // Notary, time window commands
+        Mockito.when(filteredTx.getNotary()).thenReturn(null);
```
relyafi, line 24: For the notary use case, it would probably make sense to set this - no point in hiding the notary from itself, and it potentially allows a bit of extra protection (ensuring that the notary in question is the intended recipient)

ledger/ledger-utxo/src/test/java/net/corda/v5/ledger/utxo/transaction/UtxoFilteredTransactionJavaApiTest.java
```diff
@@ -0,0 +1,66 @@
...
+        Mockito.when(outputs.getSize()).thenReturn(5);
+        Mockito.when(filteredTx.getOutputStateAndRefs()).thenReturn(outputs);
+
+        // Inputs is a filtered Audit proof
+        FilteredEntry<StateRef> inputData = Mockito.mock(FilteredEntry.class);
+        Mockito.when(inputData.getIndex()).thenReturn(1);
+        Mockito.when(inputData.getValue()).thenReturn(new StateRef(hash, 0));
+        UtxoFilteredData.UtxoFilteredDataAudit<StateRef> inputs =
+                Mockito.mock(UtxoFilteredData.UtxoFilteredDataAudit.class);
+        Mockito.when(inputs.getSize()).thenReturn(2);
```
relyafi, line 37: So is the implication that a filtered audit proof will always imply a size proof? I.e. it looks like this test case is choosing to selectively reveal one input state, but is also declaring a size of 2. I assume there are no use cases where we would want to reveal a subset of the data, but hide the total number?

ledger/ledger-utxo/src/main/kotlin/net/corda/v5/ledger/utxo/transaction/UtxoFilteredTransaction.kt
blsemo: They are nullable because technically, they are part of the same component group, but on the interfaces for transactions, they are separate, and they are never lists, always just one entry. 
So it's mostly a convenience thing that they are just nullable vals rather than being put through the whole removed interface shenanigans (and a size proof makes inherentely no sense).
They can technically be filtered, but that would obviously be wrong if they do get filtered out for a non-validating notary payload.

ledger/ledger-utxo/src/test/java/net/corda/v5/ledger/utxo/transaction/UtxoFilteredTransactionJavaApiTest.java
blsemo: I don't think that can be done - an audit proof of the underling tx always has the size, and needs to give you the position of the revealed fields, so you'd always get a lower bound on the size.

ledger/ledger-utxo/src/main/kotlin/net/corda/v5/ledger/utxo/transaction/UtxoFilteredTransaction.kt
relyafi: More generally though, I guess its possible the transaction tear off use-case might want to filter them?

I see now though what you mean about its only list based data that is using the filtered data objects. In that case, either the naming or the abstraction doesn't quite feel right to me, if filtered data is specifically filtered _list_ data. It doesn't seem consistent from an API perspective to have two ways of checking for absence depending on whether the data of interest is a list or not. Maybe what we're missing is a fourth filtered data type - we need the current 3 - absent, size (only makes sense to lists), _list_ data audit, and a new _single_ data audit. Or capture audit single items as a list of size 1.

blsemo: Hmm, I am not quite sure I follow here - clearly the UtxoFilteredTransaction has some more specifications, so we know a priori that the SignedTransaction has one notary and one time window - so in the filtered transaction, they are either present or filtered out (null).

ledger/ledger-utxo/src/main/kotlin/net/corda/v5/ledger/utxo/transaction/UtxoFilteredData.kt
```diff
@@ -0,0 +1,56 @@
...
+ * * A size-only merkle proof - in this case, we can only retrieve the size of the original list
+ * * An audit proof - in this case, we get access to some or all of the entries of the original list.
+ */
+sealed interface UtxoFilteredData<T> {
+
+    /**
+     * Marker interface for a completely filtered out component group. No further information is available
+     */
+    @DoNotImplement
+    interface UtxoFilteredDataRemoved<T> : UtxoFilteredData<T>
```
lankydan: Since these are inner classes, I think we can drop the `UtxoFiltered` from their names.

I'm 50/50 on whether we keep the `sealed` part though. Generally we don't want `sealed` interfaces, but on the other hand, I don't see this being extended in the future. Note, they were sealed in my implementation partly because they were internal.

ledger/ledger-utxo/src/main/kotlin/net/corda/v5/ledger/utxo/transaction/UtxoFilteredTransaction.kt
relyafi: Yeah, I'm not saying that what's there doesn't work. It's just putting my API user hat on, it feels a bit inconsistent - that _how_ I check for whether something is filtered or not depends on whether the item is a list or not. And the other assumption which you've touched on, is that the null approach only works for items that are mandatory. That may be a safe assumption for notary and time window when constructing a filtered transaction for the notary to consume, but not 100% sure if it would hold for transaction tear offs between peers. Also, if we extend this interface in future with (non-list) optional attributes, then we would have no ability to distinguish between "filtered" and "supplied but absent".

Personally I would ensure that there is an explicit "filtered" type for everything, rather than relying on null for singular arguments, but it's your call.

mnesbit (approved):
Not sure I quite see all the pieces, but harmless

PR #722, opened 2022-12-05
CORE-8672 Storing relevancy information

mbrkic-r3:
- Added `relevantStates` to `PersistTransaction` ledger Avro schema
- Removed column `utxo_transaction_relevancy.is_relevant`


lankydan (dismissed):

data/avro-schema/src/main/resources/avro/net/corda/data/ledger/persistence/ComponentPosition.avsc
```diff
@@ -0,0 +1,18 @@
+{
+  "type": "record",
+  "name": "ComponentPosition",
+  "doc": "The group index and the leaf index of the transaction component",
+  "namespace": "net.corda.data.ledger.persistence",
+  "fields": [
+    {
+      "name": "groupIndex",
```
blsemo: Do we need the group index here? Shouldn't that always be OutputStates?

data/db-schema/src/main/resources/net/corda/db/schema/vnode-vault/migration/ledger-utxo-creation-v1.0.xml
```diff
@@ -166,9 +166,6 @@
             <column name="leaf_idx" type="INT">
                 <constraints nullable="false"/>
             </column>
-            <column name="is_relevant" type="BOOLEAN">
-                <constraints nullable="false"/>
-            </column>
```
blsemo, line 171: So we don't store it this way, but where is the schema for storing the relevancy information in the new format? What did I miss?

blsemo: Reading the code - do we reuse the tx output table to only record relevant states in it?

mbrkic-r3: My understanding was that we only insert into table `utxo_transaction_relevancy` when component is relevant, in which case this field would always be `true`, isn't that the case?

data/avro-schema/src/main/resources/avro/net/corda/data/ledger/persistence/ComponentPosition.avsc
mbrkic-r3: I was not aware that is always fixed, I'll change it

data/db-schema/src/main/resources/net/corda/db/schema/vnode-vault/migration/ledger-utxo-creation-v1.0.xml
blsemo: Ok, that makes sense.

mbrkic-r3: > Reading the code - do we reuse the tx output table to only record relevant states in it?

Tx output table is not used at the moment.

data/avro-schema/src/main/resources/avro/net/corda/data/ledger/persistence/ComponentPosition.avsc
mbrkic-r3: fixed

blsemo (dismissed):
LGTM

blsemo (approved):
Re-approved

PR #723, opened 2022-12-06
CORE-7248 Add notary plugin loading logic

nkovacsx:
## Overview

This PR introduces a new CorDapp manifest element called `Corda-NotaryPluginProvider-Classes`. This property replaces the old notary service classes which was not used anyway.

This property will perform a discovery on classes that implements `PluggableNotaryClientFlowProvider`. These classes will be loaded (in the `runtime-os` PR) into a factory called `PluggableNotaryClientFlowFactory` which can be injected into flows, and with its `create` method the given plugin client flow can be instantiated.


ledger/notary-plugin/src/main/kotlin/net/corda/v5/ledger/notary/plugin/api/PluggableNotaryClientFlowProvider.kt
```diff
@@ -0,0 +1,17 @@
...
+ * This interface is used to instantiate the client-side logic (flow) of the plugin.
+ *
+ * Implementers of this interface must have the [PluggableNotaryType] annotation,
+ * otherwise an exception will be thrown on loading.
+ *
+ * The client-side of the plugin must have the same constructor parameters as the
+ * [create] function (notary, stx).
+ */
+interface PluggableNotaryClientFlowProvider {
+    fun create(notary: Party, stx: UtxoSignedTransaction): PluggableNotaryClientFlow
```
relyafi, line 16: I don't think this dependency is new, but it's interesting that the `notary-plugin` module depends on `ledger-utxo`.

Clearly this isn't a major issue right now, but worth thinking about whether this is what we want given there may be e.g. a utxo v2 ledger in future, and this is in API. Maybe it's just a case of the plugin module / package name needs to reflect that plugins are ledger model specific, if that is the implication? Or would we envisage this single interface being used with different models, and an overloaded `create` function?

relyafi (dismissed):
Pretty much LGTM subject to issues in corresponding runtime-os changes being resolved. Comment is for consideration, can be ignored if we don't think it's relevant.

ledger/notary-plugin/src/main/kotlin/net/corda/v5/ledger/notary/plugin/api/PluggableNotaryClientFlowProvider.kt
nkovacsx: I understand your concern but I'm not entirely sure what would be the best solution here. Overloading is something I'd like to avoid. Maybe we could use a type parameter and make this interface generic, so it can be extended in the future?

PR #725, opened 2022-12-07
CORE-8870 Increase the interval between group parameters database reconciliation

charlieR3:
Group parameters reconciliation was set to a 30 second interval by default. This can be an expensive operation as the number of vnodes increases and for the combined worker this can take up processing time and impact performance. In general we don't expect the group parameters to change often so the PR increases that interval to 2 minutes.

Reconciliation time for group parameters on the combined worker when 11 vnodes are created was observed to be ~4 seconds.


nikinagy (approved):
LGTM

PR #727, opened 2022-12-09
CORE-5096 : Revert Encumbrance Changes

MrMatthewLayton:
Need to revert encumbrance changes as runtime is not working.


MrMatthewLayton:
Not needed.

PR #728, opened 2022-12-09
CORE-7248 Add notary plugin loading logic API

nkovacsx:
## Overview

This PR introduces a new CorDapp manifest element called `Corda-NotaryPluginProvider-Classes`. This property replaces the old notary service classes which was not used anyway.

This property will perform a discovery on classes that implements `PluggableNotaryClientFlowProvider`. These classes will be loaded (in the `runtime-os` PR) into a factory called `PluggableNotaryClientFlowFactory` which can be injected into flows, and with its `create` method the given plugin client flow can be instantiated.


nkovacsx:
I had to re-open this PR because the previous one got mixed up with all the reverts etc. due to the 24+ hours lockings.

ledger/notary-plugin/src/main/kotlin/net/corda/v5/ledger/notary/plugin/api/PluggableNotaryClientFlowProvider.kt
```diff
@@ -0,0 +1,17 @@
...
+ *
+ * Implementers of this interface must have the [PluggableNotaryType] annotation,
+ * otherwise an exception will be thrown on loading.
+ *
+ * The client-side of the plugin must have the same constructor parameters as the
+ * [create] function (notary, stx).
+ */
+interface PluggableNotaryClientFlowProvider {
+    fun create(notary: Party, stx: UtxoSignedTransaction): PluggableNotaryClientFlow
+}
```
blsemo, line 17: This runs a bit contrary how to system flows are implemented in Corda 5. 
For system flows, user pass in the information + flow sessions, and a subflow is started under the hood - I guess in this case, all you do is get the subflow created by this method and then start it?

blsemo: But maybe that is ok because in this case, the implementation is potentially done by the customer, and the caller is a system flow?

nkovacsx: This won't start the plugin flow, just load the proper flow (that is corresponding to the notary's `pluginClass` property) from the sandbox and instantiate it. It will be started by the finality flow later on.

blsemo (approved):
LGTM

PR #730, opened 2022-12-12
CORE-8672 State relevancy calculation

mbrkic-r3:
- Implemented calculation of state relevancy in Contract interface


ledger/ledger-utxo/src/main/kotlin/net/corda/v5/ledger/utxo/Contract.kt
```diff
@@ -13,6 +15,17 @@ import net.corda.v5.ledger.utxo.transaction.UtxoLedgerTransaction
 @CordaSerializable
 interface Contract {
 
+    companion object {
+        /**
+         * Establish whether a given state is relevant to a node, given the node's public keys.
+         *
+         * A state is relevant if any of the participants key matching one of this node's public keys.
+         */
+        fun isRelevant(state: ContractState, myKeys: Set<PublicKey>): Boolean {
```
blsemo: I am not sure that is what we had in mind for this - the companion object of an interface is just a static class, not part of the interface, so customer contracts could not override this method.
I was thinking to have a method with this or a similar signature on the Contract interface so contracts can override the relevancy calculation. We'd need to provide this implementation as a default.

lankydan: To clarify you're saying having something like?:

```kotlin
interface Contract {
  // default implementation
  fun isRelevant(state: ContractState, keys: Set<PublicKey>): Boolean {
    return state.participants.any { it.containsAny(keys) }
  }
}
```

So that users can override how relevancy per contract is calculated?

Meaning someone can do?:

```kotlin
class MyContract : Contract {
  override fun isRelevant(state: ContractState, keys: Set<PublicKey>): Boolean {
    return true
  }
}
```

I think this sounds fine, however I do think its different from C4, so just making sure we're consciously making this decision.

@mnesbit @blsemo

blsemo: :+1: That is exactly how I understood the outcome of the isRelevant discussion.

mnesbit: I think we can't pass in a general set of our keys as it could be massive. I think we can pass in the key manager service once we have the my keys API. I am wondering if there are other signatures that might be helpful. e.g. we could have isRelevant change to a getRelevantKeys function that extracts keys from the state e.g. owner, issuer and then we check if any of those are ours.

mnesbit: C4 code is https://github.com/corda/corda/blob/release/os/4.10/node/src/main/kotlin/net/corda/node/services/vault/NodeVaultService.kt#L73

mnesbit: But it has definite scaling issues that we hope we can fix by cacheing the my key results on key manager

mbrkic-r3: if that function is not static, how do I reach contract from transaction?

lankydan: You can get contracts from states. Similar to how was done here. https://github.com/corda/corda-runtime-os/pull/2501/files#diff-22160ff83ade8c279f41057503e1571a4e22505a61bcce7c33ab1f4fc58a678d

mbrkic-r3: When will we have "my keys API", do I leave it as it is for now?

mbrkic-r3: function moved out of companion object

lankydan: So it sounds like there is a chance that this API (after moving to interface) is not correct. Due to time, and the fact the calculation is pretty much the same as it was in C4, I say we either leave it as is and review it later (after friday) or make it internal and then review.

lankydan: @mnesbit ^

lankydan (approved):

ledger/ledger-utxo/src/main/kotlin/net/corda/v5/ledger/utxo/Contract.kt
mbrkic-r3: Due to time, I'm going to merge this and if something needs to be changed, it will be done in a different PR

PR #733, opened 2022-12-13
CORE-8915: Set producers and consumers for certificates and membership topics

yift-r3:
Runtime PR in https://github.com/corda/corda-runtime-os/pull/2721

This PR adds consumers and producers to the certificates and membership topics. Also a little cleanup for the scheme names.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -7,7 +7,7 @@ cordaProductVersion = 5.0.0
 # NOTE: update this each time this module contains a breaking change
 ## NOTE: currently this is a top level revision, so all API versions will line up, but this could be moved to
 ##   a per module property in which case module versions can change independently.
-cordaApiRevision = 518
+cordaApiRevision = 519
 
 # Main
 kotlinVersion = 1.7.21
--- a/data/topic-schema/src/main/kotlin/net/corda/schema/Schemas.kt
+++ b/data/topic-schema/src/main/kotlin/net/corda/schema/Schemas.kt
@@ -42,7 +42,7 @@ class Schemas {
             const val CONFIG_TOPIC = "config.topic"
             const val CONFIG_MGMT_TOPIC = "config.management"
             const val CONFIG_MGMT_REQUEST_TOPIC = "config.management.request"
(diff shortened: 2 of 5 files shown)


gradle.properties
```diff
@@ -7,7 +7,7 @@ cordaProductVersion = 5.0.0
 # NOTE: update this each time this module contains a breaking change
 ## NOTE: currently this is a top level revision, so all API versions will line up, but this could be moved to
 ##   a per module property in which case module versions can change independently.
-cordaApiRevision = 518
```
vkolomeyko, line 10: Does it really need a version bump?
(it is easier to integrate a change if not)

yift-r3: I think so (without it, if you branched out before https://github.com/corda/corda-runtime-os/pull/2726 was merged, your PRs will fail)

driessamyn: I don't think it needs it.
You can still run a runtime-os test to be sure by creating a draft PR with the precise alpha version that this build is producing (similar to what I did here: https://github.com/corda/corda-runtime-os/pull/2730)

yift-r3: That's what I did, and that is how I discovered that it is failing, and I think it will fail on any branch without the fix...

charlieR3 (approved):

gradle.properties
driessamyn: It's not needed.
See: https://github.com/corda/corda-api/pull/737

yift-r3: (You just broke all the branches that merged out before today...)

driessamyn: How so? The build proves it works: https://ci02.dev.r3.com/blue/organizations/jenkins/Corda5%2Fcorda-runtime-os/detail/PR-2730/3/pipeline

yift-r3: That branch has the fix, but a branch that was branched out before the fix was merged would have failed.

driessamyn: That doesn't matter. Jenkins merges before it builds, so any new builds will pick up David's fix.
Locally it doesn't matter as far as I understand the dev values file doesn't lock down the topics.

PR #749, opened 2022-12-17
NOTICK: javadocJar depends on assemble task

ronanbrowne:
add javadoc generation to assemble task to prevent the need for a separate dedicated target to be called in the CI in a effort to reduce build times (less gradle stages = less configuration overhead if we can do the work in one stage) , ./gradlew build or assemble will now produce this java doc.

`generateJavaDocWithBuild` property must be present for this, property introduced as in most cases this is not required when building locally.

Associated Shared-library PR https://github.com/corda/corda-shared-build-pipeline-steps/pull/457

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -214,8 +214,10 @@ subprojects {
             from(dokkaHtml.outputDirectory)
         }
 
-        if(project.hasProperty('generateJavaDocWithBuild')) {
-                assemble.dependsOn javadocJar
+        if (project.hasProperty('generateJavaDocWithBuild')) {
+            artifacts {
+                archives javadocJar
+            }
         }
     }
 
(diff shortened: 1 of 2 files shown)


build.gradle
```diff
@@ -213,6 +213,10 @@ subprojects {
             archiveClassifier.set("javadoc")
             from(dokkaHtml.outputDirectory)
         }
+
+        if(project.hasProperty('generateJavaDocWithBuild')) {
```
yarivus: fix style (indent) and spaces next to () and if

yarivus (changes requested):

build.gradle
```diff
@@ -213,6 +213,10 @@ subprojects {
...
+                assemble.dependsOn javadocJar
```
chrisr3: Alternatively:
```gradle
artifacts {
    archives javadocJar
}
```

yarivus (approved):
LGTM

PR #750, opened 2022-12-19
CORE-8989 Removed unused non-nullable field that was causing the toke…

owenstanford:
BUG FIX:
Removing a field that was added by mistake in a previous commit, this field was causing the token cache to fail.


ronanbrowne (approved):
lgtm

blsemo (approved):

PR #754, opened 2022-12-20
CORE-8988 POC for using JSON instead of custom schemas with UTXO ledger

mbrkic-r3:
- Added column "json_representation" to DB table "utxo_transaction_output".
- Added function "toJsonRepresentation" to ContractState.


ledger/ledger-utxo/src/main/kotlin/net/corda/v5/ledger/utxo/ContractState.kt
```diff
@@ -18,4 +18,6 @@ import java.security.PublicKey
 @CordaSerializable
 interface ContractState {
     val participants: List<PublicKey>
+
+    fun toJsonRepresentation(): String? = null
```
lankydan, line 22: I know this is a POC, but we might want to pass in a `JsonMarshallingService` into this. This means that users can define a separate POJO that represents the JSON representation of their state and can leverage the Jackson object mapper functionality inside the `JsonMarshallingService`.

My downside to that is naming though, in that `JsonMarshallingService` might want to become `JsonSerializer`, `JsonMapper` or `JsonService` instead since I'm not sure its "marshalling" anymore.

For now we can leave this just there for discussion.

PR #755, opened 2022-12-21
CORE-7009 Change UniquenessCheckResponse to LedgerUniquenessCheckResponse

nkovacsx:
### Overview

The client service for the uniqueness checker has been renamed to `LedgerUniquenessCheckerClientService`. However, the response it produces remained `UniquenessCheckResponse` which is quite misleading. This PR will rename it to `LedgerUniquenessCheckResponse` to align with the service's name.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -7,7 +7,7 @@ cordaProductVersion = 5.0.0
 # NOTE: update this each time this module contains a breaking change
 ## NOTE: currently this is a top level revision, so all API versions will line up, but this could be moved to
 ##   a per module property in which case module versions can change independently.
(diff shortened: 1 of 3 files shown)


relyafi (approved):
Looks good in isolation. I'm assuming we're intentionally not making a similar change to `UniquenessCheckResult` (and related classes), because there's nothing ledger specific about these?

nkovacsx:
@relyafi That's right. There needs to be some refactoring there too but that'll be in a different ticket, which requires potentially more work than this one so wanted to get this one done ASAP.

dimosr (approved):

PR #757, opened 2022-12-21
CORE-8859 Add TLS type to the Gateway Config to allow mutual TLS

williamvigorr3:


data/config-schema/src/main/resources/net/corda/schema/configuration/p2p.gateway/1.0/corda.p2p.gateway.json
```diff
@@ -25,25 +25,33 @@
...
           }
         },
+        "tlsType": {
+          "description": "How to validate TLS certificates in the gateway. This setting MUST be the same on all Gateways that need to communicate via HTTPS.",
+          "oneOf": [
+            {"const": "ONE_WAY", "title": "Only validate the TLS certificates for the gateway server."},
+            {"const": "MUTUAL", "title": "Validate the TLS certificates for both the gateway server and client."}
+          ],
+          "default": "ONE_WAY"
+        },
```
```suggestion
        "tlsType": {
          "description": "What kind of TLS connections to establish between the gateways in our cluster and gateways in other clusters. This setting MUST be the same on all Gateways that need to communicate via HTTPS.",
          "oneOf": [
            {"const": "ONE_WAY", "title": "Establishes a regular TLS connection, where the server gateway will be authenticated."},
            {"const": "MUTUAL", "title": "Establishes a mutual TLS connection, where both the server and client gateway are authenticated."}
          ],
          "default": "ONE_WAY"
        },
```
Suggestion to make the phrasing a bit more generic towards TLS connections, not just the certificates.

dimosr (approved):

PR #759, opened 2022-12-21
CORE-8610 - change primary key of cpk related tables

conal-smith:
Runtime OS change: https://github.com/corda/corda-runtime-os/pull/2804

Changing PK of CPK entities to be the CPK's file checksum rather than its name, version, signerSummaryHash.

Consequences of this change mean we no longer need to update existing CPK entities if we force upload a CPI that changes some CPKs, because they will have a different checksum therefore different row in CPK tables.

That means we don't need to track versions of CPKs in the audit trail, then reference these to get the content during rollback of migrations (force-resync REST endpoint). Instead, we just get the changelogs from `cpk_db_change_log` table as we can reference these by the checksum. When running migrations we tag the changelogs with the CPK file checksum they came from. If we are rolling back, we use this to find the changelogs we need to roll back.

The audit table becomes a simple append-only trail of CPK related changes. Possible future improvement of this table is to add some sort of "operationType" that records what happened in each audit row, for example, `FORCE_UPLOAD_CPI` or `UPLOAD_CPI`.

### DIFF
--- a/data/db-schema/src/main/resources/net/corda/db/schema/config/migration/cpx-creation-v1.0.xml
+++ b/data/db-schema/src/main/resources/net/corda/db/schema/config/migration/cpx-creation-v1.0.xml
@@ -127,10 +127,10 @@
             <column name="file_path" type="VARCHAR(255)">
                 <constraints nullable="false"/>
             </column>
-            <column name="content" type="TEXT">
+            <column name="changeset_id" type="UUID">
                 <constraints nullable="false"/>
             </column>
-            <column name="changeset_id" type="UUID">
+            <column name="content" type="TEXT">
                 <constraints nullable="false"/>
             </column>
             <!-- audit -->
@@ -146,7 +146,7 @@
             </column>
         </createTable>
 
-        <addPrimaryKey tableName="cpk_db_change_log" columnNames="cpk_file_checksum, file_path"
+        <addPrimaryKey tableName="cpk_db_change_log" columnNames="cpk_file_checksum, file_path, changeset_id"
                        constraintName="cpk_db_change_log_pk" schemaName="${schema.name}"/>
 
         <addForeignKeyConstraint baseTableName="cpk_db_change_log" baseColumnNames="cpk_file_checksum"
--- a/data/db-schema/src/main/resources/net/corda/db/schema/config/migration/cpx-creation-v1.0.xml
+++ b/data/db-schema/src/main/resources/net/corda/db/schema/config/migration/cpx-creation-v1.0.xml
@@ -156,6 +156,15 @@
                                  referencedTableSchemaName="${schema.name}"/>
         <!-- Append only audit table. Unlike cpk_db_change_logs which will be updated during CPI update process. -->
         <createTable tableName="cpk_db_change_log_audit" schemaName="${schema.name}">
+            <column name="changeset_id" type="UUID">
+                <constraints nullable="false"/>
+            </column>
+            <column name="cpk_file_checksum" type="VARCHAR(255)">
+                <constraints nullable="false"/>
+            </column>
+            <column name="file_path" type="VARCHAR(255)">
+                <constraints nullable="false"/>
+            </column>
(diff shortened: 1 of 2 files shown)


data/db-schema/src/main/resources/net/corda/db/schema/config/migration/cpx-creation-v1.0.xml
```diff
@@ -123,38 +111,26 @@
...
-            <column name="cpk_signer_summary_hash" type="VARCHAR(255)">
-                <constraints nullable="false"/>
-            </column>
             <column name="cpk_file_checksum" type="VARCHAR(255)">
                 <constraints nullable="false"/>
             </column>
             <column name="file_path" type="VARCHAR(255)">
                 <constraints nullable="false"/>
             </column>
-            <column name="content" type="TEXT">
```
sprintengine, line 154: I added changeset_id into the primary key of cpk_db_change_log.

Every time the liquibase changelogs are extracted from a CPI, new changelog entities are created with a unique changeset ID.

This may cause some duplication (i.e. if a CPI is force uploaded with the same CPKs as exists already, we'll get duplicate changelogs for those CPKs but with a different changesetId). This simplifies some of the rollback logic.

vkolomeyko (dismissed):
LGTM

But probably need another approval from @driessamyn or @josephzunigadaly

josephzunigadaly (dismissed):

data/db-schema/src/main/resources/net/corda/db/schema/config/migration/cpx-creation-v1.0.xml
driessamyn: Is this only during force upload?
Isn't this for each CPI upload? E.g. say you have CPI v1 that includes foo.cpk and bar.cpk, then CPI v2 that updates just foo.cpk (but leaves bar.cpk), will it not also create a new (duplicate) entry here?
It may be ok, just feels a bit odd.
A more natural model in this case may be to have a changeset table which keeps a link between changeset_id and cpk_file_checksum where 1 cpk_file_checksum could link to multiple changesets.

data/db-schema/src/main/resources/net/corda/db/schema/config/migration/cpx-creation-v1.0.xml
```diff
@@ -195,13 +165,16 @@
...
                 <constraints nullable="false"/>
             </column>
-            <column name="changeset_id" type="UUID">
+            <column name="cpi_version" type="VARCHAR(255)">
                 <constraints nullable="false"/>
             </column>
-            <column name="entity_version" type="INT">
+            <column name="cpi_signer_summary_hash" type="VARCHAR(255)">
+                <constraints nullable="false"/>
+            </column>
```
driessamyn: Do we need these? And do we not need the changeset id (instead)?

vkolomeyko (dismissed):
LGTM

data/db-schema/src/main/resources/net/corda/db/schema/config/migration/cpx-creation-v1.0.xml
sprintengine: Updated: I've removed changeset_id completely, we tag changelogs in liquibase with the CPk file checksum. When we need to rollback these changes, we simply get list of CPK file checksums we've applied, load the cpk_db_change_log entries for these checksums, and roll them back, then apply migrations for the new CPI's CPKs.

data/db-schema/src/main/resources/net/corda/db/schema/config/migration/cpx-creation-v1.0.xml
sprintengine: I removed the cpi related columns, and also the changeset id, audit table becomes more of an "append-only" history of changelogs and CPK file checksums.

driessamyn (approved):
LGTM

PR #769, opened 2023-01-01
CORE-8108 Add `SignatureSpec` field to `DigitalSignatureMetadata`

kyriathar:
- adds `SignatureSpec` field to `DigitalSignatureMetadata`.
- marks `SignatureSpec` as `CordaSerializable`.

runtime-os PR: [#2873](https://github.com/corda/corda-runtime-os/pull/2873)

### DIFF
--- a/crypto/src/main/kotlin/net/corda/v5/crypto/ParameterizedSignatureSpec.kt
+++ b/crypto/src/main/kotlin/net/corda/v5/crypto/ParameterizedSignatureSpec.kt
@@ -1,6 +1,5 @@
 package net.corda.v5.crypto
 
-import net.corda.v5.base.annotations.CordaSerializable
 import java.security.spec.AlgorithmParameterSpec
 
 /**
--- a/crypto/src/main/kotlin/net/corda/v5/crypto/ParameterizedSignatureSpec.kt
+++ b/crypto/src/main/kotlin/net/corda/v5/crypto/ParameterizedSignatureSpec.kt
@@ -1,5 +1,6 @@
 package net.corda.v5.crypto
 
+import net.corda.v5.base.annotations.CordaSerializable
 import java.security.spec.AlgorithmParameterSpec
 
 /**
@@ -14,6 +15,7 @@ import java.security.spec.AlgorithmParameterSpec
  * When used for signing the [signatureName] must match the corresponding key scheme, e.g. you cannot use
  * "SHA256withECDSA" with "RSA" keys.
  */
+@CordaSerializable
 class ParameterizedSignatureSpec(
     signatureName: String,
     val params: AlgorithmParameterSpec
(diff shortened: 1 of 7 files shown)


crypto/src/main/kotlin/net/corda/v5/crypto/ParameterizedSignatureSpec.kt
```diff
@@ -14,6 +15,7 @@ import java.security.spec.AlgorithmParameterSpec
  * When used for signing the [signatureName] must match the corresponding key scheme, e.g. you cannot use
  * "SHA256withECDSA" with "RSA" keys.
  */
+@CordaSerializable
 class ParameterizedSignatureSpec(
     signatureName: String,
     val params: AlgorithmParameterSpec
```
kyriathar: The instantiations of `ParameterizedSignatureSpec` we offer in:

https://github.com/corda/corda-api/blob/9cd61de16677a712fdebe41e2824ed46e6e4df78/crypto/src/main/kotlin/net/corda/v5/crypto/SignatureSpec.kt#L50

are using `PSSParameterSpec` as `AlgorithmParameterSpec`. Thanks to custom serializers for `AlgorithmParameterSpec`, `PSSParameterSpec` and `MGF1ParameterSpec` provided in complement runtime-os PR: https://github.com/corda/corda-runtime-os/blob/7f7e92a7ceb699eb6aae79679ad5214a39b23e6b/libs/serialization/serialization-amqp/src/main/kotlin/net/corda/internal/serialization/amqp/custom/ParameterizedSignatureSpecSerializers.kt

our  instantiations of `ParameterizedSignatureSpec` are `CordaSerializable`. However, if users instantiate their own `ParameterizedSignatureSpec` using some other implementation of `AlgorithmParameterSpec` then corda serializability will break. The way it currently is they could still go and define a custom serializer for that other `AlgorithmParameterSpec` implementation and it would work. But I believe once marked `CordaSerializable` it should always be corda serializable?

In our crypto codebase, in terms of  `ParameterizedSignatureSpec.params` we seem to only be expecting to find `PSSParameterSpec` as `AlgorithmParameterSpec`. Should we be therefore changing `params` to be `PSSParameterSpec`? Even then though `PSSParameterSpec.mgfSpec` is of type `AlgorithmParameterSpec` and we end up with the same above issue. @mnesbit could you please advice? Perhaps we could provide a copy type of our own holding the types we actually use i.e. a type to hold `PSSParameterSpec` info with `MGF1ParameterSpec` in place of `PSSParameterSpec.mgfSpec`?

mnesbit: We can't allow custom signature algorithms (beyond the ability of some algorithms to support a plugabble hash i.e. for ECDSA and RSA) so I think I am OK with this being only available for in-house platform supported algorithms. If we think we can't do custom hash+ECDSA with SignatureSpec then we would be best to make the algorithm parameters a MAP<String,String> to ensure flexibility and evolution support.

mnesbit (dismissed):
ok

kyriathar:
Registered [CORE-9662](https://r3-cev.atlassian.net/browse/CORE-9662) for offering signature specs via a factory API to users.  

[CORE-9662]: https://r3-cev.atlassian.net/browse/CORE-9662?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

dickon (approved):
LGTM (mostly going on Matthew's review)

driessamyn (approved):
LGTM

PR #770, opened 2023-01-03
CORE-8868: Add Holding identity to the Gateway TLS certificates

yift-r3:
Runtime PR in https://github.com/corda/corda-runtime-os/pull/2846

This change adds the holding identity to the Gateway TLS certificates (to be used as client certificates)

### DIFF
--- a/data/avro-schema/src/main/resources/avro/net/corda/data/p2p/GatewayTlsCertificates.avsc
+++ b/data/avro-schema/src/main/resources/avro/net/corda/data/p2p/GatewayTlsCertificates.avsc
@@ -9,6 +9,11 @@
       "type": "string"
     },
     {
+      "doc": "The holding identity that the certificate belongs to",
+      "name": "holdingIdentity",
+      "type": "net.corda.data.identity.HoldingIdentity"
+    },
+    {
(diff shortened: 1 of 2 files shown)


data/avro-schema/src/main/resources/avro/net/corda/data/p2p/GatewayTlsCertificates.avsc
```diff
@@ -8,6 +8,11 @@
       "name": "tenantId",
       "type": "string"
     },
+    {
+      "doc": "The holding identity that the certificate belongs to",
```
dimosr: maybe better to say here "that the certificate is used by"?

Since the certificate can be a cluster-level certificate, shared by many vnodes (i.e. not really belonging to any of them specifically).

dimosr: ~Should we also have a field indicating if it's a client or server certificate?~

actually scratch the second comment, IIRC I think we decided to have a single certificate for now.

dimosr (dismissed):

dimosr (approved):

PR #773, opened 2023-01-04
CORE-8875: Mutual TLS allowed subject list APIs

yift-r3:
* Add allow certificate subjects table
*  Add APIs to maintain the allowed certificates table

Runtime PR in https://github.com/corda/corda-runtime-os/pull/2859


data/avro-schema/src/main/resources/avro/net/corda/data/membership/rpc/response/MutualTlsListAllowedCertificatesResponse.avsc
```diff
@@ -0,0 +1,16 @@
+{
+  "type": "record",
+  "name": "MutualTlsListAllowedCertificatesResponse",
+  "namespace": "net.corda.data.membership.rpc.response",
+  "doc": "List of client certificate subject that had been allowed to be used in mutual TLS.",
```
charlieR3:
```suggestion
  "doc": "List of client certificate subjects that have been allowed to be used in mutual TLS.",
```

data/avro-schema/src/main/resources/avro/net/corda/data/membership/rpc/request/MembershipRpcRequest.avsc
```diff
@@ -16,7 +16,10 @@
         "RegistrationRpcRequest",
         "RegistrationStatusRpcRequest",
         "MGMGroupPolicyRequest",
-        "RegistrationStatusSpecificRpcRequest"
+        "RegistrationStatusSpecificRpcRequest",
+        "MutualTlsAllowCertificateRequest",
```
charlieR3: Not sure if better to comment on this PR or the runtime-os PR, but this change made me wonder if there's any point to wrap this as an RPC request to the membership worker? I think it's something we can review on some of the other membership operations such as querying for registration status. For these operations we're basically trying to interact with the database so it should be possible to use the membership persistence client from the RPC worker and go direct to the database instead of going from rpc worker -> membership worker -> db worker. It seems like the only thing the membership worker is doing is checking if it's a valid holding ID which we can do anywhere.

data/db-schema/src/main/resources/net/corda/db/schema/vnode-vault/migration/vnode-vault-creation-v1.0.xml
```diff
@@ -28,6 +28,13 @@
             <column name="usage"/>
         </createIndex>
 
+        <createTable tableName="vnode_allowed_certificate">
```
charlieR3: Will this have any usage outside of mutual TLS? If not, it could be good to rename this to something like "vnode_mtls_allowed_certificate"

charlieR3 (approved):

williamvigorr3 (approved):
LGTM

PR #775, opened 2023-01-04
CORE-8374 Remove default argument in `MerkleTreeFactory`

lankydan:


vlajos (dismissed):
I think there are some plans to use OSGI to find the the hash digest providers like with the serialisers, but I think this change is still useful.

blsemo (approved):
LGTM

PR #776, opened 2023-01-04
CORE-9012 Use custom Avro type for response to FindUnconsumedStatesByType

mbrkic-r3:
- Added avro messages for ledger transaction outputs

### DIFF
--- a/data/avro-schema/src/main/resources/avro/net/corda/data/ledger/persistence/TransactionOutput.avsc
+++ b/data/avro-schema/src/main/resources/avro/net/corda/data/ledger/persistence/TransactionOutput.avsc
@@ -1,0 +1,28 @@
+{
+  "type": "record",
+  "name": "TransactionOutput",
+  "doc": "Transaction output data of signed transaction.",
+  "namespace": "net.corda.data.ledger.persistence",
+  "fields": [
+    {
+        "name": "id",
+        "type": "string",
+        "doc": "The transaction ID, derived from the root hash of its Merkle tree"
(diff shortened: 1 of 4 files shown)


data/avro-schema/src/main/resources/avro/net/corda/data/ledger/persistence/TransactionOutput.avsc
```diff
@@ -0,0 +1,28 @@
+{
+  "type": "record",
+  "name": "TransactionOutput",
+  "doc": "Transaction output data of signed transaction.",
+  "namespace": "net.corda.data.ledger.persistence",
+  "fields": [
+    {
+        "name": "id",
+        "type": "string",
```
vlajos: Would not be net.corda.data.crypto.SecureHash better for this?

mbrkic-r3: I've kept it consistent with types we used so far (e.g. as in https://github.com/corda/corda-api/blob/release/os/5.0/data/avro-schema/src/main/resources/avro/net/corda/data/ledger/persistence/FindTransaction.avsc)

vlajos:
I think we should add utxo somewhere to the name/package.
These are quite UTXO specific.

mbrkic-r3:
> I think we should add utxo somewhere to the name/package. These are quite UTXO specific.

renamed

vlajos (approved):

PR #779, opened 2023-01-06
CORE-7995 Make sign and verify operation APIs more symmetric 

kyriathar:
- adds verify overload taking `DigitalSignature` instead of `ByteArray` as the signature to be more symmetric with sign operation which returns `DigitalSignature.WithKey` (extends `DigitalSignature`).

runtime-os PR: [#2875](https://github.com/corda/corda-runtime-os/pull/2875)

### DIFF
--- a/application/src/main/kotlin/net/corda/v5/application/crypto/DigitalSignatureVerificationService.kt
+++ b/application/src/main/kotlin/net/corda/v5/application/crypto/DigitalSignatureVerificationService.kt
@@ -28,5 +28,18 @@ interface DigitalSignatureVerificationService {
      */
     fun verify(publicKey: PublicKey, signatureSpec: SignatureSpec, signatureData: ByteArray, clearData: ByteArray)
 
+    /**
+     * Verifies a digital signature by using [signatureSpec].
+     * Always throws an exception if verification fails.
+     *
+     * @param publicKey The signer's [PublicKey].
+     * @param signatureSpec The signature spec.
+     * @param signature The signature on a message.
+     * @param clearData The clear data/message that was signed (usually the Merkle root).
+     *
+     * @throws CryptoSignatureException If verification of the digital signature fails.
+     * @throws IllegalArgumentException If the signature scheme is not supported or if any of the clear or signature
+     * data is empty.
+     */
     fun verify(publicKey: PublicKey, signatureSpec: SignatureSpec, signature: DigitalSignature, clearData: ByteArray)
 }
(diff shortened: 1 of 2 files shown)


application/src/main/kotlin/net/corda/v5/application/crypto/DigitalSignatureVerificationService.kt
```diff
@@ -26,4 +27,6 @@ interface DigitalSignatureVerificationService {
      * data is empty.
      */
     fun verify(publicKey: PublicKey, signatureSpec: SignatureSpec, signatureData: ByteArray, clearData: ByteArray)
+
```
mnesbit, line 31: Needs doc comments. Signature is fine

application/src/main/kotlin/net/corda/v5/application/crypto/DigitalSignatureVerificationService.kt
```diff
@@ -26,4 +27,6 @@ interface DigitalSignatureVerificationService {
      * data is empty.
      */
     fun verify(publicKey: PublicKey, signatureSpec: SignatureSpec, signatureData: ByteArray, clearData: ByteArray)
```
mnesbit, line 30: clearData is a terrible name. Perhaps dataBeingVerified, or originalData ?

application/src/main/kotlin/net/corda/v5/application/crypto/DigitalSignatureVerificationService.kt
kyriathar: kDoc added

application/src/main/kotlin/net/corda/v5/application/crypto/DigitalSignatureVerificationService.kt
kyriathar: Changed to `originalData`. 

Also did some parameters re-ordering to hopefully make it a bit more intuitive. I understand it should be replacing the other overload so it should be ok? It is now also more consistent with: 

`fun sign(bytes: ByteArray, publicKey: PublicKey, signatureSpec: SignatureSpec): DigitalSignature.WithKey`

Otherwise I will revert it.

kyriathar: Parameters order agreed offline. Also registered [CORE-9332](https://r3-cev.atlassian.net/browse/CORE-9332) to align existing `verify` overload with the one added now.

mnesbit (approved):
LGTM

PR #784, opened 2023-01-09
CORE-6173 Add notary keys to the member info

nkovacsx:
### Overview

Add a new property to the `MemberInfo` which is called `notaryKeys`. This will return the keys that belong to the given Notary VNode (**NOT the service**).

If the VNode is not a Notary VNode it will return an empty list as there are no keys registered under `corda.notary.keys`.

### DIFF
--- a/membership/src/main/kotlin/net/corda/v5/membership/MemberInfo.kt
+++ b/membership/src/main/kotlin/net/corda/v5/membership/MemberInfo.kt
@@ -47,6 +47,8 @@ import java.security.PublicKey
  * @property ledgerKeys List of current and previous (rotated) ledger keys, which member can still use to sign unspent
  * transactions on ledger.
  * Key at index 0 is always the latest added ledger key.
+ * @property notaryKeys List of current and previous (rotated) notary keys. Key at index 0 is always the latest added
+ * notary key. Might be an empty list if the member is not a notary VNode.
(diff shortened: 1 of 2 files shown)


membership/src/main/kotlin/net/corda/v5/membership/MemberInfo.kt
```diff
@@ -58,6 +60,7 @@ interface MemberInfo {
     val name: MemberX500Name
     val sessionInitiationKey: PublicKey
     val ledgerKeys: List<PublicKey>
+    val notaryKeys: List<PublicKey>
```
charlieR3, line 63: Does this have to be exposed on the public API? I think notary keys will only be exposed publicly through the Notary Service. This could be better as an extension property within the layer that requires it (e.g. the ledger layer). For example, we have other non-public fields defined in `MemberInfoExtension.kt` in the corda runtime project.

nkovacsx: Ah, that's a good point thanks @charlieR3. Probably we can get away with having this internally.

charlieR3: That'd be good to have as an extension then. Thanks 👍

PR #785, opened 2023-01-10
CORE-8860: Add TLS type to Mgm registration context

yift-r3:
Adds `corda.group.tls.type` to Mgm registration context

The runtime pull request is in https://github.com/corda/corda-runtime-os/pull/2890

### DIFF
--- a/data/membership-schema/src/main/resources/net/corda/schema/membership/mgm/registration/1.0/corda.mgm.registration.json
+++ b/data/membership-schema/src/main/resources/net/corda/schema/membership/mgm/registration/1.0/corda.mgm.registration.json
@@ -80,7 +80,7 @@
       ]
     },
     "corda.group.tls.type": {
-      "description": "The TLS type. Default to OneWay",
+      "description": "What kind of TLS connections to establish between the gateways in our cluster and gateways in other clusters. This setting MUST be the same on all Gateways that need to communicate via HTTPS. The default value is OneWay",
       "type": "string",
       "examples": [
         "Mutual",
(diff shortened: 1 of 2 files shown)


charlieR3 (dismissed):

data/membership-schema/src/main/resources/net/corda/schema/membership/mgm/registration/1.0/corda.mgm.registration.json
```diff
@@ -79,6 +79,14 @@
         "1.3"
       ]
     },
+    "corda.group.tls.type": {
+      "description": "The TLS type. Default to OneWay",
```
williamvigorr3: We could add more detail here. e.g. copy the docs from the gateway json schema.
```
          "description": "What kind of TLS connections to establish between the gateways in our cluster and gateways in other clusters. This setting MUST be the same on all Gateways that need to communicate via HTTPS.",
          "oneOf": [
            {"const": "ONE_WAY", "title": "Establishes a regular TLS connection, where the server gateway will be authenticated."},
            {"const": "MUTUAL", "title": "Establishes a mutual TLS connection, where both the server and client gateway are authenticated."}
```

yift-r3: I want to make it case insensitive, so I prefer not to put `oneOf` (as the rest of the doc).

williamvigorr3: I meant reformat it (I was being lazy and just copied the whole thing).

williamvigorr3: What you did looks great.

williamvigorr3 (approved):
LGTM

charlieR3 (approved):

PR #787, opened 2023-01-10
CORE-8016 - renaming 'http rpc' to 'rest' in corda-api

pmck91:
Initial work on renaming our 'http rpc' service to 'rest'

More information can be found in CORE-7040

please advise on the following file:
https://github.com/corda/corda-api/blob/e618fc70bb20237d39f6f416cba17631d509ce7e/data/topic-schema/src/main/resources/net/corda/schema/RPC.yaml

### DIFF
--- a/data/config-schema/src/main/kotlin/net/corda/schema/configuration/ConfigKeys.kt
+++ b/data/config-schema/src/main/kotlin/net/corda/schema/configuration/ConfigKeys.kt
@@ -20,7 +20,7 @@ object ConfigKeys {
     const val MEMBERSHIP_CONFIG = "corda.membership"
     const val SECURITY_CONFIG = "corda.security"
 
-    //  RPC
+    //  REST
     const val REST_ADDRESS = "address"
     const val REST_CONTEXT_DESCRIPTION = "context.description"
     const val REST_CONTEXT_TITLE = "context.title"
--- a/data/config-schema/src/main/kotlin/net/corda/schema/configuration/ConfigKeys.kt
+++ b/data/config-schema/src/main/kotlin/net/corda/schema/configuration/ConfigKeys.kt
@@ -13,7 +13,7 @@ object ConfigKeys {
     const val UTXO_LEDGER_CONFIG = "corda.ledger.utxo"
     const val P2P_LINK_MANAGER_CONFIG = "corda.p2p.linkManager"
     const val P2P_GATEWAY_CONFIG = "corda.p2p.gateway"
-    const val RPC_CONFIG = "corda.rpc"
+    const val REST_CONFIG = "corda.rest"
     const val SECRETS_CONFIG = "corda.secrets"
     const val SANDBOX_CONFIG = "corda.sandbox"
     const val RECONCILIATION_CONFIG = "corda.reconciliation"
@@ -21,15 +21,15 @@ object ConfigKeys {
     const val SECURITY_CONFIG = "corda.security"
 
     //  RPC
-    const val RPC_ADDRESS = "address"
-    const val RPC_CONTEXT_DESCRIPTION = "context.description"
-    const val RPC_CONTEXT_TITLE = "context.title"
-    const val RPC_ENDPOINT_TIMEOUT_MILLIS = "endpoint.timeoutMs"
-    const val RPC_MAX_CONTENT_LENGTH = "maxContentLength"
-    const val RPC_AZUREAD_CLIENT_ID = "sso.azureAd.clientId"
-    const val RPC_AZUREAD_CLIENT_SECRET = "sso.azureAd.clientSecret"
-    const val RPC_AZUREAD_TENANT_ID = "sso.azureAd.tenantId"
-    const val RPC_WEBSOCKET_CONNECTION_IDLE_TIMEOUT_MS = "websocket.idleTimeoutMs"
+    const val REST_ADDRESS = "address"
+    const val REST_CONTEXT_DESCRIPTION = "context.description"
+    const val REST_CONTEXT_TITLE = "context.title"
+    const val REST_ENDPOINT_TIMEOUT_MILLIS = "endpoint.timeoutMs"
+    const val REST_MAX_CONTENT_LENGTH = "maxContentLength"
+    const val REST_AZUREAD_CLIENT_ID = "sso.azureAd.clientId"
+    const val REST_AZUREAD_CLIENT_SECRET = "sso.azureAd.clientSecret"
+    const val REST_AZUREAD_TENANT_ID = "sso.azureAd.tenantId"
+    const val REST_WEBSOCKET_CONNECTION_IDLE_TIMEOUT_MS = "websocket.idleTimeoutMs"
 
     // Secrets Service
     const val SECRETS_PASSPHRASE = "passphrase"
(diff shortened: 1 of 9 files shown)


data/db-schema/src/main/resources/net/corda/db/schema/rbac/db.changelog-master.xml
```diff
@@ -2,5 +2,5 @@
 <databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
                    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
                    xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.5.xsd">
-    <include file="net/corda/db/schema/rbac/migration/rpc-rbac-creation-v1.0.xml"/>
+    <include file="net/corda/db/schema/rbac/migration/rest-rbac-creation-v1.0.xml"/>
```
sprintengine: RBAC isn't only for REST so this should be just "rbac-creation-v1.0.xml" IMO

But I wouldn't be against just reverting the rbac schema changes as they probably aren't top priority.

data/db-schema/src/main/resources/net/corda/db/schema/rbac/migration/rest-rbac-creation-v1.0.xml
```diff
@@ -3,15 +3,15 @@
...
 
-    <property name="schema.name" value="RPC_RBAC" global="false"/>
+    <property name="schema.name" value="REST_RBAC" global="false"/>
 
-    <changeSet author="vkolomeyko" id="rpc-perm-creation-v1.0">
+    <changeSet author="vkolomeyko" id="rest-perm-creation-v1.0">
 
         <sql>
-            CREATE SCHEMA IF NOT EXISTS RPC_RBAC;
+            CREATE SCHEMA IF NOT EXISTS REST_RBAC;
```
sprintengine: should be just "RBAC" IMO

data/db-schema/src/main/resources/net/corda/db/schema/rbac/migration/rest-rbac-creation-v1.0.xml
```diff
@@ -3,15 +3,15 @@
...
         </sql>
 
-        <createTable tableName="rpc_user" schemaName="${schema.name}">
+        <createTable tableName="rest_user" schemaName="${schema.name}">
```
sprintengine: Should be either "user" or "rbac_user" IMO.

Same for the rest of the changes to "rest_" because the rbac stuff isn't only rest- worker.

conal-smith (commented):
I think you could leave out the changes to the rbac schema as these probably aren't top priority, but also renaming them to "rest" isn't correct since they are cluster level tables for RBAC that are used in several workers. If you do want to go ahead with changing rbac schema I'd just remove the rpc prefix (and in some cases it might make sense to replace with "rbac_".

data/db-schema/src/main/resources/net/corda/db/schema/rbac/migration/rest-rbac-creation-v1.0.xml
vkolomeyko: Yes, I agree it should be just `RBAC`. (i.e. without REST prefix)
Since RBAC can be used in other C5 sub-systems.

data/db-schema/src/main/kotlin/net/corda/db/schema/DbSchema.kt
```diff
@@ -6,7 +6,7 @@ package net.corda.db.schema
  * For more information, please see [here](https://docs.liquibase.com/concepts/advanced/liquibase-schema-name-parameter.html).
  */
 object DbSchema {
-    const val RPC_RBAC = "RPC_RBAC"
+    const val REST_RBAC = "REST_RBAC"
```
vkolomeyko:
```suggestion
    const val RBAC = "RBAC"
```

data/db-schema/src/main/resources/net/corda/db/schema/rbac/db.changelog-master.xml
vkolomeyko:
```suggestion
    <include file="net/corda/db/schema/rbac/migration/rbac-creation-v1.0.xml"/>
```

data/db-schema/src/main/resources/net/corda/db/schema/rbac/migration/rest-rbac-creation-v1.0.xml
vkolomeyko:
```suggestion
        <createTable tableName="rbac_user" schemaName="${schema.name}">
```

vkolomeyko (changes requested):

emilybowe (dismissed):
LGTM bar the RBAC vs REST points made by Conal & Viktor

vkolomeyko:
Since this is a breaking change we need to increment the version of `corda-api` or else CI build will keep failing.
Also we need to have a Draft PR in `corda-runtime-os` repo which accomodates changes made here to `corda-api`.

data/config-schema/src/main/kotlin/net/corda/schema/configuration/ConfigKeys.kt
```diff
@@ -13,23 +13,23 @@ object ConfigKeys {
...
     const val P2P_GATEWAY_CONFIG = "corda.p2p.gateway"
-    const val RPC_CONFIG = "corda.rpc"
+    const val REST_CONFIG = "corda.rest"
     const val SECRETS_CONFIG = "corda.secrets"
     const val SANDBOX_CONFIG = "corda.sandbox"
     const val RECONCILIATION_CONFIG = "corda.reconciliation"
     const val MEMBERSHIP_CONFIG = "corda.membership"
     const val SECURITY_CONFIG = "corda.security"
 
     //  RPC
```
blsemo: Shouldn't that also be ``REST``?

data/config-schema/src/main/kotlin/net/corda/schema/configuration/ConfigKeys.kt
```diff
@@ -13,23 +13,23 @@ object ConfigKeys {
...
-    const val RPC_AZUREAD_TENANT_ID = "sso.azureAd.tenantId"
-    const val RPC_WEBSOCKET_CONNECTION_IDLE_TIMEOUT_MS = "websocket.idleTimeoutMs"
+    const val REST_ADDRESS = "address"
+    const val REST_CONTEXT_DESCRIPTION = "context.description"
+    const val REST_CONTEXT_TITLE = "context.title"
+    const val REST_ENDPOINT_TIMEOUT_MILLIS = "endpoint.timeoutMs"
+    const val REST_MAX_CONTENT_LENGTH = "maxContentLength"
+    const val REST_AZUREAD_CLIENT_ID = "sso.azureAd.clientId"
+    const val REST_AZUREAD_CLIENT_SECRET = "sso.azureAd.clientSecret"
+    const val REST_AZUREAD_TENANT_ID = "sso.azureAd.tenantId"
```
blsemo, line 31: I know it's not part of this change, but this azure specific config/code looks wrong to me...

vkolomeyko: This is should be fine as at the moment we only support Azure AD SSO.

data/config-schema/src/main/kotlin/net/corda/schema/configuration/ConfigKeys.kt
pmck91: good catch

vkolomeyko (approved):

PR #789, opened 2023-01-11
CORE-9335 Avro messages for ledger transaction contracts verification

mbrkic-r3:
Avro messages for ledger transaction contracts verification


data/avro-schema/src/main/resources/avro/net/corda/data/ledger/utxo/contract.verification/CordaPackageSummary.avsc
```diff
@@ -0,0 +1,22 @@
...
+    {
+      "name": "name",
+      "type": "string",
+      "doc": "Name of the package."
+    },
+    {
+      "name": "version",
+      "type": "string",
+      "doc": "Version of the package."
+    },
```
mnesbit: This isn't including signer summary hash which is technically required to disambiguate the name. I presume mostly we will lookup Cpk's for the sandbox by file hash and then confirm that this aligns with the name, version, signer metadata

data/avro-schema/src/main/resources/avro/net/corda/data/ledger/utxo/contract.verification/VerifyContractsRequest.avsc
```diff
@@ -0,0 +1,39 @@
...
+    },
+    {
+      "name": "holdingIdentity",
+      "type": "net.corda.data.identity.HoldingIdentity",
+      "doc": "The holding identity of the user making the request."
+    },
+    {
+      "name": "transaction",
+      "type": "bytes",
+      "doc": "The serialized transaction"
```
mnesbit: We may have to consider exactly what goes here. theoretically this should be a serialized ledger transaction with input StateAndRefs. My concern is that is a security error, because we will have already deserialized the data in the flow/DB sandboxes so the harm from malicious code is already done. The ideal would be the data is the untouched WireTransaction plus input state blobs so that the verification sandbox is the first thing to try to deserialize anything. However, I am not sure how that would fit with the current finality flow.

data/avro-schema/src/main/resources/avro/net/corda/data/ledger/utxo/contract.verification/VerifyContractsResponse.avsc
```diff
@@ -0,0 +1,13 @@
+{
+  "type": "record",
+  "name": "VerifyContractsResponse",
+  "doc": "The response to an {@link VerifyContractsRequest}.",
+  "namespace": "net.corda.ledger.utxo.contract.verification",
+  "fields": [
+    {
+      "name": "result",
+      "type": "boolean",
```
mnesbit: This might be a bit strict, though I agree right now we can only envisage a boolean response. Perhaps this should be an Enum with VERIFIED, INVALID for now and also add a message field to let us put some logging into flow worker.

Does this also need the TxId?

data/avro-schema/src/main/resources/avro/net/corda/data/ledger/utxo/contract.verification/VerifyContractsRequest.avsc
mbrkic-r3: This will be serialized `UtxoLedgerTransactionContainer` containing `WireTransaction`, `InputStateAndRefs` and `ReferenceStateAndRefs`:
https://github.com/corda/corda-runtime-os/blob/mbrkic-r3/CORE-9384/contract-verification-flow-processor/components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/verification/UtxoLedgerVerificationServiceImpl.kt#L46

data/avro-schema/src/main/resources/avro/net/corda/data/ledger/utxo/contract.verification/CordaPackageSummary.avsc
mbrkic-r3: added signer summary hash

data/avro-schema/src/main/resources/avro/net/corda/data/ledger/utxo/contract.verification/VerifyContractsResponse.avsc
mbrkic-r3: Replaced with enum and added message filed.

Flow side is sending transaction as an argument to verification, so it has TxId

vlajos:
Is the `.` intentional in the directories?

mbrkic-r3:
> Is the `.` intentional in the directories?

It was mistake, fixed now

data/avro-schema/src/main/resources/avro/net/corda/data/ledger/utxo/contract/verification/CordaPackageSummary.avsc
```diff
@@ -0,0 +1,27 @@
...
+      "doc": "Name of the package."
+    },
+    {
+      "name": "version",
+      "type": "string",
+      "doc": "Version of the package."
+    },
+    {
+      "name": "signerSummaryHash",
+      "type": ["null", "string"],
```
vlajos, line 18: data/avro-schema/src/main/resources/avro/net/corda/data/packaging/CpkIdentifier.avsc
has a different type for the same field.
```
      "type": ["null", "net.corda.data.crypto.SecureHash"]
```

data/avro-schema/src/main/resources/avro/net/corda/data/ledger/utxo/contract/verification/VerifyContractsRequest.avsc
```diff
@@ -0,0 +1,39 @@
...
+      "type": "net.corda.data.identity.HoldingIdentity",
+      "doc": "The holding identity of the user making the request."
+    },
+    {
+      "name": "transaction",
+      "type": "bytes",
+      "doc": "The serialized transaction"
+    },
+    {
+      "name": "cpkMetadata",
```
vlajos, line 26: Probably it would worth checking the terminology in data/avro-schema/src/main/resources/avro/net/corda/data/packaging
to be a bit more consistent.
I think cpkIdentifier is closer than cpkMetadata. (Although none of them matches exactly.)

blsemo (approved):

data/avro-schema/src/main/resources/avro/net/corda/data/ledger/utxo/contract/verification/VerifyContractsRequest.avsc
mbrkic-r3: This was designed to match [CordaPackageSummary](https://github.com/corda/corda-api/blob/release/os/5.0/ledger/ledger-common/src/main/kotlin/net/corda/v5/ledger/common/transaction/CordaPackageSummary.kt) that is stored in transaction metadata [TransactionMetadataImpl](https://github.com/corda/corda-runtime-os/blob/07199b9b2cbecae03e837ab80f8427899d4eaa32/libs/ledger/ledger-common-data/src/main/kotlin/net/corda/ledger/common/data/transaction/TransactionMetadataImpl.kt#L55) and same naming was used

data/avro-schema/src/main/resources/avro/net/corda/data/ledger/utxo/contract/verification/CordaPackageSummary.avsc
mbrkic-r3: This was designed to match [CordaPackageSummary](https://github.com/corda/corda-api/blob/release/os/5.0/ledger/ledger-common/src/main/kotlin/net/corda/v5/ledger/common/transaction/CordaPackageSummary.kt)

PR #793, opened 2023-01-12
CORE-8873: Allow membership worker to read the locally hosted identeties

yift-r3:
This PR should allow the membership worker to read from the locally hosted identities topic. It will be used to identify the identity client TLS certificate (to add the subject to the member info so it will be used in mutual TLS).

I think it should not break anything, so I didn't change the API version.


williamvigorr3 (approved):
LGTM

PR #794, opened 2023-01-12
CORE-7309 Api to find my signing keys

kyriathar:
- adds find my signing keys API on `SigningService` which takes a set of keys to look into for any keys owned by the caller. Returns a mapping from requested key to the same key if it is being owned by the caller or to null if key is not being owned by the caller. In case of composite key it maps the composite key to the firstly found composite key leaf.
- adds avro type for flow to crypto request to filter keys by key ids.

runtime-os PR: [#2912](https://github.com/corda/corda-runtime-os/pull/2912)

### DIFF
--- a/application/src/main/kotlin/net/corda/v5/application/crypto/SigningService.kt
+++ b/application/src/main/kotlin/net/corda/v5/application/crypto/SigningService.kt
@@ -38,5 +38,5 @@ interface SigningService {
     fun sign(bytes: ByteArray, publicKey: PublicKey, signatureSpec: SignatureSpec): DigitalSignature.WithKey
 
     @Suspendable
-    fun getMyKeys(keys: Set<PublicKey>): Set<PublicKey>
+    fun getMySigningKeys(keys: Set<PublicKey>): Set<PublicKey>
 }
--- a/application/src/main/kotlin/net/corda/v5/application/crypto/SigningService.kt
+++ b/application/src/main/kotlin/net/corda/v5/application/crypto/SigningService.kt
@@ -43,7 +43,7 @@ interface SigningService {
      * composite key leaf.
      *
      * @param keys The signing keys to look into.
-     * @return Mapping of requested signing keys to found signing keys to be owned by the caller or null if not owned.
+     * @return A mapping of requested signing keys to found signing keys to be owned by the caller or null if not found to be owned.
      */
     @Suspendable
     fun findMySigningKeys(keys: Set<PublicKey>): Map<PublicKey, PublicKey?>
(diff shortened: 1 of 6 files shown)


application/src/main/kotlin/net/corda/v5/application/crypto/SigningService.kt
```diff
@@ -36,4 +36,7 @@ interface SigningService {
      */
     @Suspendable
     fun sign(bytes: ByteArray, publicKey: PublicKey, signatureSpec: SignatureSpec): DigitalSignature.WithKey
+
+    @Suspendable
+    fun getMyKeys(keys: Set<PublicKey>): Set<PublicKey>
```
kyriathar: This will be changed to return a `Map<PublicKey, PublicKey>` mapping the input keys to the ones found to be "mine".

mnesbit: As discussed will return Map<PublicKey, PublicKey?> in fact where the keys are 1-to-1 with the requested set and the values are the PublicKeys (or leaf of composite keys) found to be locally available with a resident PrivateKey

application/src/main/kotlin/net/corda/v5/application/crypto/SigningService.kt
```diff
@@ -36,4 +36,14 @@ interface SigningService {
...
+
+    /**
+     * Gets a set of signing keys to look into and returns a mapping of the requested signing keys to found to be owned
+     * by the caller signing keys.
+     *
+     * @param keys The signing keys to look into.
+     * @return Mapping of requested signing keys to owned keys by the caller or null if not owned.
+     */
+    @Suspendable
+    fun getMySigningKeys(keys: Set<PublicKey>): Map<PublicKey, PublicKey?>
```
kyriathar: Maybe `filterMySigningKeys` is more intuitive name for this API.

kyriathar: Agreed offline to go with `findMySigningKeys`.

mnesbit (approved):
OK, though I think 12 character id is a bit short for key ids (so me may have to revisit)

kyriathar:
> OK, though I think 12 character id is a bit short for key ids (so me may have to revisit)

Raised [CORE-9550](https://r3-cev.atlassian.net/browse/CORE-9550) to keep track of it.

[CORE-9550]: https://r3-cev.atlassian.net/browse/CORE-9550?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #796, opened 2023-01-13
CORE-9000 Corda specific values in c5 are pulled out into a separate properties file cordaInternalDependencies.properties with different code owners

E1P3:
Updating the api version currently requires a codeowners review from a team lead this slows down development quite a bit. this is due to the review rules on gradle.properties.

A new properties file was created [here](https://github.com/corda/corda-runtime-os/blob/jansz/CORE-9000/new-properties-file/cordaInternalDependencies.properties), where we can house c5 specific properties with a different codeowners policy to C5

Change to be applied to [corda-runtime-os](https://github.com/corda/corda-runtime-os/pull/2916) and corda-api where the new cordaInternalDependecies.properties file can be created, any internal dependencies should be moved to there


build.gradle
```diff
@@ -3,9 +3,14 @@ import static org.gradle.api.JavaVersion.VERSION_11
 import static org.gradle.jvm.toolchain.JavaLanguageVersion.of
 
 buildscript {
-    ext {
+    Properties cordaInternalDependencies = new Properties()
+    file("$projectDir/cordaInternalDependencies.properties").withInputStream { InputStream input -> cordaInternalDependencies.load(input) }
+    ext{
```
ronanbrowne:
```suggestion
    ext {
```

build.gradle
```diff
@@ -3,9 +3,14 @@ import static org.gradle.api.JavaVersion.VERSION_11
...
+        cordaApiRevision = cordaInternalDependencies.getProperty("cordaApiVersion")
+        internalPublishVersion = cordaInternalDependencies.getProperty("internalPublishVersion")
+        cordaProductVersion = cordaInternalDependencies.getProperty("cordaProductVersion")
         vcsUrl = System.getenv('GIT_URL') ?: 'https://github.com/corda/corda-api.git'
-    }
+    }
```
ronanbrowne:
```suggestion
    }
```

PR #797, opened 2023-01-16
CORE-9041 - add KILLED to flowStates (flow status)

conal-smith:
This is not a breaking change ([counterpart API build proving this](https://github.com/corda/corda-runtime-os/pull/2928)).

Runtime-os related change - https://github.com/corda/corda-runtime-os/pull/2925

### DIFF
--- a/data/avro-schema/src/main/resources/avro/net/corda/data/flow/output/FlowStatus.avsc
+++ b/data/avro-schema/src/main/resources/avro/net/corda/data/flow/output/FlowStatus.avsc
@@ -51,9 +51,9 @@
       "doc": "Optional error message, this will be set if the flow status is 'FAILED'"
     },
     {
-      "name": "processingTerminationDetails",
-      "type": ["null", "net.corda.data.KeyValuePairList"],
-      "doc": "Optional details about processing termination, for example, reasons and additional info about why a flow was killed."
+      "name": "processingTerminatedReason",
+      "type": ["null", "string"],
+      "doc": "Optional message indicating reasoning why processing a flow has been terminated."
     },
     {
       "name": "createdTimestamp",
--- a/data/avro-schema/src/main/resources/avro/net/corda/data/flow/output/FlowStatus.avsc
+++ b/data/avro-schema/src/main/resources/avro/net/corda/data/flow/output/FlowStatus.avsc
@@ -52,13 +52,7 @@
     },
     {
       "name": "processingTerminationDetails",
-      "type": [
-        "null",
-        {
-            "type": "map",
-            "values": "string"
-        }
-      ],
+      "type": ["null", "net.corda.data.KeyValuePairList"],
       "doc": "Optional details about processing termination, for example, reasons and additional info about why a flow was killed."
     },
     {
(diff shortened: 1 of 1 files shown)


data/avro-schema/src/main/resources/avro/net/corda/data/flow/output/FlowStatus.avsc
```diff
@@ -49,6 +50,17 @@
...
+      "name": "processingTerminationDetails",
+      "type": [
+        "null",
+        {
+            "type": "map",
+            "values": "string"
+        }
+      ],
+      "doc": "Optional details about processing termination, for example, reasons and additional info about why a flow was killed."
+    },
```
sprintengine, line 57: I think it is useful to see on the flow status why a flow was killed / why processing was terminated.

I have not exposed this on the REST endpoint, it's just on the avro message.

vkolomeyko (dismissed):
LGTM
but will need input from @corda/flow-worker team.

data/avro-schema/src/main/resources/avro/net/corda/data/flow/output/FlowStatus.avsc
```diff
@@ -49,6 +50,17 @@
       "type": ["null", "net.corda.data.ExceptionEnvelope"],
       "doc": "Optional error message, this will be set if the flow status is 'FAILED'"
     },
+    {
+      "name": "processingTerminationDetails",
+      "type": [
+        "null",
+        {
+            "type": "map",
```
simon-johnson-r3: I'm pretty sure we cannot use Avro map types because the order in which it sends the entries back is non deterministic which means we cannot sign the messages to validate authenticity (i.e. we don't know the exact content in advance at the receiving end).

This has been solved already however, there's a reusable Corda type called `KeyValuePairList` and there are some helper methods in the codebase to help you interact with it in map like ways too. E.g.
https://github.com/corda/corda-api/blob/dc4a95b7728aa9b4ac84b135395459def407d068/data/avro-schema/src/main/resources/avro/net/corda/data/flow/event/external/ExternalEventContext.avsc#L18

simon-johnson-r3 (changes requested):
Blocking until map question is resolved.

data/avro-schema/src/main/resources/avro/net/corda/data/flow/output/FlowStatus.avsc
LWogan: Does this only apply to things which are stored in the checkpoint and input flow events? or did we make it a general rule just in case?
FlowStatus is only used by the RPC worker to report back status to the client.

LWogan: im wondering should I raise a ticket for FindWithNamedQuery

sprintengine: Thanks Simon I wasn't aware of this, I assume the signing is applicable to all messages we put on Kafka

simon-johnson-r3: I assume it applies to everything. @mnesbit or @JamesHR3 would have more information.

simon-johnson-r3 (dismissed):

vkolomeyko (dismissed):
LGTM

data/avro-schema/src/main/resources/avro/net/corda/data/flow/output/FlowStatus.avsc
sprintengine: I'm actually getting rid of the map, it's not necessary, a simple string is suffice

simon-johnson-r3 (approved):

PR #805, opened 2023-01-23
CORE-9566 Fix containsAny to not use intersect

nkovacsx:
### Overview

Before Kotlin 1.6, the intersect function was converting the parameter to a Set. 
This functionality was removed after 1.6 but it can still be applied, if the `convert_arg_to_set_in_removeAll` JVM property is set to true. 

Obviously the intersect function tries to access this property to see if it should apply the “old” logic. 

However, this JVM property cannot be accessed from flows (probably because of OSGi sandboxing) thus causing the flow to fail.
Please note that this can only be reproduced if the provided parameter is a CompositeKey because the else branch does not use intersect.

### How to reproduce

This issue was caught by @vlajos when running some new ledger tests from the `corda-e2e-tests` repository. For the first run, the network will only have one notary VNode so the notary service's key is a simple `PublicKey`. 

When running the tests for the second time, a new notary VNode is registered and the notary service's key will become a `CompositeKey`. This time the `if (this is CompositeKey) keys.intersect(otherKeys).isNotEmpty()` branch is executed and the `intersect` function is called and the tests fail with:
```
2023-01-19 10:30:55.265 [single-threaded-scheduled-executor-1] WARN  net.corda.flow.fiber.FlowFiberImpl {client_id=59653b4b-a4c0-4979-97af-5d78d189c50a, flow_id=b276b1ce-a688-4bc0-b491-4d19264ff6e0, vnode_id=0238E39C0C01} - Flow failed
java.lang.ExceptionInInitializerError: null
    at kotlin.collections.BrittleContainsOptimizationKt.safeToConvertToSet(BrittleContainsOptimization.kt:67) ~[?:?]
    at kotlin.collections.BrittleContainsOptimizationKt.convertToSetForSetOperationWith(BrittleContainsOptimization.kt:33) ~[?:?]
    at kotlin.collections.CollectionsKt__MutableCollectionsKt.retainAll(MutableCollections.kt:170) ~[?:?]
    at kotlin.collections.CollectionsKt___CollectionsKt.intersect(_Collections.kt:1674) ~[?:?]
    at net.corda.v5.crypto.CryptoUtils.containsAny(CryptoUtils.kt:100) ~[?:?]
    at net.corda.ledger.utxo.flow.impl.flows.finality.UtxoFinalityBase.verifyAndAddNotarySignature(UtxoFinalityBase.kt:78) ~[?:?]
    at net.corda.ledger.utxo.flow.impl.flows.finality.UtxoFinalityFlow.notarize(UtxoFinalityFlow.kt:182) ~[?:?]
    at net.corda.ledger.utxo.flow.impl.flows.finality.UtxoFinalityFlow.call(UtxoFinalityFlow.kt:50) ~[?:?]
    at net.corda.ledger.utxo.flow.impl.flows.finality.UtxoFinalityFlow.call(UtxoFinalityFlow.kt:22) ~[?:?]
    at net.corda.flow.application.services.FlowEngineImpl.subFlow(FlowEngineImpl.kt:52) ~[?:?]
    at net.corda.ledger.utxo.flow.impl.UtxoLedgerServiceImpl.finalize(UtxoLedgerServiceImpl.kt:101) ~[?:?]
    at net.cordapp.demo.obligation.workflow.CreateObligationFlow.call(CreateObligationFlow.kt:64) ~[?:?]
    at net.cordapp.demo.obligation.workflow.CreateObligationFlow.call(CreateObligationFlow.kt:28) ~[?:?]
    at net.corda.flow.application.services.FlowEngineImpl.subFlow(FlowEngineImpl.kt:52) ~[?:?]
    at net.cordapp.demo.obligation.workflow.CreateObligationFlow$Initiator.call(CreateObligationFlow.kt:116) ~[?:?]
    at net.corda.flow.fiber.RPCStartedFlow.invoke(RPCStartedFlow.kt:14) ~[?:?]
    at net.corda.flow.fiber.FlowFiberImpl.runFlow(FlowFiberImpl.kt:96) ~[?:?]
    at net.corda.flow.fiber.FlowFiberImpl.run(FlowFiberImpl.kt:65) ~[?:?]
    at net.corda.flow.fiber.FlowFiberImpl.run(FlowFiberImpl.kt:22) ~[?:?]
    at co.paralleluniverse.fibers.Fiber.run1(Fiber.java:1164) ~[?:?]
    at co.paralleluniverse.fibers.Fiber.exec(Fiber.java:860) ~[?:?]
    at co.paralleluniverse.fibers.RunnableFiberTask.doExec(RunnableFiberTask.java:100) ~[?:?]
    at co.paralleluniverse.fibers.RunnableFiberTask.run(RunnableFiberTask.java:91) ~[?:?]
    at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:515) ~[?:?]
    at java.util.concurrent.FutureTask.run(FutureTask.java:264) ~[?:?]
    at co.paralleluniverse.concurrent.util.ScheduledSingleThreadExecutor$ScheduledFutureTask.access$001(ScheduledSingleThreadExecutor.java:160) ~[?:?]
    at co.paralleluniverse.concurrent.util.ScheduledSingleThreadExecutor$ScheduledFutureTask.run(ScheduledSingleThreadExecutor.java:272) ~[?:?]
    at co.paralleluniverse.concurrent.util.ScheduledSingleThreadExecutor.work(ScheduledSingleThreadExecutor.java:119) ~[?:?]
    at java.lang.Thread.run(Thread.java:829) ~[?:?]
Caused by: java.security.AccessControlException: access denied ("java.util.PropertyPermission" "kotlin.collections.convert_arg_to_set_in_removeAll" "read")
```

### Test notes

Re-run the tests on the same combined worker multiple times, no issues after the fix.

### DIFF
--- a/crypto/src/main/kotlin/net/corda/v5/crypto/CryptoUtils.kt
+++ b/crypto/src/main/kotlin/net/corda/v5/crypto/CryptoUtils.kt
@@ -97,8 +97,11 @@ fun PublicKey.isFulfilledBy(otherKeys: Iterable<PublicKey>): Boolean =
  * [otherKeys] is a [CompositeKey], this function will not find a match.</i>
  */
 fun PublicKey.containsAny(otherKeys: Iterable<PublicKey>): Boolean {
-    return if (this is CompositeKey) otherKeys.any { keys.contains(it) }
-    else this in otherKeys
+    return if (this is CompositeKey) {
+        // `keys` is exported to a standalone value here to make sure it is not evaluated in every loop tick
+        val currentKeys = keys
+        otherKeys.any { currentKeys.contains(it) }
+    } else this in otherKeys
 }
 
 /** Returns the set of all [PublicKey]s of the signatures. */
--- a/crypto/src/main/kotlin/net/corda/v5/crypto/CryptoUtils.kt
+++ b/crypto/src/main/kotlin/net/corda/v5/crypto/CryptoUtils.kt
@@ -97,7 +97,7 @@ fun PublicKey.isFulfilledBy(otherKeys: Iterable<PublicKey>): Boolean =
  * [otherKeys] is a [CompositeKey], this function will not find a match.</i>
  */
 fun PublicKey.containsAny(otherKeys: Iterable<PublicKey>): Boolean {
-    return if (this is CompositeKey) keys.intersect(otherKeys).isNotEmpty()
+    return if (this is CompositeKey) otherKeys.any { keys.contains(it) }
     else this in otherKeys
 }
 
--- a/crypto/src/test/kotlin/net/corda/v5/crypto/CryptoUtilsTests.kt
+++ b/crypto/src/test/kotlin/net/corda/v5/crypto/CryptoUtilsTests.kt
@@ -11,8 +11,11 @@ import org.junit.jupiter.api.Assertions.assertArrayEquals
 import org.junit.jupiter.api.Assertions.assertEquals
 import org.junit.jupiter.api.Assertions.assertFalse
 import org.junit.jupiter.api.Test
+import org.junit.jupiter.api.assertAll
 import org.junit.jupiter.params.ParameterizedTest
 import org.junit.jupiter.params.provider.MethodSource
+import org.mockito.kotlin.doReturn
+import org.mockito.kotlin.mock
 import java.io.ByteArrayInputStream
 import java.security.MessageDigest
 import java.security.PublicKey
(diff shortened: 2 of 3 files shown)


crypto/src/main/kotlin/net/corda/v5/crypto/CryptoUtils.kt
```diff
@@ -97,7 +97,7 @@ fun PublicKey.isFulfilledBy(otherKeys: Iterable<PublicKey>): Boolean =
  * [otherKeys] is a [CompositeKey], this function will not find a match.</i>
  */
 fun PublicKey.containsAny(otherKeys: Iterable<PublicKey>): Boolean {
-    return if (this is CompositeKey) keys.intersect(otherKeys).isNotEmpty()
+    return if (this is CompositeKey) otherKeys.any { keys.contains(it) }
```
kyriathar: Would it make sense to add a `val keys = this.keys` above this so we don't rerun it on every lambda call?

kyriathar (dismissed):

crypto/src/main/kotlin/net/corda/v5/crypto/CryptoUtils.kt
kyriathar: Or disregard please, not overhead in still doing it anyway.

relyafi (dismissed):

driessamyn:
Change looks ok, @nkovacsx, but could you add a unit test that exercises the compositekey execution path?

crypto/src/main/kotlin/net/corda/v5/crypto/CryptoUtils.kt
blsemo: The keys extension for PublicKey (or the `leafkeys` val) should probably be lazy so we don't flatmap the children collection in a tight loop? If that was a hash set fixed for the runtime of this method, this implementation would be good. If `this.keys` has non-trivial runtime, less so.

blsemo (commented):
Looks fine to me

nkovacsx:
@blsemo I'm not entirely sure what is your suggestion here?

crypto/src/main/kotlin/net/corda/v5/crypto/CryptoUtils.kt
nkovacsx: I'm not entirely sure what your suggestion is. Could you elaborate further?

kyriathar: I think this is a good observation. But `leafKeys` implementation is internal so let's not worry about this here?

We should be getting full set of keys though to look into.

But maybe after all let's do a `val keys = this.keys` after all? Because `getKeys()` might delegate to `CompositeKey.getLeafKeys()` which might be doing whatever under the hood like recomputing same things.

driessamyn (dismissed):
Thanks for adding that test @nkovacsx 
Looks good for the change I requested.

crypto/src/main/kotlin/net/corda/v5/crypto/CryptoUtils.kt
nkovacsx: I ended up just exporting the `this.keys` to a `val`. Making `leafKeys` lazy is something we could discuss later on, not sure if it is possible though. What happens if the underlying keys change? Is that a regular thing in a composite key? In that case `lazy` is not an option because it will only be evaluated once.

blsemo (approved):
LGTM

kyriathar (approved):

PR #806, opened 2023-01-23
CORE-8954 Notary Batch Signatures

vlajos:
Introduce Batch Signature creation and verification through `TransactionSignatureService` with extending 
`DigitalSignatureAndMetadata` with an optional `MerkleProof` field to handle single and batch signatures the same way.

* And a few other supporting changes:
  * Add `MerkleProof.calculateRoot()` to `MerkleProof`s to support Batch Signature verification. (To let us extract the root which got signed from the proof.) 
  * Remove the now obsolete `verifyNotarySignature()` from `TransactionSignatureService`.
  * Change `TransactionMetadata.getDigestSettings()` to return String values.
  * Introduce `TransactionWithMetadata` to let us handle properties required for Batch Signing consistently.

Runtime-os: https://github.com/corda/corda-runtime-os/pull/2958

### DIFF
--- a/application/src/main/kotlin/net/corda/v5/application/crypto/DigitalSignatureAndMetadata.kt
+++ b/application/src/main/kotlin/net/corda/v5/application/crypto/DigitalSignatureAndMetadata.kt
@@ -10,7 +10,7 @@ import net.corda.v5.crypto.merkle.MerkleProof
  *
  * @property signature The signature that was applied.
  * @property metadata Attached [DigitalSignatureMetadata] for this signature.
- * @property proof Attached [MerkleProof] to support batch signatures.
+ * @property proof Attached [MerkleProof] if this is a batch signature.
  * @property by The [PublicKey] that created the signature.
  *
  * @constructor Creates a [DigitalSignatureAndMetadata].
--- a/application/src/main/kotlin/net/corda/v5/application/crypto/DigitalSignatureAndMetadata.kt
+++ b/application/src/main/kotlin/net/corda/v5/application/crypto/DigitalSignatureAndMetadata.kt
@@ -3,12 +3,14 @@ package net.corda.v5.application.crypto
 import java.security.PublicKey
 import net.corda.v5.base.annotations.CordaSerializable
 import net.corda.v5.crypto.DigitalSignature
+import net.corda.v5.crypto.merkle.MerkleProof
 
 /**
  * A wrapper over the signature output accompanied by signer's public key and signature metadata.
  *
  * @property signature The signature that was applied.
  * @property metadata Attached [DigitalSignatureMetadata] for this signature.
+ * @property proof Attached [MerkleProof] to support batch signatures.
  * @property by The [PublicKey] that created the signature.
  *
  * @constructor Creates a [DigitalSignatureAndMetadata].
@@ -16,7 +18,13 @@ import net.corda.v5.crypto.DigitalSignature
 @CordaSerializable
 data class DigitalSignatureAndMetadata(
     val signature: DigitalSignature.WithKey,
-    val metadata: DigitalSignatureMetadata
+    val metadata: DigitalSignatureMetadata,
+    val proof: MerkleProof?
 ) {
+    constructor(signature: DigitalSignature.WithKey, metadata: DigitalSignatureMetadata) : this(
+        signature,
+        metadata,
+        null
+    )
     val by: PublicKey = signature.by
 }
--- a/crypto/src/main/kotlin/net/corda/v5/crypto/merkle/MerkleProof.kt
+++ b/crypto/src/main/kotlin/net/corda/v5/crypto/merkle/MerkleProof.kt
@@ -41,9 +41,11 @@ interface MerkleProof {
      *
      * @param digest The tree's digest.
      *
-     * @returns Root hash of the tree or null if any details are incorrect.
+     * @returns Root hash of the tree.
+     *
+     * @throws MerkleProofRebuildFailureException if the calculation of the root hash failed.
      */
     fun calculateRoot(
         digest: MerkleTreeHashDigest
-    ): SecureHash?
+    ): SecureHash
 }
--- a/crypto/src/main/kotlin/net/corda/v5/crypto/merkle/MerkleProof.kt
+++ b/crypto/src/main/kotlin/net/corda/v5/crypto/merkle/MerkleProof.kt
@@ -35,4 +35,15 @@ interface MerkleProof {
         root: SecureHash,
         digest: MerkleTreeHashDigest
     ): Boolean
+
+    /**
+     * Rebuilds the [MerkleTree] from the [MerkleProof] and returns its root [SecureHash].
+     *
+     * @param digest The tree's digest.
+     *
+     * @returns Root hash of the tree or null if any details are incorrect.
+     */
+    fun calculateRoot(
+        digest: MerkleTreeHashDigest
+    ): SecureHash?
 }
--- a/ledger/ledger-common/src/main/kotlin/net/corda/v5/ledger/common/transaction/TransactionSignatureService.kt
+++ b/ledger/ledger-common/src/main/kotlin/net/corda/v5/ledger/common/transaction/TransactionSignatureService.kt
@@ -3,7 +3,6 @@ package net.corda.v5.ledger.common.transaction
 import net.corda.v5.application.crypto.DigitalSignatureAndMetadata
 import net.corda.v5.application.crypto.DigitalSignatureVerificationService
 import net.corda.v5.base.annotations.Suspendable
-import net.corda.v5.crypto.SecureHash
 import net.corda.v5.crypto.merkle.MerkleProof
 import java.security.PublicKey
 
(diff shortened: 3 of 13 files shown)


application/src/main/kotlin/net/corda/v5/application/crypto/DigitalSignatureAndMetadata.kt
```diff
@@ -3,20 +3,28 @@ package net.corda.v5.application.crypto
...
 import net.corda.v5.base.annotations.CordaSerializable
 import net.corda.v5.crypto.DigitalSignature
+import net.corda.v5.crypto.merkle.MerkleProof
 
 /**
  * A wrapper over the signature output accompanied by signer's public key and signature metadata.
  *
  * @property signature The signature that was applied.
  * @property metadata Attached [DigitalSignatureMetadata] for this signature.
+ * @property proof Attached [MerkleProof] to support batch signatures.
```
relyafi:
```suggestion
 * @property proof Attached [MerkleProof] if this is a batch signature.
```

crypto/src/main/kotlin/net/corda/v5/crypto/merkle/MerkleProof.kt
```diff
@@ -35,4 +35,15 @@ interface MerkleProof {
         root: SecureHash,
         digest: MerkleTreeHashDigest
     ): Boolean
+
+    /**
+     * Rebuilds the [MerkleTree] from the [MerkleProof] and returns its root [SecureHash].
+     *
+     * @param digest The tree's digest.
+     *
+     * @returns Root hash of the tree or null if any details are incorrect.
```
relyafi: Would it not make more sense for this to throw an exception? How would a caller know what exactly was wrong?

ledger/ledger-common/src/main/kotlin/net/corda/v5/ledger/common/transaction/TransactionSignatureService.kt
```diff
@@ -0,0 +1,57 @@
...
+ * TransactionSignatureService can be used to sign and verify transaction signatures.
+ * It supports both single and batch signatures.
+ */
+
+interface TransactionSignatureService {
+    /**
+     * Signs a transaction id with all the available keys.
+     *
+     * @param transactionId The transaction id to be signed.
+     * @param publicKeys The keys which should be tried to sign with.
```
relyafi:
```suggestion
     * @param publicKeys Public keys that correspond to the private keys which should be used to sign with.
```
Is this what we meant?

ledger/ledger-common/src/main/kotlin/net/corda/v5/ledger/common/transaction/TransactionWithMetadata.kt
```diff
@@ -0,0 +1,22 @@
+package net.corda.v5.ledger.common.transaction
+
+import net.corda.v5.base.annotations.DoNotImplement
+import net.corda.v5.crypto.SecureHash
+
+/**
+ * TransactionWithMetadata lets access metadata properties of the different transactions of different ledger implementations.
```
relyafi:
```suggestion
 * TransactionWithMetadata contains metadata properties of transactions common across different ledger implementations.
```

ledger/ledger-common/src/main/kotlin/net/corda/v5/ledger/common/transaction/TransactionSignatureService.kt
```diff
@@ -0,0 +1,57 @@
...
+     *
+     * @param transactions The transactions to be signed.
+     * @param publicKeys The keys which should be tried to sign with.
+     *
+     * @returns Resulting signatures.
+     *
+     * @throws TransactionNoAvailableKeysException If none of the keys are available.
+     */
+    @Suspendable
+    fun sign(
```
relyafi: I'd consider renaming this to `signBatch` - for me it is a bit too subtle to rely on being given a list of transactions, rather than a single transaction id.

ledger/ledger-common/src/main/kotlin/net/corda/v5/ledger/common/transaction/TransactionSignatureService.kt
```diff
@@ -0,0 +1,57 @@
...
+     *
+     * @param transactionId The transaction id to be signed.
+     * @param publicKeys The keys which should be tried to sign with.
+     *
+     * @returns Resulting signatures.
+     *
+     * @throws TransactionNoAvailableKeysException If none of the keys are available.
+     */
+    @Suspendable
+    fun sign(transactionId: SecureHash, publicKeys: Iterable<PublicKey>): List<DigitalSignatureAndMetadata>
```
relyafi: Is there a reason we are passing in just the transaction id here, rather than the full transaction as we do with the overload? Wouldn't this be more future-proof and consistent?

ledger/ledger-common/src/main/kotlin/net/corda/v5/ledger/common/transaction/TransactionSignatureService.kt
```diff
@@ -0,0 +1,57 @@
...
+
+    /**
+     * Signs a list of transactions with each the available keys.
+     * It creates one batch signature for each available keys.
+     * Then returns the signatures for each transaction with a [MerkleProof] to prove that they are included in the batch.
+     *
+     * @param transactions The transactions to be signed.
+     * @param publicKeys The keys which should be tried to sign with.
+     *
+     * @returns Resulting signatures.
```
relyafi:
```suggestion
     * @returns List of signatures for each supplied transaction. The list will always be of the same size and in the same order as the supplied [transactions]
```

People may assume this, but I think it's important to make this guarantee on the API. We should also ensure (in runtime-os) that our unit tests verify this. I've been bitten before with a similar list return interface in the C4 uniqueness provider where we inadvertently broke the ordering in a change.

relyafi (commented):
Generally looks ok for a POC. Ahead of the finished submission, I'd like to see the PR notes updated so that it provides a summary section, explaining with a written description the functional changes we are making and why, rather than simply listing the code changes we have made. There should also be test notes, even if this is only running existing tests + adding new ones, or referring to the runtime-os PR.

crypto/src/main/kotlin/net/corda/v5/crypto/merkle/MerkleProof.kt
vlajos: The original implementation came from @mnesbit . I modified it only as less as possible.
But yes, exceptions would be more developer friendly.
What do you think @mnesbit ?

application/src/main/kotlin/net/corda/v5/application/crypto/DigitalSignatureAndMetadata.kt
vlajos: Thanks. Applied.

ledger/ledger-common/src/main/kotlin/net/corda/v5/ledger/common/transaction/TransactionSignatureService.kt
vlajos: Nearly.
I wanted to highlight that it only tries to sign with those keys. It is possible that only a few of them available to sign with.
I modified your suggestion a bit to this:
```Public keys that correspond to the private keys which should be attempted to sign with.```

ledger/ledger-common/src/main/kotlin/net/corda/v5/ledger/common/transaction/TransactionWithMetadata.kt
vlajos: I have applied this.

ledger/ledger-common/src/main/kotlin/net/corda/v5/ledger/common/transaction/TransactionSignatureService.kt
vlajos: OK. I have renamed it.

ledger/ledger-common/src/main/kotlin/net/corda/v5/ledger/common/transaction/TransactionSignatureService.kt
vlajos: Originally I started with that approach. Then realized that metadata is not used at all, and `TransactionWithMetadata` just limits its usability. Also that time we were not too sure if ledger will be the final layer.
But this last argument is already clarified and I agree that it would be more consistent with `TransactionWithMetadata`.
So I changed it back to `TransactionWithMetadata`.

ledger/ledger-common/src/main/kotlin/net/corda/v5/ledger/common/transaction/TransactionSignatureService.kt
vlajos: Thanks. I applied it and added an `outer` before the list at the beginning of the second phrase.

crypto/src/main/kotlin/net/corda/v5/crypto/merkle/MerkleProof.kt
vlajos: We had a chat with @mnesbit .
The results:
> getting the root hash should not be nullable
> it is indeed an exception if you can't
> but the validate should remain non-throwing
> so make the one returning null an internal method
> and wrap with one that does comparison still nonthrowing and returning null
> the other throwing if it gets null indicator (don't throw nullpointerexception use something more meaningful and derived from CordaRuntimeException)
...
> so yes I agree with Ramzi, but don't want verify to start throwing as a result of the refactor

vlajos: I have updated calculateRoot to throw a newly introduced `MerkleProofRebuildFailureException` with some details.

relyafi (approved):
LGTM

nkovacsx (approved):

blsemo (approved):
LGTM

PR #807, opened 2023-01-23
CORE-8880: Publish client certificate subject list to Gateway

yift-r3:
Runtime pull request in https://github.com/corda/corda-runtime-os/pull/2969

* Adds a topic and a class to publish the list of allowed certificates from the link manager to the gateway.
* Change the MGM-allowed certificates list type to contain the group ID (as we have decided to allow more than one group)


data/topic-schema/src/main/kotlin/net/corda/schema/Schemas.kt
```diff
@@ -129,6 +129,7 @@ class Schemas {
             const val P2P_IN_TOPIC = "p2p.in"
             const val P2P_HOSTED_IDENTITIES_TOPIC = "p2p.hosted.identities"
             const val P2P_MGM_ALLOWED_CLIENT_CERTIFICATE_SUBJECTS = "p2p.mgm.allowed.client.certificate.subjects"
+            const val P2P_ALL_ALLOWED_CLIENT_CERTIFICATE_SUBJECTS = "p2p.all.allowed.client.certificate.subjects"
```
dimosr: how about the following (and drag it slightly below with the gateway topics)?
```suggestion
            const val GATEWAY_ALLOWED_CLIENT_CERTIFICATE_SUBJECTS = "gateway.allowed.client.certificate.subjects"
```
A lot of the topics where we forward data to the gateway are prefixed with that at the moment, so it'd be nice to keep it consistent.

data/avro-schema/src/main/resources/avro/net/corda/data/p2p/mtls/AllClientCertificateSubjects.avsc
```diff
@@ -0,0 +1,12 @@
+{
+  "type": "record",
+  "name": "AllClientCertificateSubjects",
```
dimosr: I'd suggest we move that into `net.corda.data.p2p.mtls.gateway` namespace/package and then rename the class into `AllowedCertificateSubject`. I suppose the clash is the reason that prevented you from using the same name..

dimosr: also, let's add a doc at the root level with some docs for the class, e.g. `Certificate subject to allow TLS connections from by the gateway`.

yift-r3: I renamed all three of them so the names will be more explicit...

data/avro-schema/src/main/resources/avro/net/corda/data/p2p/mtls/MemeberAllowedCertificateSubject.avsc
```diff
@@ -1,6 +1,7 @@
 {
   "type": "record",
-  "name": "ClientCertificateSubject",
+  "name": "MemeberAllowedCertificateSubject",
```
dimosr: small typo
```suggestion
  "name": "MemberAllowedCertificateSubject",
```

dimosr (dismissed):

dimosr (approved):

PR #808, opened 2023-01-23
CORE-9469 Use PublicKey instead of collection in uniqueness client service

nkovacsx:
### Overview

Since the Crypto API we are calling from the implementation of the `LedgerUniquenessCheckerClientService` can handle composite `PublicKey` as an input we no longer need to extract the leaf keys into a collection. 

We can simply can use the type of `PublicKey` because `CompositeKey` is a subtype of `PublicKey` and the Crypto implementation will do the rest.

Test notes have been added to the runtime-os PR: https://github.com/corda/corda-runtime-os/pull/2960

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -7,7 +7,7 @@ cordaProductVersion = 5.0.0
 # NOTE: update this each time this module contains a breaking change
 ## NOTE: currently this is a top level revision, so all API versions will line up, but this could be moved to
 ##   a per module property in which case module versions can change independently.
-cordaApiRevision = 624
+cordaApiRevision = 625
 
 # Main
 kotlinVersion = 1.7.21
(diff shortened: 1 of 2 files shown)


blsemo (dismissed):
LGTM

relyafi (dismissed):
Approved, subject to the following minor comments being addressed in the PR description:

In the PR notes overview:

_...can handle PublicKey as an input..._

I assume you meant to say a _composite_ PublicKey? This is an important detail as it's fundamental to the reason we are doing the change.

There are also no test notes in the PR description. It's probably fine to simply reference the runtime-os PR where these have been added.

relyafi (dismissed):

blsemo (dismissed):

relyafi (approved):

blsemo (approved):
Re-approved

PR #809, opened 2023-01-24
CORE-7276 argumentless sign

vlajos:
CORE-7276 Remove toSignedTransaction(PublicKey)s from TxBuilders

Runtime-os: https://github.com/corda/corda-runtime-os/pull/2981

### DIFF
--- a/ledger/ledger-consensual/src/main/kotlin/net/corda/v5/ledger/consensual/transaction/ConsensualTransactionBuilder.kt
+++ b/ledger/ledger-consensual/src/main/kotlin/net/corda/v5/ledger/consensual/transaction/ConsensualTransactionBuilder.kt
@@ -3,6 +3,7 @@ package net.corda.v5.ledger.consensual.transaction
 import net.corda.v5.base.annotations.DoNotImplement
 import net.corda.v5.ledger.consensual.ConsensualState
 import net.corda.v5.base.annotations.Suspendable
+import net.corda.v5.base.exceptions.CordaRuntimeException
 
 /**
  * Defines a builder for [ConsensualSignedTransaction]s.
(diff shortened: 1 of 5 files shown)


MrMatthewLayton (approved):
LGTM

MrMatthewLayton (dismissed):

MrMatthewLayton (approved):
LGTM

lankydan (approved):

ledger/ledger-consensual/src/main/kotlin/net/corda/v5/ledger/consensual/transaction/ConsensualTransactionBuilder.kt
```diff
@@ -38,24 +38,8 @@ interface ConsensualTransactionBuilder {
      *
      * @throws IllegalStateException when called a second time on the same object to prevent
      *      unintentional duplicate transactions.
+     * @throws CordaRuntimeException if none of the required keys are available to sign the transaction.
```
kyriathar, line 41: Feels like this should be of a more specific exception type? (Also doesn't feel like a runtime one)

kyriathar (approved):
Only a minor comment otherwise LGTM

ledger/ledger-consensual/src/main/kotlin/net/corda/v5/ledger/consensual/transaction/ConsensualTransactionBuilder.kt
vlajos: I am a bit uncertain about this one.
We should not have SignedTxs without signatures.
But the TransactionSIgnature service should not fail if there are no keys. That can just return an empty list.
So I ended up testing if any signatures were generated in the SignedTxFactories and throwing this exception if there were none.

PR #811, opened 2023-01-25
CORE-9382 Retries when CPKs not available

mbrkic-r3:
Added Avro record `VerifyContractsRequestRedelivery` as redelivery state of `VerifyContractsRequest` for implementing retries when CPKs required by Verification Sandbox are not available.


mbrkic-r3:
This is not needed anymore as it was decided that retries will be implemented using transient exceptions.

PR #812, opened 2023-01-25
CORE-9631: New APIs for addSignatories and receiveFinality

MrMatthewLayton:
Added `vararg` overload for `addSignatories` to `UtxoTransactionBuilder`, and added overload for `receiveFinality` which does not require a `UtxoTransactionValidator` where transaction checking is not required.

### DIFF
--- a/ledger/ledger-utxo/src/main/kotlin/net/corda/v5/ledger/utxo/UtxoLedgerService.kt
+++ b/ledger/ledger-utxo/src/main/kotlin/net/corda/v5/ledger/utxo/UtxoLedgerService.kt
@@ -4,12 +4,12 @@ import net.corda.v5.application.messaging.FlowSession
 import net.corda.v5.base.annotations.DoNotImplement
 import net.corda.v5.base.annotations.Suspendable
 import net.corda.v5.crypto.SecureHash
-import net.corda.v5.ledger.utxo.transaction.filtered.UtxoFilteredTransaction
-import net.corda.v5.ledger.utxo.transaction.filtered.UtxoFilteredTransactionBuilder
 import net.corda.v5.ledger.utxo.transaction.UtxoLedgerTransaction
 import net.corda.v5.ledger.utxo.transaction.UtxoSignedTransaction
-import net.corda.v5.ledger.utxo.transaction.UtxoTransactionValidator
 import net.corda.v5.ledger.utxo.transaction.UtxoTransactionBuilder
+import net.corda.v5.ledger.utxo.transaction.UtxoTransactionValidator
+import net.corda.v5.ledger.utxo.transaction.filtered.UtxoFilteredTransaction
+import net.corda.v5.ledger.utxo.transaction.filtered.UtxoFilteredTransactionBuilder
 
 /**
  * Defines UTXO ledger services.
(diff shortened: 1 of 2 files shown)


ledger/ledger-utxo/src/main/kotlin/net/corda/v5/ledger/utxo/UtxoLedgerService.kt
```diff
@@ -98,10 +98,19 @@ interface UtxoLedgerService {
...
+     * Verifies, signs and records a [UtxoSignedTransaction].
+     *
+     * [receiveFinality] should be called in response to [finalize].
+     *
+     * @param session The [FlowSession] to receive the [UtxoSignedTransaction] from.
+     * @return The fully signed [UtxoSignedTransaction] that was received and recorded.
+     * @throws ContractVerificationException If the transaction failed contract verification.
+     */
+    @Suspendable
+    fun receiveFinality(session: FlowSession): UtxoSignedTransaction
```
vlajos, line 113: Could we highlight somehow that this method is not really supposed to be used generally and using this instead of the other similarly named one can lead to bad things.

lankydan: Yeah, I think that is a good idea. I understand the reasoning behind adding this overload but we shouldn't encourage people to use it. @blsemo and maybe @mattbradburyr3 thoughts?

blsemo: From reading the kDoc comments on this function and the following, I am not sure it's totally clear what the difference is - this clearly needs to be worked out in detail, i.e. that the one below (with a validator) allows you to apply business logic to an otherwise valid transaction to say you don't like its content.
Am I right in assuming that this new function is mainly for Java developers so they don't have to implement a no-op class implementing the `Validator` interface? Because for kotlin devs, I don't see why just adding a comment like 
"You can just pass `{}` as validator if there is not further business logic validation to be done on the transaction" would be good enough.

MrMatthewLayton: The issue I found was that instead of adding `{}` I saw a Kotlin implementation by one of our developers defined like this:

```kotlin
class Responder : SubFlow<UtxoSignedTransaction>, UtxoTransactionValidator {

    @Suspendable
    override fun call(session: FlowSession): SignedTransaction {
        return receiveFinality(session, this)
    }

    @Suspendable
    override fun checkTransaction(transaction: UtxoLedgerTransaction) {
    }
}
```

Using `{}` would obviously be clearer, but notice that the above code still isn't performing any additional checks. There are many cases where a responder needs to receive a transaction, but doesn't need to check anything.

Does Java/Kotlin offer any annotation which would raise an IDE warning to the developer; i.e. something like

```kotlin
@Warning("Maybe you want to check something in your transaction?")
fun receiveFinality(session: FlowSession) { ... }
```

Another interesting question - consider the following...

```kotlin
class ResponderFlow<UtxoSignedTransaction> : SubFlow<SignedTransaction>, UtxoTransactionValidator {

    @Suspendable
    override fun checkTransaction(transaction: UtxoLedgerTransaction) { ... }
```
`checkTransaction` here **is suspendable**

```kotlin
class ResponderFlow<UtxoSignedTransaction> : SubFlow<SignedTransaction>, UtxoTransactionValidator {

    @Suspendable
    override fun call(session: FlowSession) {
        return receiveFinality(session) {}
    }
```

Is `{}` suspendable here?

blsemo: `{}` doens't need to be suspendable - only functions that call other suspendable functions or actually suspend need to be suspendable.

MrMatthewLayton:
@vkolomeyko This won't get done, so closing this out.

PR #813, opened 2023-01-26
CORE-9732 - Deprecate `contextLogger` before removing it.

driessamyn:
Deprecate `contextLogger` as it shouldn't be in the API library.

runtime-os change: https://github.com/corda/corda-runtime-os/pull/2996

### DIFF
--- a/base/src/main/kotlin/net/corda/v5/base/util/KotlinUtils.kt
+++ b/base/src/main/kotlin/net/corda/v5/base/util/KotlinUtils.kt
@@ -28,12 +28,20 @@ infix fun Long.exactAdd(b: Long): Long = Math.addExact(this, b)
  *
  * `private val log = loggerFor<MyClass>()`
  */
+@Deprecated(
+    "Will be removed from api library",
+    ReplaceWith("LoggerFactory.getLogger(this::class.java)", "org.slf4j.LoggerFactory")
+)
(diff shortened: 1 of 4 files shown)


vkolomeyko (approved):
LGTM

base/src/main/kotlin/net/corda/v5/base/util/KotlinUtils.kt
```diff
@@ -28,12 +28,20 @@ infix fun Long.exactAdd(b: Long): Long = Math.addExact(this, b)
  *
  * `private val log = loggerFor<MyClass>()`
  */
+@Deprecated(
+    "Will be removed from api library",
+    ReplaceWith("LoggerFactory.getLogger(this::class.java)", "org.slf4j.LoggerFactory")
+)
 inline fun <reified T : Any> loggerFor(): Logger = LoggerFactory.getLogger(T::class.java)
```
lankydan, line 35: I thought `loggerFor` was ok. I don't mind if we want to remove it though.

lankydan: Agreed that we'd remove it.

lankydan (approved):

dimosr (approved):

PR #814, opened 2023-01-26
CORE-9338 add chunk key to allow chunks to have unique keys. make optional fields nullable

LWogan:
Removal of the fields made nullable in this pr will be handled by CORE-9481


data/avro-schema/src/main/resources/avro/net/corda/data/chunking/Chunk.avsc
```diff
@@ -29,7 +32,10 @@
     },
     {
       "name": "offset",
-      "type": "long",
+      "type": [
+        "null",
+        "long"
+      ],
```
driessamyn: Why does the offset need to be nullable?

LWogan: its specific to writing to a file. It has no use in the message bus integration or the flow session integration. I've a JIRA ticket to remove filename and offset and set them in the chunk property map instead.

driessamyn: I understand the filename one, but isn't the offset just an indication of where the chunk is in the list of bytes? Isn't that completely generic for all chunking usecases?
(i.e. it allows a more efficient way to re-constitute the chunks regardless of whether it's a file, db or in-memory byte array)

LWogan: Ye I see your point. I haven't been tracking or using the bytes position in the message bus piece though, so it feels like it would be a bit of a no-op to set it.

driessamyn: It's cheap to set as it's just an index that is available as you go through the stream when creating the chunks, and I suspect that it's always going to be more efficient to use this when putting the chunks back together on the other end, so I don't really see a downside in setting it.

LWogan: ive kept the changes minimal so far to the existing code. I've tried to avoid any refactoring the chunking stuff which sends chunks individually as it generates them and returns an id of the chunks. i dont bother with a stream as I need to avro serialize the data and it returns the full ByteArray. So then split it into chunk sized arrays. the offset will always be the bytearray size so i think i can set it based on that.

gradle.properties
```diff
@@ -7,7 +7,7 @@ cordaProductVersion = 5.0.0
 # NOTE: update this each time this module contains a breaking change
 ## NOTE: currently this is a top level revision, so all API versions will line up, but this could be moved to
 ##   a per module property in which case module versions can change independently.
-cordaApiRevision = 626
+cordaApiRevision = 627
```
driessamyn: This will need a bump, otherwise LGTM.

jujoramos (dismissed):
LGTM

conal-smith (dismissed):
lgtm

mbrkic-r3:
Why don't we have a flag that indicates it's the last chunk? What is the benefit of sending additional empty chunk (which could be considerable overhead for small messages)?

data/avro-schema/src/main/resources/avro/net/corda/data/chunking/Chunk.avsc
mbrkic-r3: For uses cases like saving chunked data to file, it would be good to have original data size as well, then we could save chunks directly to file as they come and considerably lower memory footprint for big files.

driessamyn: @mbrkic-r3 we can't necessarily do this as we could be streaming chunks from the origin (e.g. the HTTP file upload). 
But I agree, if it is available, it could be put into the property map and used as an optimisation.

JamesHR3 (approved):

PR #816, opened 2023-01-26
CORE-9451: fix config field name

dickon:
Align the two different places we describe  the configuration key for crypto. We need this change to allow the crypto configuration to be overridable via the boot configuration mechanism.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -7,7 +7,7 @@ cordaProductVersion = 5.0.0
 # NOTE: update this each time this module contains a breaking change
 ## NOTE: currently this is a top level revision, so all API versions will line up, but this could be moved to
 ##   a per module property in which case module versions can change independently.
-cordaApiRevision = 626
+cordaApiRevision = 627
 
 # Main
 kotlinVersion = 1.7.21
--- a/crypto/src/main/kotlin/net/corda/v5/crypto/exceptions/CryptoException.kt
+++ b/crypto/src/main/kotlin/net/corda/v5/crypto/exceptions/CryptoException.kt
@@ -4,11 +4,11 @@ import net.corda.v5.base.annotations.CordaSerializable
 import net.corda.v5.base.exceptions.CordaRuntimeException
 
 /**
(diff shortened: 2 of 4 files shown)


gradle.properties
```diff
@@ -7,7 +7,7 @@ cordaProductVersion = 5.0.0
 # NOTE: update this each time this module contains a breaking change
 ## NOTE: currently this is a top level revision, so all API versions will line up, but this could be moved to
 ##   a per module property in which case module versions can change independently.
-cordaApiRevision = 626
+cordaApiRevision = 627
```
kyriathar, line 10: I think this PR doesn't include API breaking changes.

dickon: Without this change I can't actually pick up boot config parameter for crypto, which is in a sense a breaking change.

driessamyn: I agree it's a breaking change because `runtime-os` uses this constant for when inserting config, so without a version change somebody would pick this up and it would break an already created DB.

If we're cleaning up, I wonder if an `insert` for this key is missing from `data/db-schema/src/main/resources/net/corda/db/schema/config/migration/config-creation-v1.0.xml`
In theory it shouldn't matter as we, currently, always require the DB to be "primed" with the master key, but it would be more consistent, and if we were to move to a model where the key is passed as boot configuration, then this wouldn't require a further DB change.

Also, BTW, I think we should wait to merge this until the runtime-os changes are also in a PR as we don't want unnecessary churn on this repo.

dickon: Left open for

> If we're cleaning up, I wonder if an insert for this key is missing from data/db-schema/src/main/resources/net/corda/db/schema/config/migration/config-creation-v1.0.xml
In theory it shouldn't matter as we, currently, always require the DB to be "primed" with the master key, but it would be more consistent, and if we were to move to a model where the key is passed as boot configuration, then this wouldn't require a further DB change.

which feels like a separate change, but I agree with minimising API churn to may be good to combine.

PR #819, opened 2023-01-27
CORE-9728 Add UtxoLedgerTransaction.notary, UtxoLedgerTransaction.metadata

vlajos:
CORE-9728 Add UtxoLedgerTransaction.notary, UtxoLedgerTransaction.metadata

### DIFF
--- a/ledger/ledger-utxo/src/main/kotlin/net/corda/v5/ledger/utxo/transaction/UtxoLedgerTransaction.kt
+++ b/ledger/ledger-utxo/src/main/kotlin/net/corda/v5/ledger/utxo/transaction/UtxoLedgerTransaction.kt
@@ -2,6 +2,8 @@ package net.corda.v5.ledger.utxo.transaction
 
 import net.corda.v5.base.annotations.DoNotImplement
 import net.corda.v5.crypto.SecureHash
+import net.corda.v5.ledger.common.Party
+import net.corda.v5.ledger.common.transaction.TransactionMetadata
 import net.corda.v5.ledger.utxo.Attachment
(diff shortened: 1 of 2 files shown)


mbrkic-r3 (approved):
LGTM

lankydan (approved):

ledger/ledger-utxo/src/main/kotlin/net/corda/v5/ledger/utxo/transaction/UtxoLedgerTransaction.kt
```diff
@@ -36,6 +38,16 @@ interface UtxoLedgerTransaction {
...
 
+    /**
+     * @property notary The notary used for notarising this transaction.
+     */
+    val notary: Party
+
+    /**
+     * @property metadata The transaction's metadata.
+     */
+    val metadata: TransactionMetadata
```
lankydan, line 49: Should we add the metadata to the consensual one as well?

vlajos: It seemed to be less useful for me, since ledgerTx verify is only called from two places:
1. ConsensualSignedTransactionFactoryImpl which verifies metadata already before creating the signed Tx.
2. ConsensualFinalityBase where the signedTransaction is also available.

Also the ConsensualLedgerTransaction API is quite restricted at the moment. I did not want to expose anything just for code simplicity.

But tbh I do not see a big difference, so happy to add it there if you think that is better.

PR #820, opened 2023-01-27
CORE-9486: api change - get cpk from checkpoint

emilybowe:
**Description**
Adds a cpk array property to the Pipeline State
required for this PR -> https://github.com/corda/corda-runtime-os/pull/3017/


data/avro-schema/src/main/resources/avro/net/corda/data/flow/state/checkpoint/PipelineState.avsc
```diff
@@ -20,6 +20,14 @@
...
       "default": null,
       "doc": "Used for platform generated errors reported back to user code."
+    },
+    {
+      "name": "cpks",
+      "type": {
+        "type": "array",
+        "items": "net.corda.data.crypto.SecureHash"
+      },
+      "doc": "Array of stored cpks from the Virtual Node."
```
LWogan: These are hashes of the CPKs right?  Might be better to call them cpkHashes/cpkChecksums or something like that to not confuse them with actual CPKs. The doc should be updated to reflect that too,

conal-smith (dismissed):
Yes this look good

JamesHR3 (approved):

PR #824, opened 2023-01-27
CORE-9578: Create topic for Worker info (version heartbeats)

fowlerrr:
Create the topic for workers to send heartbeat information.  To start, this will be version information, but I'm leaving the topic more generic in case we find other use for these messages.

### DIFF
--- a/data/topic-schema/src/main/resources/net/corda/schema/Cluster.yaml
+++ b/data/topic-schema/src/main/resources/net/corda/schema/Cluster.yaml
@@ -1,0 +1,18 @@
+topics:
+  ClusterWorkerInfo:
+    name: cluster.worker.info
+    consumers:
+      - crypto
+      - db
+      - flow
+      - membership
+      - link-manager
+      - rpc
+    producers:
+      - crypto
+      - db
+      - flow
+      - membership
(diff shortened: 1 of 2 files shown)


data/topic-schema/src/main/resources/net/corda/schema/Cluster.yaml
```diff
@@ -0,0 +1,18 @@
+topics:
+  ClusterWorkerInfo:
```
driessamyn, line 2: Shouldn't this be a compacted topic?

fowlerrr: I was expecting it to be PubSub.  As I understand it, P2P was going to need it to know the minimum version on the system and Compacted would hide if some worker was somehow unavailable (turned off or decommissioned),

Though, maybe that's not a terribly likely scenario for the workers?

driessamyn: I assumed the worker ID would maybe be part of the key. 
If it's pub/sub, how would a new consumer get the full list?
What's the schema for the message on the topic?

PR #825, opened 2023-01-30
CORE-9577: Add a record for publishing worker info

fowlerrr:
This message will, initially, be for transmitting version information.  But we might wish to leverage it for other info so I've left the name generic

### DIFF
--- a/data/avro-schema/src/main/resources/avro/net/corda/data/cluster/WorkerInfo.avsc
+++ b/data/avro-schema/src/main/resources/avro/net/corda/data/cluster/WorkerInfo.avsc
@@ -1,0 +1,18 @@
+{
+  "type": "record",
+  "name": "WorkerInfo",
+  "namespace": "net.corda.data.cluster",
+  "doc": "Version information for a worker in the cluster",
+  "fields": [
+    {
+      "name": "workerPlatformVersion",
+      "type": "int",
+      "doc": "the platform version the worker has been built against"
+    },
+    {
+      "name": "workerSoftwareVersion",
+      "type": "string",
+      "doc": "the string representation of the Corda version the worker is running"
+    }
+  ]
+}
--- a/data/avro-schema/src/main/resources/avro/net/corda/data/cluster/WorkerInfo.avsc
+++ b/data/avro-schema/src/main/resources/avro/net/corda/data/cluster/WorkerInfo.avsc
@@ -2,7 +2,7 @@
   "type": "record",
   "name": "WorkerInfo",
   "namespace": "net.corda.data.cluster",
-  "doc": "Version information for a worker in the cluster",
+  "doc": "General information for a worker in the cluster",
   "fields": [
     {
       "name": "workerPlatformVersion",


simon-johnson-r3 (approved):

data/avro-schema/src/main/resources/avro/net/corda/data/cluster/WorkerInfo.avsc
```diff
@@ -0,0 +1,18 @@
+{
+  "type": "record",
+  "name": "WorkerInfo",
+  "namespace": "net.corda.data.cluster",
+  "doc": "General information for a worker in the cluster",
+  "fields": [
```
driessamyn, line 6: Should this include the worker type and the instance ID?

Would it be useful to have a timestamp on the message?

Do we have a design that describes how this would work? I'm curious on how/when we get consensus around what the cluster's version(s) is/are, in particularly in a scenario where workers appear/disappear.
How often does a worker publish this data?

driessamyn (changes requested):
I think we need a design doc before implementing this.

data/avro-schema/src/main/resources/avro/net/corda/data/cluster/WorkerInfo.avsc
fowlerrr: Maybe the worker type and instance id.  In the PoC I'd used the worker type as the key to the topic.  But this is really just a draft until we know what information is necessary.

And, no, we don't have much of a design yet.  The requirement for _knowing_ the versions comes from @dimosr  and P2P as a function of the discussion between clusters and their versions.

As for how often, I expect it to be configurable, but probably defaulted to a second.

I'm happy to put up a design, though.  I'll liaise with Dimos and @simon-johnson-r3 to lock down the requirements a bit further.

driessamyn: I wonder if the interval needs to be part of the message as well (even when it's configured) as I'm still wondering how you interpret these.
For example, say a v1 worker heartbeats, then gets killed off, how long does the consumer wait till it can assume that that worker doesn't exist anymore? If it has the poll interval along with the timestamp, then it could assume that, for example, if it hasn't had a heartbeat from somebody in 10 x interval, it probably doesn't exist anymore.

This is also why I'm wondering if a compacted topic would be better as a new worker can get a full view of the state immediately. In that case we'd still need to think about clean-up of the topic for workers that no longer exist, but maybe we can use the reconciler pattern for this.

dimosr: If that compacted topic is expected to only contain ephemeral data (e.g. say heartbeats are relevant for the last few minutes), we could also require it to be configured with a tighter retention period at the Kafka level. So, old/stale heartbeat records are cleaned up by the brokers.

fowlerrr: I like that idea, as it means less chance of us screwing it up.  Though we'd then lose the flexibility of configuration.

I'd still prefer it as the code to make those decisions is left with a time-tested library.  But I could easily go either way.

driessamyn: Y. It could be a combination of both. I.e. we rely on Kafka for cleanup with a quite large window (e.g. 1hour), but at a code level we can make a decision on whether we consider a worker not relevant anymore.

fowlerrr: Okay, let's plan on having both the timestamp _and_ using Kafka to prune for us.  I imagine the instance id would be the key.  And maybe the worker type, though it wouldn't really be necessary.   For now I'll keep both the instance id _and_ the worker type in the message.  I feel like, even though it's the key, it's simpler to have it in the message.

simon-johnson-r3: If you want to use a compacted topic + timestamp rather than a heartbeat system (and I think that's a reasonable idea as we already decided these weren't heartbeats in the sense everyone would interpret them) I suggest the _only_ valid key is instance id, and therefore it makes no sense to repeat that in the message.

dimosr: I think we want both worker type + instance as part of the key, because we currently provide the option to the user to specify their own worker ID (the default is a random value). And they might want to use sequential numbers.

simon-johnson-r3: Ok in that case I would change my mind! If we are to encode worker type and instance together in the key it makes sense to me to make that opaque to the reader of the record and explicitly add those into the message.

driessamyn: @dimosr comment made me thing maybe we shouldn't use `instanceId` in the key, but generate a unique worker ID when a worker starts up (e.g. `workertype-<uuid>`). `instanceId` as I understand, is actually messaging config, so may it's a bad idea to overload it?

fowlerrr: Maybe, though as long as it's unique I think we'd be okay.  I don't like the idea of using something that a user can specify, though, as it seems like we're giving them rope to hang themselves, so I'm okay with generating our own UUID.  The only issue there would be if that id were then useful somewhere else.  IE, how do we let them know the id if they want it for filtering or to action on a specific worker?  Or would that matter if they have the other one?

fowlerrr: To be clear, I still think we should put these into the message as well as the key

dimosr: Back in the days I was closer to Kafka, I remember there was a benefit from having stable IDs (IIRC, we use that ID as part of the producer IDs and it's involved in logic for Kafka transactions around cleanup of zombie transactions etc.). This is one case where allowing the user to provide an ID might be preferable. Needs someone to take a closer look at this. 

Also, for operational reasons, people might prefer to have easy-to-refer sequential numbers starting from 1 instead of random IDs. For example, before we had the helm charts, we used to use k8s ordinals for that ID ins our ad-hoc k8s scripts, for the reasons described above.

fowlerrr: That is a good point.  Does restarting a worker mean it's the _same_ worker?  Or does it become a different one?  With the concept of dynamic clusters it might not matter much.  Also, if we're trusting Kafka to prune old messages we probably don't (in this case) need to worry about zombie messages so much.

I'm looking through the messaging code to see how we currently use the instance id

simon-johnson-r3: I'm not really sure how a user could technically allocate an Id to an instance of a Worker when it's a pod in a scalable replica set which itself is driven by a deployment. You are two steps of abstraction away from creating worker pods at this point. There are also no stable ids or ordinals in such a configuration.

I think I would start from asking the question of why the end user needs to know this id, and the only answer I can come up with is to identify the pod, and the only useful thing I believe they can then do with that knowledge is to kill it or possibly extract a log from it if we've asked for one. So then I would conclude a unique pod name I can use with `kubectl` or another k8s tool is all I'm really after.

Perhaps there are scenarios I'm unaware of though?

simon-johnson-r3: > That is a good point. Does restarting a worker mean it's the same worker? Or does it become a different one?

Right now it's technically considered a different worker. Whether it's replacing an old one which crashed, it's part of a rolling update, or a consequence of a scale up is incidental. We do not have StatefulSets. Note that if we started to introduce stable Kafka transaction ids we'd suddenly by thrown into a world where fencing errors were something we'd need to be considering. That's not necessarily a problem, but we'd probably need a good reason to explain why we were considering giving workers a specific role like that.

dimosr: I agree that the operational convenience of numerical IDs does not probably warrant the complexity, a unique ID would not be that different there.

But, the interaction with Kafka and its guarantees is an important one. And I think we need to worry about fencing exceptions, mostly in the context of preserving the atomicity of our subscriptions (which rely on Kafka transactions under the hood). Thinking a bit more about it (and after reading the [producer docs](https://kafka.apache.org/10/javadoc/index.html?org/apache/kafka/clients/producer/KafkaProducer.html) around `transactional.id` again and refreshing some old memories) though, I think that providing a stable ID via a stateful set would not fully solve the issue either, since that ID needs to be stable for a specific set of partitions, which I don't think we do at the moment anyway.

simon-johnson-r3: The current pattern is that a worker instance is assigned one or more partitions and becomes entirely responsible for processing events on those partitions. The offsets for events it consumed are updated within the same transaction that produces the output events. If a rebalance occurs whilst a subscription is in the middle of processing an event it is aborted and the offsets for consuming are never set, thus the event is "returned" for whatever worker next picks up that partition. This is how we stop duplicate processing of events using transactions.

If we tried to keep partitions tied to a certain worker's role (unique transaction id) then there would perhaps be merit in having some id which told that worker what part it played, but currently we don't. I think that's what you describe here. Perhaps an advantage there would be to reduce the number of rebalances which are costly, but the price would be partitions potentially not being serviced for longer during a pod crash (this is speculative) and having to deal with the possibility workers are replaced by new copies of themselves (we do handle fencing exceptions as it happens, but it's a unexpected scenario right now so wouldn't be fully tested).

Is there any other gap in the pattern we have now which you see transaction ids would solve? In my mind they are there for when you have more stateful services which perform a specific role based on knowledge they are playing a particular role in a big system of component parts. We don't have that because all workers are considered equal.

dimosr: I think they are supposed to protect against scenarios where you have a zombie producer running that leads to records being written twice (violating the "exactly-once" guarantees), by ensuring that the instantiation of the new producer would fence any old zombie producers with the same ID.

fowlerrr: I spoke with Lorcan about this and it was something that we'd thought would be useful before k8s came along and meant that each instance id will _always_ be unique (because it has to be with k8s).

So there's no real way to use it for fencing producers as it'll always be unique.

fowlerrr: So what we've come up with (for instance ids) is to have a UUID for the key on the topic, and to use the HOSTNAME from k8s (as an env var) for the field in the message.  That name will be something readable and meaningful for an operator who might look at this message.  But might possibly be reused, therefore UUID for the key.

driessamyn: Maybe a bit off-topic, but is there ever a usecase for setting instanceID? if not, should we not remove it as a start-up param?

PR #826, opened 2023-01-30
CORE-9775 Automatic Token Locking

owenstanford:
This PR changes:
 - The return type of the UTXO finalize API to potentially return a token claim as part of the response.
 - Allows the flow to specify the token claim behavior of a finalized transaction.
The purpose of this change is to allow automatic claiming of tokens output from a finalized transaction. This can be used in cases where the finalizing flow intends to immediately spend the output states from a transaction it has finalized. In this case there is a potential race condition with another flow claiming and spending the states before the original flow can. automatically locking or claiming the states during finalization prevents this.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -7,7 +7,7 @@ cordaProductVersion = 5.0.0
 # NOTE: update this each time this module contains a breaking change
 ## NOTE: currently this is a top level revision, so all API versions will line up, but this could be moved to
 ##   a per module property in which case module versions can change independently.
-cordaApiRevision = 732
+cordaApiRevision = 730
 
 # Main
 kotlinVersion = 1.8.10
--- a/gradle.properties
+++ b/gradle.properties
@@ -7,7 +7,7 @@ cordaProductVersion = 5.0.0
 # NOTE: update this each time this module contains a breaking change
 ## NOTE: currently this is a top level revision, so all API versions will line up, but this could be moved to
 ##   a per module property in which case module versions can change independently.
-cordaApiRevision = 729
+cordaApiRevision = 732
 
 # Main
 kotlinVersion = 1.8.10
--- a/ledger/ledger-utxo/src/main/java/net/corda/v5/ledger/utxo/FinalizationResult.java
+++ b/ledger/ledger-utxo/src/main/java/net/corda/v5/ledger/utxo/FinalizationResult.java
@@ -1,0 +1,20 @@
+package net.corda.v5.ledger.utxo;
+
+import net.corda.v5.base.annotations.DoNotImplement;
+import net.corda.v5.ledger.utxo.transaction.UtxoSignedTransaction;
+import org.jetbrains.annotations.NotNull;
+
+/**
+ * Defines the result of transaction finalization
+ *
+ */
+@DoNotImplement
+public interface FinalizationResult {
+    /**
+     * Gets the finalized transaction.
+     *
+     * @return An instance of {@link UtxoSignedTransaction}.
+     */
+    @NotNull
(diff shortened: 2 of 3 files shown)


ledger/ledger-utxo/src/main/kotlin/net/corda/v5/ledger/utxo/TokenClaimBehaviorType.kt
```diff
@@ -0,0 +1,26 @@
...
+import net.corda.v5.base.annotations.CordaSerializable
+
+/**
+ * Defines the token claim behavior when finalizing a transaction.
+ */
+@CordaSerializable
+enum class TokenClaimBehaviorType {
+    /**
+     * Automatically claim any tokens output during finalization. Any claimed tokens are returned via the
+     * [FinalizationResult.tokenClaim] property. This is the default option.
```
blsemo: Is that the correct default option? I guess this only  makes sense if your states are actually tokens that can be claimed/selected?
Also, if that isn't implemented, having this as default means that the default behaviour will change when this gets implemented.

ledger/ledger-utxo/src/main/kotlin/net/corda/v5/ledger/utxo/transaction/UtxoTransactionBuilder.kt
```diff
@@ -180,6 +181,16 @@ interface UtxoTransactionBuilder {
...
 
+    /**
+     * Sets the type of token claim behavior when finalizing a transaction
+     *
+     * @param behaviorType The [TokenClaimBehaviorType] for the transaction. If unspecified defaults to
+     * [TokenClaimBehaviorType.CLAIM]
+
+     * @return Returns a [UtxoTransactionBuilder] including the specified claim behavior.
+     */
+    fun setTokenClaimBehavior(behaviorType: TokenClaimBehaviorType): UtxoTransactionBuilder
```
vlajos: Out of curiosity, why is not this a finality flow parameter?

ledger/ledger-utxo/src/main/kotlin/net/corda/v5/ledger/utxo/TokenClaimBehaviorType.kt
lankydan: Agree with this comment that the default should be to not claim the tokens, make it opt in.

lankydan: I think `ACKNOWLEDGE` seems reasonable as a default but will cause behaviour change if this isn't implemented on the backend yet.

ledger/ledger-utxo/src/main/kotlin/net/corda/v5/ledger/utxo/transaction/UtxoTransactionBuilder.kt
lankydan: I agree that it should be part of `finalize`.

Do we also want something similar for `receiveFinality`, is there a benefit in having them receive the tokens as well, as they will also have them stored and available to spend. This would mean receive finality would have both a `TokenClaimBehaviourType` argument and also a return type of `FinalizationResult`.

owenstanford (commented):
This change has been reduced to just changing the return type of finalization to a new interface that can be extended in the future.
Any future automated token locking will require an option passed into the finalize methods, if/when this is done it will be added as an overload of the existing method and therefore will not be a breaking change and can be left for a future version.

lankydan (approved):

blsemo (approved):
LGTM

PR #828, opened 2023-01-31
CORE-9062 Add rule type to the group approval rules delete message

charlieR3:
Add rule type to the delete avro schema for approval rules so that we can segment rules in the database between standard and preauth. This is mainly so that we can have separate delete endpoints for standard and preauth rules and avoid a scenario where we can delete preauth rules through that standard rules endpoint, and vice versa.


nikinagy (dismissed):
LGTM

yift-r3 (dismissed):

williamvigorr3 (commented):
Do you need to update `data/db-schema/src/main/resources/net/corda/db/schema/vnode-vault/migration/vnode-vault-creation-v1.0.xml` so vnode_group_approval_rules has a composite primary key (to match the runtime-os PR)?

charlieR3:
> Do you need to update `data/db-schema/src/main/resources/net/corda/db/schema/vnode-vault/migration/vnode-vault-creation-v1.0.xml` so vnode_group_approval_rules has a composite primary key (to match the runtime-os PR)?

That's a very good point. It seems to work fine without it but I think it should be done. I'll take a look now.

dimosr (approved):

PR #829, opened 2023-01-31
CORE-8735: Move registration request from RPC to async

yift-r3:
Runtime Pull Request - https://github.com/corda/corda-runtime-os/pull/3023

Creates a new durable queue for registration requests to be sent from the RPC worker to the membership worker asynchronously.


data/avro-schema/src/main/resources/avro/net/corda/data/membership/common/RegistrationStatus.avsc
```diff
@@ -4,7 +4,8 @@
   "namespace": "net.corda.data.membership.common",
   "doc": "Registration request status.",
   "symbols": [
-    "NEW",
+    "SUBMITTED",
+    "PENDING_MGM_NETWORK_ACCESS",
```
dimosr: Is it worth to keep `NEW`, instead of `SUBMITTED`, to avoid misleading people into thinking it's been submitted to the MGM?

charlieR3: Or alternatively `ACCEPTED` instead of `NEW`

charlieR3: Just for my own understanding, is it correct that `PENDING_MGM_NETWORK_ACCESS` means the registration has been sent to the MGM and we are now waiting for the MGM to grant access to the network?

yift-r3: I think so; the member had received the request and sent it to the MGM to approve - I wonder if it's the best name...

charlieR3: Perhaps `SUBMITTED` could be a good status for when it was sent to the MGM. Or else a `SENT_TO_MGM` status. 

I think as it stands, the members view of the status would be `PENDING_MGM_NETWORK_ACCESS` -> `PENDING_MEMBER_VERIFICATION` -> `PENDING_MANUAL_APPROVAL`/`PENDING_AUTO_APPROVAL` -> `APPROVED`/`DECLINED`. So there's a few more statuses after `PENDING_MGM_NETWORK_ACCESS` before actually getting access so it might be better to name it something more specific to indicate the request has been sent and we're waiting for the MGM to respond.

dimosr: the issue I see with `ACCEPTED` is people might be thinking "accepted by the MGM" (not the cluster). As always, naming is the biggest problem..

data/avro-schema/src/main/resources/avro/net/corda/data/membership/common/RegistrationStatus.avsc
```diff
@@ -5,11 +5,14 @@
   "doc": "Registration request status.",
   "symbols": [
     "NEW",
+    "SENT_TO_MGM",
+    "RECEIVER_BY_MGM",
```
dimosr:
```suggestion
    "RECEIVED_BY_MGM",
```
?

dimosr (dismissed):
LGTM

dimosr (approved):

PR #830, opened 2023-01-31
CORE-9787: Enable local build cache

knguyen-r3:
* Enable usage of build cache locally
* Implemented new identity access base using gradleEnterprise.buildCache
Local Test Result: https://gradle.dev.r3.com/s/ck3nkkoewko56
Remote Test Result: https://gradle.dev.r3.com/scans?search.tags=PR-830&search.timeZoneId=Europe/London

### DIFF
--- a/settings.gradle
+++ b/settings.gradle
@@ -107,15 +107,15 @@ include 'serialization'
 gradleEnterprise {
     server = gradleEnterpriseUrl
     allowUntrustedServer = false
+    def apiKey = settings.ext.find('CORDA_GRADLE_SCAN_KEY') ?: System.getenv('CORDA_GRADLE_SCAN_KEY')
+    accessKey = apiKey
     buildScan {
-        def apiKey = settings.ext.find('CORDA_GRADLE_SCAN_KEY') ?: System.getenv('CORDA_GRADLE_SCAN_KEY')
         if (apiKey?.trim()) {
             publishAlways()
             capture {
                 taskInputFiles = true
             }
             uploadInBackground = false
-            accessKey = apiKey
         }
     }
     buildCache {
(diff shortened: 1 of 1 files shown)


settings.gradle
```diff
@@ -107,35 +107,30 @@ include 'serialization'
 gradleEnterprise {
     server = gradleEnterpriseUrl
     allowUntrustedServer = false
+    def apiKey = settings.ext.find('CORDA_GRADLE_SCAN_KEY') ?: System.getenv('CORDA_GRADLE_SCAN_KEY')
+    accessKey = apiKey
```
yarivus, line 111: is accessKey used at all ?

yarivus (commented):
see my question/comment

settings.gradle
ronanbrowne: @yarivus yes it's used internally by the `gradleEnterprise`  plugin, which provides this closure `accessKey` is a property which is provided by the plugin a user may set for authentication.

this is the same as other vars you see here such as `server ` / `allowUntrustedServer `

ronanbrowne (approved):
LGTM

settings.gradle
```diff
@@ -107,35 +107,30 @@ include 'serialization'
...
-            }
+        remote(gradleEnterprise.buildCache) {
             // For the remote build cache we will populate cache only from Jenkins, all machines can pull from cache however.
             if (System.getenv().containsKey("JENKINS_URL")) {
                 push = true
                 enabled = true
             } else {
                 push = false
-                enabled = false
+                enabled = true
```
ronanbrowne:
```suggestion
                enabled = apiKey?.trim() ? true : false
```

We need to ensure not active if key not set

ronanbrowne (changes requested):
see comment

settings.gradle
knguyen-r3: It is used within Gradle. You can search for accessKey in Gradle documentation [here](https://docs.gradle.com/enterprise/gradle-plugin/#using_the_gradle_enterprise_connector).

yarivus (approved):
LGTM

ronanbrowne (approved):
LGTM

PR #831, opened 2023-02-01
CORE-7647 Add async request support for creating virtual nodes.

owenstanford:
This change allows virtual node creation requests to be sent async.

### DIFF
--- a/data/avro-schema/src/main/resources/avro/net/corda/data/virtualnode/VirtualNodeAsynchronousRequest.avsc
+++ b/data/avro-schema/src/main/resources/avro/net/corda/data/virtualnode/VirtualNodeAsynchronousRequest.avsc
@@ -21,7 +21,8 @@
       "name": "request",
       "doc": "The operation specific request.",
       "type": [
-        "net.corda.data.virtualnode.VirtualNodeUpgradeRequest"
+        "net.corda.data.virtualnode.VirtualNodeUpgradeRequest",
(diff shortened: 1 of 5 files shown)


conal-smith (dismissed):
LGTM

data/avro-schema/src/main/resources/avro/net/corda/data/virtualnode/VirtualNodeAsynchronousRequest.avsc
```diff
@@ -21,7 +21,8 @@
       "name": "request",
       "doc": "The operation specific request.",
       "type": [
-        "net.corda.data.virtualnode.VirtualNodeUpgradeRequest"
+        "net.corda.data.virtualnode.VirtualNodeUpgradeRequest",
+        "net.corda.data.virtualnode.VirtualNodeCreateRequest"
```
vkolomeyko, line 25: Any particular reason why we cannot keep `VirtualNodeCreateRequest` inside `VirtualNodeManagementRequest` envelope?
Is it because there is a `Future` attached to any `VirtualNodeManagementRequest` processing code?

Since we are making changes to `corda-api` would it also not make sense to remove `VirtualNodeCreateRequest` from `VirtualNodeManagementRequest` then?

Also, as discussed, `groupId` will need to be passed as part of `VirtualNodeCreateRequest` to be populated from Rest worker, since we are changing `corda-api` shall we add this breaking change as well?

vkolomeyko (approved):
LGTM

PR #836, opened 2023-02-02
CORE-9029: add schemas for Virtual Node Operation Status support. 

Tom-Fitzpatrick:
This adds schemas for `VirtualNodeOperationStatusResponse` , `VirtualNodeOperationStatusRequest `and makes some changes to the `VirtualNodeOperationStatus `schema to support getting Virtual Node Operation Statuses through the new endpoint. 

This will support the changes being added in https://github.com/corda/corda-runtime-os/pull/3089


data/db-schema/src/main/resources/net/corda/db/schema/config/migration/config-creation-v1.0.xml
```diff
@@ -263,6 +263,9 @@
         </createTable>
         <addPrimaryKey columnNames="id" constraintName="virtual_node_operation_pk" tableName="virtual_node_operation"
                        schemaName="${schema.name}"/>
+        <createIndex indexName="virtual_node_operation_request_id" tableName="virtual_node_operation" schemaName="${schema.name}">
+            <column name="request_id"/>
+        </createIndex>
```
sprintengine, line 255: We'll be querying the virtual node operation table using the `request_id`, not the primary key `id`, FYI @vkolomeyko on this index I suggested adding.

data/avro-schema/src/main/resources/avro/net/corda/data/virtualnode/VirtualNodeOperationStatusResponse.avsc
```diff
@@ -0,0 +1,18 @@
...
+  "type": "record",
+  "name": "VirtualNodeOperationStatusResponse",
+  "namespace": "net.corda.data.virtualnode",
+  "fields": [
+    {
+      "name": "requestId",
+      "type": "string"
+    },
+    {
+      "name": "status",
```
sprintengine: since this is list, might be best to name this something slightly different, just thinking out aloud:
- statuses
- statusList
- operationHistory
- requestHistory
- requestStatuses
- statusHistory
- requestHistory

Possibly requestHistory is jumping out at me the most, what do you think?

conal-smith (dismissed):
Looks good to me, though think about renaming the status

data/avro-schema/src/main/resources/avro/net/corda/data/virtualnode/VirtualNodeOperationStatusResponse.avsc
Tom-Fitzpatrick: Thanks, definitely should be something other than "status". Thinking through them, I like "operationHistory" as there will be one entry for each operation that was made, rather than for each request or individual status change. I'll go with this for now, but happy to change this if we prefer one of the others or something else altogether.

data/avro-schema/src/main/resources/avro/net/corda/data/virtualnode/VirtualNodeOperationStatus.avsc
```diff
@@ -43,28 +31,23 @@
...
-        "type": "long",
-        "logicalType": "timestamp-millis"
-      },
+      "type": [
+        "null",
+        {
+          "type": "long",
+          "logicalType": "timestamp-millis"
+        }
+      ],
```
Tom-Fitzpatrick, line 40: At the moment, upgrading a CPI is generating an Operation Status with a null heartbeatTimestamp. I've updated the schema to allow this, but maybe that's something we need to change in the implementation of the Update instead, so just wanted to highlight this.

sprintengine: Good point, yes we can just make this nullable as it's nullable in the database too

conal-smith (dismissed):
lgtm

conal-smith (approved):
https://github.com/corda/corda-runtime-os/pull/3279 proves this is non-breaking

vkolomeyko (approved):
LGTM

PR #837, opened 2023-02-02
CORE-9716 Remove explicit use of schemas from Liquibase files

simon-johnson-r3:
Removed explicit schema definitions from Liquibase XML files. Instead it is up to the user of the generated SQL files to apply them to a schema of their choice, which may even include having them in the public schema of an isolated database for each functional area.

Corda's `database` plugin will automatically add schemas for customers if required and Corda's default bootstrapping in the helm chart will run the plugin in this mode. That PR is here: https://github.com/corda/corda-runtime-os/pull/3057

Without this change it is not possible to migrate database to later versions of the schemas, because the change log tables generated by Liquibase must be in the same schema as the tables and other resources it adds. Before this PR there was a mismatch of where they ended up per file.


gradle.properties
```diff
@@ -7,6 +7,7 @@ cordaProductVersion = 5.0.0
 # NOTE: update this each time this module contains a breaking change
 ## NOTE: currently this is a top level revision, so all API versions will line up, but this could be moved to
 ##   a per module property in which case module versions can change independently.
+cordaApiRevision = 631
 cordaApiRevision = 630
```
fowlerrr: Did this render wrong or am I seeing an extra line instead of a replaced one?

simon-johnson-r3: you're correct, merge error, will update

benmcmahon100 (dismissed):
All looks good to me

driessamyn (dismissed):
LGTM

driessamyn (dismissed):
LGTM

fowlerrr (approved):

JamesHR3 (approved):

PR #838, opened 2023-02-03
CORE-9701 - Handle multiple versions of member info

nikinagy:
Ability to handle multiple versions of MemberInfo base don statuses. Added status based lookups and new keying strategies.
P2P layer got modified to use the status filters.
To achieve these avro schemas had to be changed to contain the status filter.


data/db-schema/src/main/resources/net/corda/db/schema/vnode-vault/migration/vnode-vault-creation-v1.0.xml
```diff
@@ -88,6 +88,9 @@
             <column name="member_name" type="NVARCHAR(255)">
                 <constraints nullable="false"/>
             </column>
+            <column name="is_pending" type="BOOLEAN">
+                <constraints nullable="false"/>
+            </column>
             <column name="status" type="NVARCHAR(64)">
```
charlieR3, line 97: Do we need both `status` and `is_pending`? Or can we just add `status` to the primary key?

data/avro-schema/src/main/resources/avro/net/corda/data/membership/p2p/MembershipRegistrationRequest.avsc
```diff
@@ -21,6 +21,11 @@
       "name": "memberSignature",
       "doc": "Signature of the member over the memberContext.",
       "type": "net.corda.data.crypto.wire.CryptoSignatureWithKey"
+    },
+    {
+      "name": "isPending",
```
charlieR3, line 26: Why do we need this on a registration request? Aren't all registration requests pending when sent from member to MGM?

data/avro-schema/src/main/resources/avro/net/corda/data/p2p/app/AuthenticatedMessageHeader.avsc
```diff
@@ -38,6 +38,16 @@
       "doc": "This value identifies the upstream user of the p2p layer that this message is sent from and should be received by. It can be used to filter incoming messages from the p2p layer and process only the ones destined for a specific system.",
       "name": "subsystem",
       "type": "string"
+    },
+    {
+      "doc": "Indicates what type of membership information should be used for P2P communication: pending, active, active or suspended. Should be only set by MGM services. Default value should be used everywhere else, which means that active member information will be used.",
```
charlieR3: _"Should be only set by MGM services."_ Is that correct? Isn't it used by any service creating an authenticated message header? If I understand right, would a better description here mention that non-MGM services should only use `ACTIVE`, and only MGM services will use other options?

data/db-schema/src/main/resources/net/corda/db/schema/vnode-vault/migration/vnode-vault-creation-v1.0.xml
nikinagy: I think is_pending was added to have less entries, because when we will have suspended as well then it won't overwrite the active member info, because of the keying strategy, so effectively we will have unnecessary data in the db.

data/avro-schema/src/main/resources/avro/net/corda/data/membership/p2p/MembershipRegistrationRequest.avsc
nikinagy: Unfortunately, we are persisting the signatures from the request and in a different table, but that table is using the same keys as the member info one and queries for them during distribution. As a temporary solution until we do https://r3-cev.atlassian.net/browse/CORE-10689 i put the signatures as pending and then update the is_pending field once the members are approved, so that their signatures can be used.

data/db-schema/src/main/resources/net/corda/db/schema/vnode-vault/migration/vnode-vault-creation-v1.0.xml
charlieR3: When we have suspended shouldn't we remove the active version? So we can only have a pending member info and either and active or a suspended but not all three? Or at least a soft delete (i.e. the is_deleted column) so that the reconciler (when added) will remove the stale entries from the DB even when keying with status.

data/avro-schema/src/main/resources/avro/net/corda/data/membership/p2p/MembershipRegistrationRequest.avsc
charlieR3: Ok, sounds fine as long as it will be updated in CORE-10689

data/db-schema/src/main/resources/net/corda/db/schema/vnode-vault/migration/vnode-vault-creation-v1.0.xml
nikinagy: I think here: https://r3-cev.atlassian.net/wiki/spaces/CB/pages/4345921553/MGM+keying+strategy+for+membership+list
We decided on approach number 3 without any deletion/clean-up.
We can review the approach if we want.

data/avro-schema/src/main/resources/avro/net/corda/data/membership/p2p/MembershipRegistrationRequest.avsc
charlieR3: Hmm, ok. Doesn't seem ideal but if it will be fixed in CORE-10689 it's not too bad.

data/avro-schema/src/main/resources/avro/net/corda/data/p2p/app/AuthenticatedMessageHeader.avsc
nikinagy: Addressed in [9e55b51](https://github.com/corda/corda-api/pull/838/commits/9e55b51eb09a822c5ab8e20083c34edceb8bfa97)

charlieR3 (dismissed):

williamvigorr3 (approved):
LGTM (I checked the commits since Charlie's approval).

PR #840, opened 2023-02-06
CORE-9881 : Changes for interop message schema

SaurabhR3:
1. Added Interop Message schema
2. Changed the version 634

### DIFF
--- a/data/avro-schema/src/main/resources/avro/net/corda/data/interop/InteropMessage.avsc
+++ b/data/avro-schema/src/main/resources/avro/net/corda/data/interop/InteropMessage.avsc
@@ -3,19 +3,14 @@
   "name": "InteropMessage",
   "namespace": "net.corda.data.interop",
   "fields": [
-      {
-        "doc": "Unique message identifier.",
-        "name": "messageId",
-        "type": "string"
-     },
     {
-      "doc": "Request contains json data of facade request.",
-      "name": "request",
+      "doc": "Unique message identifier.",
+      "name": "messageId",
       "type": "string"
     },
     {
-      "doc": "Response contains json data of facade response.",
-      "name": "response",
+      "doc": "Payload contains json data of facade request.",
+      "name": "payload",
       "type": "string"
     }
   ]
--- a/data/avro-schema/src/main/resources/avro/net/corda/data/interop/InteropMessage.avsc
+++ b/data/avro-schema/src/main/resources/avro/net/corda/data/interop/InteropMessage.avsc
@@ -1,0 +1,22 @@
+{
+  "type": "record",
+  "name": "InteropMessage",
+  "namespace": "net.corda.data.interop",
+  "fields": [
+      {
+        "doc": "Unique message identifier.",
+        "name": "messageId",
+        "type": "string"
+     },
+    {
+      "doc": "Request contains json data of facade request.",
+      "name": "request",
+      "type": "string"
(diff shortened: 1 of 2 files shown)


data/avro-schema/src/main/resources/avro/net/corda/data/interop/InteropMessage.avsc
```diff
@@ -0,0 +1,22 @@
...
+    {
+      "doc": "Request contains json data of facade request.",
+      "name": "request",
+      "type": "string"
+    },
+    {
+      "doc": "Response contains json data of facade response.",
+      "name": "response",
+      "type": "string"
+    }
```
yarivus: Please fix to consistent indentation

gradle.properties
```diff
@@ -7,7 +7,7 @@ cordaProductVersion = 5.0.0
 # NOTE: update this each time this module contains a breaking change
 ## NOTE: currently this is a top level revision, so all API versions will line up, but this could be moved to
 ##   a per module property in which case module versions can change independently.
-cordaApiRevision = 633
+cordaApiRevision = 634
```
szymonsztuka: Since our branch will be in parallel to main, over the course they will increase this number independently anyway, so I think we could leave it now as it is. Once we do PR against main we can update this number if needed.

data/avro-schema/src/main/resources/avro/net/corda/data/interop/InteropMessage.avsc
```diff
@@ -0,0 +1,22 @@
...
+        "name": "messageId",
+        "type": "string"
+     },
+    {
+      "doc": "Request contains json data of facade request.",
+      "name": "request",
+      "type": "string"
+    },
+    {
+      "doc": "Response contains json data of facade response.",
```
szymonsztuka: This schema suggests that you have both request and response in a single message, which is wrong. We could make both fields nullable, but we couldn't impose only one to be present.
What about changing this to:  
```
{
       "fields": [
               "type": "enum",  "name": "request", value =Request/response
              name: "payload",  type: String,
              doc: ""
              ]
 }
 ```
 or anything to make only one filed obligatory.

szymonsztuka (changes requested):

data/avro-schema/src/main/resources/avro/net/corda/data/interop/InteropMessage.avsc
SaurabhR3: I agree with your first point but I think its not good representation of schema adding request enum with value request or response. I personally feel its better to keep it simple as 

> {
      "doc": "Payload contains json data of facade request.",
      "name": "payload",
      "type": "string"
    }

here payload can be request or response based on consumer. Another way I am thinking, we can have separate schema for response based on how it evolve. We may need some additional fields while returning the response.

Whats your opinion?

yarivus (approved):
LGTM

szymonsztuka (approved):
We may need some flag if this is request or response, not sure if from the payload perspective we can deduce it, for any anyway request with created in different context than response, so we can distinguish them anyway.

jmacmahonr3:
Is anywhere else in Corda 5 doing something similar ? Are we going with the grain more by having two message types or reusing one ?

jmacmahonr3 (approved):
LGTM

PR #843, opened 2023-02-07
CORE-10119-new-secret-keys

dickon:
Some secrets addons may need additional configuration keys; define enough for now.


gradle.properties
```diff
@@ -7,7 +7,7 @@ cordaProductVersion = 5.0.0
 # NOTE: update this each time this module contains a breaking change
 ## NOTE: currently this is a top level revision, so all API versions will line up, but this could be moved to
 ##   a per module property in which case module versions can change independently.
-cordaApiRevision = 635
+cordaApiRevision = 636
```
driessamyn: We don't need to bump the version number as this is not a breaking change.

driessamyn (commented):
Looks ok apart from the version bump

PR #844, opened 2023-02-07
CORE-10119: add new keys for server based secrets addons

dickon:
CORE-10119: add new keys for server based secrets addons


data/config-schema/src/main/kotlin/net/corda/schema/configuration/ConfigKeys.kt
```diff
@@ -32,10 +32,14 @@ object ConfigKeys {
     const val REST_WEBSOCKET_CONNECTION_IDLE_TIMEOUT_MS = "websocket.idleTimeoutMs"
 
     // Secrets Service
+    // See configuration of each secrets service addon for the subset of 
+    // these keys needed for that addon.
```
driessamyn: I think we can be specific here rather than refer to each add-on.

dickon: fixed

driessamyn (approved):
LGTM

PR #846, opened 2023-02-08
CORE-9550 Add full public key id for signing keys in crypto worker

kyriathar:
The reason for this change is potential clashes in short key ids during keys look up, which could lead in returning a key which is not owned by the caller.

- adds full key id in `crypto_signing_key` DB table.
- introduces avro types for short and full key ids (`ShortHashes` for short key ids and `SecureHashes` for full key ids)

runtime-os PR: [#3091](https://github.com/corda/corda-runtime-os/pull/3091)

### DIFF
--- a/data/avro-schema/src/main/resources/avro/net/corda/data/crypto/ShortHashes.avsc
+++ b/data/avro-schema/src/main/resources/avro/net/corda/data/crypto/ShortHashes.avsc
@@ -1,0 +1,15 @@
+{
+  "type": "record",
+  "name": "ShortHashes",
+  "doc": "List of short hashes. A short hash is the first 12 characters of a hex string",
+  "namespace": "net.corda.data.crypto",
+  "fields": [
+    {
+      "name": "hashes",
+      "type": {
+        "type": "array",
+        "items": "string"
+        }
+    }
+  ]
+}
--- a/data/avro-schema/src/main/resources/avro/net/corda/data/crypto/ShortHashes.avsc
+++ b/data/avro-schema/src/main/resources/avro/net/corda/data/crypto/ShortHashes.avsc
@@ -1,7 +1,7 @@
 {
   "type": "record",
   "name": "ShortHashes",
-  "doc": "List of short hashes. A short hash is the first 12 characters of the hex string form of the hash",
+  "doc": "List of short hashes. A short hash is the first 12 characters of the hex string form of the hash. Short hashes should be used with care as it could lead to collisions",
   "namespace": "net.corda.data.crypto",
   "fields": [
     {
--- a/data/avro-schema/src/main/resources/avro/net/corda/data/crypto/ShortHashes.avsc
+++ b/data/avro-schema/src/main/resources/avro/net/corda/data/crypto/ShortHashes.avsc
@@ -1,7 +1,7 @@
 {
   "type": "record",
   "name": "ShortHashes",
-  "doc": "List of short hashes. A short hash is the first 12 characters of the hex string form of the hash. Short hashes should be used with care as it could lead to collisions",
+  "doc": "List of short hashes. A short hash is the first 12 characters of the hex string form of the hash. Short hashes should be used with care as it could lead to collisions.",
   "namespace": "net.corda.data.crypto",
   "fields": [
     {
--- a/data/avro-schema/src/main/resources/avro/net/corda/data/crypto/ShortHashes.avsc
+++ b/data/avro-schema/src/main/resources/avro/net/corda/data/crypto/ShortHashes.avsc
@@ -1,7 +1,7 @@
 {
   "type": "record",
   "name": "ShortHashes",
-  "doc": "List of short hashes. A short hash is the first 12 characters of a hex string",
+  "doc": "List of short hashes. A short hash is the first 12 characters of the hex string form of the hash",
   "namespace": "net.corda.data.crypto",
   "fields": [
     {
--- a/data/avro-schema/src/main/resources/avro/net/corda/data/crypto/wire/ops/flow/queries/ByIdsFlowQuery.avsc
+++ b/data/avro-schema/src/main/resources/avro/net/corda/data/crypto/wire/ops/flow/queries/ByIdsFlowQuery.avsc
@@ -2,14 +2,14 @@
   "type": "record",
   "name": "ByIdsFlowQuery",
   "namespace": "net.corda.data.crypto.wire.ops.flow.queries",
-  "doc": "Request to lookup for keys by their ids",
+  "doc": "Request to lookup for keys by their ids. Ids can be either short or full key ids. The maximum number of items is 20",
   "fields": [
     {
       "name": "keyIds",
-      "type": {
-        "type": "array",
-        "items": "string"
-      },
+      "type": [
+        "net.corda.data.crypto.ShortHashes",
+        "net.corda.data.crypto.SecureHashes"
+        ],
       "doc": "List of key ids"
     }
   ]
--- a/data/avro-schema/src/main/resources/avro/net/corda/data/crypto/wire/ops/flow/queries/ByIdsFlowQuery.avsc
+++ b/data/avro-schema/src/main/resources/avro/net/corda/data/crypto/wire/ops/flow/queries/ByIdsFlowQuery.avsc
@@ -2,14 +2,11 @@
   "type": "record",
   "name": "ByIdsFlowQuery",
   "namespace": "net.corda.data.crypto.wire.ops.flow.queries",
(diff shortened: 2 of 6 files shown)


data/avro-schema/src/main/resources/avro/net/corda/data/crypto/wire/ops/rpc/queries/ByIdsRpcQuery.avsc
```diff
@@ -2,15 +2,15 @@
   "type": "record",
   "name": "ByIdsRpcQuery",
   "namespace": "net.corda.data.crypto.wire.ops.rpc.queries",
-  "doc": "Request to lookup for keys by their ids, the maximum number of items is 20.",
+  "doc": "Request to lookup for keys by their ids. Ids can be either short or full key ids. The maximum number of items is 20",
```
mnesbit, line 5: why the limit of 20? Does the backend code break up large requests into a series of size 20 batch requests?

data/avro-schema/src/main/resources/avro/net/corda/data/crypto/ShortHashes.avsc
```diff
@@ -0,0 +1,15 @@
+{
+  "type": "record",
+  "name": "ShortHashes",
```
dickon, line 3: Is this temporary code or do we want to keep both kinds of hash long term, and if so, why do we need 2?

data/avro-schema/src/main/resources/avro/net/corda/data/crypto/wire/ops/rpc/queries/ByIdsRpcQuery.avsc
kyriathar: `SigningKeyStoreImpl` look up methods throw if requested keys found to be in number more than 20 (this was already there).

data/avro-schema/src/main/resources/avro/net/corda/data/crypto/ShortHashes.avsc
dickon: We will keep both hash formats, so that short forms can be used in HTTP address query strings.

data/avro-schema/src/main/resources/avro/net/corda/data/crypto/wire/ops/rpc/queries/ByIdsRpcQuery.avsc
```diff
@@ -2,15 +2,15 @@
...
-      "type": {
-        "type": "array",
-        "items": "string"
-      },
-      "doc": "List of key ids."
+      "name": "keyIds",
+      "type": [
+        "net.corda.data.crypto.ShortHashes",
+        "net.corda.data.crypto.SecureHashes"
+        ],
```
dickon: whitespace looks wrong - the closing ] should be algiedn with the open double quote of "type"

data/avro-schema/src/main/resources/avro/net/corda/data/crypto/ShortHashes.avsc
```diff
@@ -0,0 +1,15 @@
...
+  "doc": "List of short hashes. A short hash is the first 12 characters of a hex string",
```
dickon: better to say something like "List of short hashes. A short hash is the first 12 characters of the hex form of the hash. Short hashes should only be used when conciseness is important."

data/avro-schema/src/main/resources/avro/net/corda/data/crypto/wire/ops/rpc/queries/ByIdsRpcQuery.avsc
dickon: This number 20 (and the version inByIdsFlowQuery) feels like a server limitation rather than something which should be in the interface.

But it's already there, we should highlight in both places as you are doing if we we are to mention it one place, and I wouldn't hold up approval on this.

dickon (changes requested):
LGTM, other than some very minor nitpicks. I should be able to quickly approve once you've done the rename you want to do and addressed the nitpicks.

data/avro-schema/src/main/resources/avro/net/corda/data/crypto/ShortHashes.avsc
dickon: actually, now I've looked at the implementation and commented there, better to simply allow unique prefixes, in which case either:

a) no AVRO change at all - anywhere you can specify a hash you can use a unique prefix
b) we have a more general AVRO change which allows a unique hash prefix to be sent, in a way that can be differentiated from a full hash.

dickon (changes requested):
Now I understand the database level, my thinking is that an arbitrary fixed prefix length is more trouble than arbitary unique prefix length, so changing review status.

data/avro-schema/src/main/resources/avro/net/corda/data/crypto/wire/ops/rpc/queries/ByIdsRpcQuery.avsc
kyriathar: Addressed

data/avro-schema/src/main/resources/avro/net/corda/data/crypto/ShortHashes.avsc
kyriathar: Doc updated with suggestion

mnesbit (dismissed):
OK

data/avro-schema/src/main/resources/avro/net/corda/data/crypto/wire/ops/flow/queries/ByIdsFlowQuery.avsc
```diff
@@ -2,14 +2,14 @@
...
+  "doc": "Request to lookup for keys by their ids. Ids can be either short or full key ids. The maximum number of items is 20",
   "fields": [
     {
       "name": "keyIds",
-      "type": {
-        "type": "array",
-        "items": "string"
-      },
+      "type": [
+        "net.corda.data.crypto.ShortHashes",
```
dickon: We should not allow 48 bit ShortHashes for this query; it'd make it pointlessly easier for flows to pick up keypairs they shouldn't.

dickon (changes requested):

data/avro-schema/src/main/resources/avro/net/corda/data/crypto/wire/ops/flow/queries/ByIdsFlowQuery.avsc
kyriathar: Removed short key ids from `ByIdsFlowQuery`.

data/avro-schema/src/main/resources/avro/net/corda/data/crypto/ShortHashes.avsc
```diff
@@ -0,0 +1,15 @@
+{
+  "type": "record",
+  "name": "ShortHashes",
+  "doc": "List of short hashes. A short hash is the first 12 characters of the hex string form of the hash",
```
dickon: Please make this something like:

  List of short hashes. A short hash is the first 12 characters of the hex string form of the secure hash, and therefore has only 48 bits of entropy. Short hashes must only be used when arrangements have been made to prevent or be safe if there are collisions, whether accidental or malicious. Only use this where necessary, for instance due to the space required for a full size secure hash being prohibitive.

dickon (changes requested):
LGTM, other than we need to clarify the collision risk with short hashes.

data/avro-schema/src/main/resources/avro/net/corda/data/crypto/ShortHashes.avsc
```diff
@@ -0,0 +1,15 @@
+{
+  "type": "record",
+  "name": "ShortHashes",
+  "doc": "List of short hashes. A short hash is the first 12 characters of the hex string form of the hash. Short hashes should be used with care as it could lead to collisions",
```
dickon: Should end in a full stop

dickon (dismissed):
LGTM, other than a small punctuation problem in the docstring which is not blocking.

data/avro-schema/src/main/resources/avro/net/corda/data/crypto/ShortHashes.avsc
kyriathar: Done

dickon (dismissed):
LGTM

mnesbit (approved):
OK

PR #847, opened 2023-02-08
CORE-10198: Upgrade to Kotlin 1.8.10.

chrisr3:
Upgrade Kotlin 1.7.21 -> 1.8.10.

Kotlin 1.8 supports Java 8+, and so `kotlin-osgi-bundle` has absorbed the contents of `kotlin-stdlib-jdk7` and `kotlin-stdlib-jdk8`. This allows us to drop our "home-grown" Kotlin OSGi bundles, although we now need to add back the transitive depedency on `org.jetbrains:annotations` that they provided us.

The Kotlin Gradle plugin has also decided to change how the `KotlinCompile` task is configured, just for gits and shiggles.

### DIFF
--- a/application/build.gradle
+++ b/application/build.gradle
@@ -7,7 +7,7 @@ plugins {
 description 'Corda Application'
 
 dependencies {
-    api 'net.corda.kotlin:kotlin-stdlib-jdk8-osgi'
+    api 'org.jetbrains.kotlin:kotlin-osgi-bundle'
     api 'javax.persistence:javax.persistence-api'
 
     api platform(project(':corda-api'))
--- a/build.gradle
+++ b/build.gradle
@@ -1,4 +1,6 @@
 import org.jetbrains.kotlin.gradle.tasks.KotlinCompile
+import static org.jetbrains.kotlin.gradle.dsl.JvmTarget.JVM_11
+import static org.jetbrains.kotlin.gradle.dsl.KotlinVersion.KOTLIN_1_8
 import static org.gradle.api.JavaVersion.VERSION_11
 import static org.gradle.jvm.toolchain.JavaLanguageVersion.of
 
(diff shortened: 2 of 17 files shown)


ronanbrowne:
@chrisr3 So this removes the need for https://github.com/corda/kotlin-stdlib-jdk-osgi/ ?

chrisr3:
> @chrisr3 So this removes the need for https://github.com/corda/kotlin-stdlib-jdk-osgi/ ?

Yes, because as of Kotlin 1.8.0, the `kotlin-osgi-bundle` supports Java 8+ (i.e. has dropped Java 6 support) and so also includes the contents of `kotlin-stdlib-jdk7` and `kotlin-stdlib-jdk8`.,

build.gradle
```diff
@@ -160,16 +162,16 @@ subprojects {
...
-                apiVersion = '1.7'
+                languageVersion = KOTLIN_1_8
+                apiVersion = KOTLIN_1_8
                 verbose = true
-                jvmTarget = javaVersion
-                freeCompilerArgs += [
-                    "-Xjvm-default=all",
-                    "-java-parameters"
-                ]
+                jvmTarget = JVM_11
```
blsemo, line 170: I assume that `javaVersion` (used in the previous version of this file) is a variable. What happened to it?

application/build.gradle
```diff
@@ -7,7 +7,7 @@ plugins {
 description 'Corda Application'
 
 dependencies {
-    api 'net.corda.kotlin:kotlin-stdlib-jdk8-osgi'
+    api 'org.jetbrains.kotlin:kotlin-osgi-bundle'
```
blsemo, line 10: Does that mean there is now an official kotlin OSGi bundle? That would be great!

blsemo (approved):
The change looks good to me - I have a couple of questions, though.

build.gradle
chrisr3: > I assume that `javaVersion` (used in the previous version of this file) is a variable. What happened to it?

It still exists, but is not assignable to the `jvmTarget` option in the Kotlin plugin's new DSL :facepalm:.

application/build.gradle
chrisr3: > Does that mean there is now an official kotlin OSGi bundle? That would be great!

There was _always_ an official Kotlin OSGi bundle, and we used it. The problem was that it previously targeted Java 6 and so did not include the contents of the additional `kotlin-stdlib-jdk7` and `kotlin-stdlib-jdk8` artifacts. However, Kotlin 1.8.x targets Java 8 and so the official OSGi bundle is now _finally_ complete.

build.gradle
blsemo: So do we still use it elsewhere, and now need to make sure things are in sync? Or can it be deleted?

chrisr3: > So do we still use it elsewhere, and now need to make sure things are in sync? Or can it be deleted?

We use it elsewhere. This isn't ideal but it is good enough for now.

ronanbrowne (approved):
Lgtm

PR #850, opened 2023-02-08
NOTICK Interop/ merge latest main into interop branch

szymonsztuka:
Keep Interop branch up-to-date


ronanbrowne:
@szymonsztuka Im just passing through here as this caught my eye .... do you folks intend on having a CI where this interop branch publishes/runs full tests suits etc along with a runtime-os branch.

As far as I can see right now that does not exist beyond this PR. as we have particular naming constraints on branches which build on jenkins.

Also versions would need to be updated so as not to throttle "real" mainline artifacts on artifactory if that is the intention

driessamyn:
> @szymonsztuka Im just passing through here as this caught my eye .... do you folks intend on having a CI where this interop branch publishes/runs full tests suits etc along with a runtime-os branch.
> 
> As far as I can see right now that does not exist beyond this PR. as we have particular naming constraints on branches which build on jenkins.
> 
> Also versions would need to be updated so as not to throttle "real" mainline artifacts on artifactory if that is the intention

On a slightly related note. I notice we merge main into this branch rather than rebase.
What's the plan for ultimately merging this into main in a few months time?
I suspect this will be nearly impossible with the merge strategy as a single big PR just won't be realistic. I suspect rebasing, then cherry picking commits for small PRs into main will be the only practical thing.

szymonsztuka:
@driessamyn  our addition will be very limited in scope (2-4 Avro schemas), potentially some new API in corda-api/application, in the end we can create a final branch for PR to main from scratch
@ronanbrowne good question, for now its appears just as any other build from the usual PR, but yes we proabbly need to introduce a marker like "-interop-alpha"

szymonsztuka:
The PR  is superseded by https://github.com/corda/corda-api/pull/859, following advice the version has distinct label postfix '-INTEROP'  ( `cordaProductVersion = 5.0.0-INTEROP`) to distinguish builds of this branch.

PR #860, opened 2023-02-13
CORE-10332,CORE-10333: Rewrite base and application modules as Java 11.

chrisr3:
Rewrite classes and interfaces from `:base` and `:application` modules as pure Java.

There are avoidable breaking changes here, but there are also still lots of Kotlin `inline` functions in `XxxUtils` classes left to deal with.

Note that Java doesn't have a concept of a "primary constructor", so if Corda will need to serialize instances of a Java class then its primary constructor must now be annotated as "@ConstructorForDeserialization`.


application/src/main/java/net/corda/v5/application/flows/FlowContextProperties.java
```diff
@@ -0,0 +1,52 @@
...
+ * Context properties of a {@link Flow} are key value pairs of Strings. They are comprised of two sets of properties, those
+ * set by the platform when a {@link Flow} is started, and those which are added by the CorDapp developer during the execution
+ * of their flow. The latter set are referred to as user context properties. All context properties are immutable once
+ * set.
+ */
+public interface FlowContextProperties {
+    /**
+     * A constant containing the Corda reserved prefix.
+     */
+    String CORDA_RESERVED_PREFIX = "corda."; // constant must be lowercase
```
lankydan, line 16: Should this be `public static final`?

lankydan: Original was:

```kotlin
companion object {
        /**
         * A constant containing the Corda reserved prefix.
         */
        @JvmField
        val CORDA_RESERVED_PREFIX = "corda." // constant must be lowercase
    }
```

chrisr3: `FlowContextProperties` is an `interface`, so `public static final` is implicit. (Anything else would also be a compiler error!)

application/src/main/java/net/corda/v5/application/flows/RestRequestBody.java
```diff
@@ -0,0 +1,49 @@
...
+     * Gets the request body and deserializes it into the given type, using a {@link MarshallingService}.
+     * <p>
+     * The selected {@link MarshallingService} will determine what format data is returned.
+     *
+     * @param marshallingService The {@link MarshallingService} to use to deserialize this request body.
+     * @param clazz The class to deserialize the data into.
+     *
+     * @return An instance of the class populated by the provided input data.
+     */
+    <T> T getRequestBodyAs(@NotNull MarshallingService marshallingService, @NotNull Class<T> clazz);
```
lankydan, line 36: I _think_ this should have `@NotNull` on the return type.

chrisr3: That is equivalent to `<T : Any>` in Kotlin, which isn't what were using... :thinking:.

lankydan: Ok, I think its makes sense to have that restriction here. @mnesbit thoughts. I'll leave some comments on similar code but they depend on what we decide here.

application/src/main/java/net/corda/v5/application/marshalling/MarshallingService.java
```diff
@@ -0,0 +1,54 @@
...
+     * <p>
+     * This method will throw an exception if the provided string does not conform to the expected format of the
+     * service.
+     *
+     * @param input The input string to parse.
+     * @param clazz The type to try and parse the data into.
+     *
+     * @return An instance of the required type containing the input data.
+     */
+    <T> T parse(@NotNull String input, @NotNull Class<T> clazz);
```
lankydan, line 41: Should we use `@NotNull` here too?

application/src/main/java/net/corda/v5/application/marshalling/json/JsonDeserializer.java
```diff
@@ -1,16 +1,18 @@
...
      *
-     * @param jsonRoot The Json object to deserialize. The methods of [JsonNodeReader] should be used to traverse the
+     * @param jsonRoot The Json object to deserialize. The methods of {@link JsonNodeReader} should be used to traverse the
      * Json and extract fields to create an instance of a T to return.
      *
-     * @returns an object of type T which has been deserialized from Json
+     * @return an object of type T which has been deserialized from Json
      */
-    fun deserialize(jsonRoot: JsonNodeReader): T
+    T deserialize(@NotNull JsonNodeReader jsonRoot);
```
lankydan, line 17: `@NotNull`?

application/src/main/java/net/corda/v5/application/marshalling/json/JsonSerializer.java
```diff
@@ -1,15 +1,19 @@
...
+public interface JsonSerializer<T> {
     /**
      * Method called when an object of type T should be serialized.
      *
      * @param item The object to serialize.
      * @param jsonWriter An interface to a writer of Json which should be used to translate fields in item of type T
      * to Json.
      */
-    fun serialize(item: T, jsonWriter: JsonWriter)
+    void serialize(T item, @NotNull JsonWriter jsonWriter) throws IOException;
```
lankydan: Are we sure we want to add `IOException` here? Especially since its a checked exception.

lankydan: Do we want `@NotNull T item`?

application/src/main/java/net/corda/v5/application/uniqueness/model/UniquenessCheckStateRef.java
```diff
@@ -1,20 +1,23 @@
...
  * Please note that this representation of a state ref is entirely different from the Ledger specific
- * [StateRef][net.corda.v5.ledger.utxo.StateRef] class. This class represents a state ref that the
+ * {@link net.corda.v5.ledger.utxo.StateRef} class. This class represents a state ref that the
  * uniqueness checker will process and may differ from the UTXO ledger data model.
  */
 @CordaSerializable
-interface UniquenessCheckStateRef {
-    val txHash: SecureHash
-    val stateIndex: Int
+public interface UniquenessCheckStateRef {
```
lankydan, line 18: Not for this PR, but @relyafi @nkovacsx should the uniqueness parts be in `ledger`? Especially since it mentions state ref (although not by class)

base/src/main/java/net/corda/v5/base/util/EncodingUtils.java
```diff
@@ -0,0 +1,134 @@
...
+public final class EncodingUtils {
+    private EncodingUtils() {}
+
+    // byte[] encoders
+
+    /**
+     * Convert a byte array to a Base58 encoded {@link String}.
+     */
+    @NotNull
+    public static String toBase64(@NotNull byte[] bytes) {
```
lankydan, line 20: Are we not going to provide extension function versions of these functions?

ledger/ledger-common/src/main/java/net/corda/v5/ledger/common/transaction/PrivacySalt.java
```diff
@@ -0,0 +1,19 @@
...
+import net.corda.v5.base.annotations.DoNotImplement;
+import org.jetbrains.annotations.NotNull;
+
+/**
+ * A privacy salt is required to compute nonces per transaction component in order to ensure that an adversary cannot
+ * use brute force techniques and reveal the content of a Merkle-leaf hashed value.
+ */
+@CordaSerializable
+@DoNotImplement
+public interface PrivacySalt{
```
lankydan: Need a space here :)

application/src/main/java/net/corda/v5/application/marshalling/MarshallingService.java
chrisr3: Same reasoning as above. You are effectively asking for `<T : Any>` here - which we _might_ be able to do, but isn't what we had.

application/src/main/java/net/corda/v5/application/marshalling/json/JsonSerializer.java
chrisr3: We were throwing `IOException` anyway without realising it (assuming the Javadocs were telling the truth about `JsonWriter`'s methods).

application/src/main/java/net/corda/v5/application/uniqueness/model/UniquenessCheckStateRef.java
chrisr3: Answers on a postcard, please :envelope:.

base/src/main/java/net/corda/v5/base/util/EncodingUtils.java
chrisr3: We cannot allow the OSGi metadata to include a _single_ `kotlin` reference. I have therefore ported these API functions to Java as the only viable alternative. (Java developers can thank me later... :wink:)

Extension function versions would need to exist _outside_ the API, or in a standalone bundle at the very least.

application/src/main/java/net/corda/v5/application/marshalling/json/JsonSerializer.java
chrisr3: I've removed the `IOException` :+1:.

application/src/main/java/net/corda/v5/application/flows/RestRequestBody.java
mnesbit: I think it should have been not nullable too

application/src/main/java/net/corda/v5/application/marshalling/MarshallingService.java
mnesbit: Here it is advantageous to be nullable and return null for null

mnesbit (dismissed):
Lets get this in even if minor refinements may occur

application/src/main/java/net/corda/v5/application/flows/RestRequestBody.java
chrisr3: :+1:

application/src/main/java/net/corda/v5/application/flows/FlowContextProperties.java
lankydan: 👍

base/src/main/java/net/corda/v5/base/util/EncodingUtils.java
lankydan: Ye, thats fine. I just mean for some we still kept the extension function variations, but not for these. Will be add the extension function versions back again later?

lankydan (approved):

mnesbit (approved):
OK

PR #861, opened 2023-02-14
CORE-9969 Add a new UniquenessCheckErrorUnhandledException type

nkovacsx:
### Overview

Add a new uniqueness checker error type called `UniquenessCheckErrorUnhandledException`. This will be returned by the uniqueness checker client service when an unexpected exception was encountered in the uniqueness checker.

This is needed so the non-validating notary plugin server can return a general error when this is encountered.


application/src/main/kotlin/net/corda/v5/application/uniqueness/model/UniquenessCheckErrors.kt
```diff
@@ -44,3 +44,9 @@ interface UniquenessCheckErrorTimeWindowOutOfBounds : UniquenessCheckError {
 interface UniquenessCheckErrorMalformedRequest : UniquenessCheckError {
     val errorText: String
 }
+
+/** Occurs when there's an unhandled exception in the uniqueness checker. */
+interface UniquenessCheckErrorUnhandledException : UniquenessCheckError {
+    val unhandledExceptionType: String
+    val unhandledExceptionMessage: String
```
relyafi: Why aren't we using an `errorText` and `cause` (throwable) as we do for `NotaryErrorGeneral`? I can see in the runtime-os PR we're having to jump through hoops to reverse this conversion.

lankydan: In general, I'm not convinced that we need the exception itself.

We just need the error message itself right? The exception should be thrown and logged in the uniqueness checker worker (whatever does that), then the notary server flow gets that back and can choose to log if it wants, otherwise passing the error back to the client flow.

There is no use in having the literal exception unless we expect the client flow to do something with it, even then an enum or string to switch over can suffice.

nkovacsx: We must do the conversion somewhere if we want to add the exception as a cause. 
If you check the avro message it only returns type and message as `Strings` so we need to create a `Throwable` from it to attach it as a cause. 
If we don't need it as a cause, we can skip this and just store message/type as a string and that's it. However, the quoting from the ticket: `return as a NotaryErrorGeneral (with this as the cause)`.

relyafi: Ok, so having discussed with @nkovacsx, it looks like it's going to be impossible to serialize the Throwable over Avro.

Regarding the exception, remember the uniqueness checker is living in a different processor / worker to the notary server flow. So there is a message bus boundary between these. So don't we need to wrap that exception sensibly in order to return a legitimate external event response back to the notary server flow?

relyafi: @lankydan having discussed again with @nkovacsx , yes I understand your point. I think it would be sufficient to simply return the message (albeit as part of the new unhandled exception class)

lankydan (dismissed):

relyafi (dismissed):
LGTM, nice pivot to Java too!

relyafi (approved):

lankydan (approved):

PR #867, opened 2023-02-15
CORE-10226 Rename TransactionVerificationException to TransactionSignatureException and remove it from Consensual verifications.

vlajos:
CORE-10226 Rename `TransactionVerificationException` to `TransactionSignatureException`, convert it to java and remove it from Consensual verifications.

Runtime-os PR: https://github.com/corda/corda-runtime-os/pull/3156


mbrkic-r3 (dismissed):

ledger/ledger-common/src/main/kotlin/net/corda/v5/ledger/common/transaction/TransactionSignatureException.kt
```diff
@@ -0,0 +1,12 @@
...
+
+import net.corda.v5.base.exceptions.CordaRuntimeException
+import net.corda.v5.crypto.SecureHash
+
+/**
+ * Indicates a problem related to signatures of the transaction named by [transactionId].
+ *
+ * @property transactionId the Merkle root hash (identifier) of the transaction that failed verification.
+ */
+open class TransactionSignatureException(val transactionId: SecureHash, message: String, cause: Throwable?) :
```
lankydan: Should make this a Java class.

ledger/ledger-consensual/src/main/kotlin/net/corda/v5/ledger/consensual/transaction/ConsensualTransactionVerificationException.kt
```diff
@@ -0,0 +1,16 @@
...
+
+import net.corda.v5.base.exceptions.CordaRuntimeException
+import net.corda.v5.crypto.SecureHash
+
+/**
+ * Indicates that some aspect of the transaction named by [transactionId] violates some rules.
+ *
+ * @property transactionId the Merkle root hash (identifier) of the transaction that failed verification.
+ */
+open class ConsensualTransactionVerificationException(
```
lankydan: Why do we need a consensual specific exception?

lankydan: The KDoc of `ConsensualState.verify` says:

```kotlin
/**
     * An override of this function needs to be provided to:
     *  - verify the state's well-formedness
     *  - verify compatibility with the other states of the encapsulating transaction
     *  - check required signing keys
     *  - check the transaction's timestamp.
     *
     * TODO(make services injectable (crypto, etc... CORE-5995)
     *
     * @param ledgerTransaction encapsulating transaction
     *
     * @throws [net.corda.v5.ledger.common.transaction.TransactionVerificationException] if the verification failed.
     *
     */
    fun verify(ledgerTransaction: ConsensualLedgerTransaction)
```

Note, that it says `TransactionVerificationException` is thrown if verification fails.

However, the implementation does this.

```kotlin
private fun verifyStates() {
        transaction.states.map { it.verify(transaction) }
    }
```

Meaning that there is no relation between the type of exception that `ConsensualState.verify` throws and the correct behaviour of the code.

I say we remove `ConsensualTransactionVerificationException`/previously `TransactionVerificationException` altogether because visible benefits. Furthermore, it makes it less like the UTXO version which we generally throw `IllegalStateException`s or `IllegalArgumentException`s, the UTXO version also doesn't care in general what is thrown.

ledger/ledger-common/src/main/java/net/corda/v5/ledger/common/transaction/TransactionSignatureException.java
```diff
@@ -0,0 +1,43 @@
...
+     * Creates a new instance of the TransactionSignatureException class.
+     *
+     * @param transactionId the Merkle root hash (identifier) of the transaction that failed verification.
+     * @param message The details of the current exception to throw.
+     * @param cause The underlying cause of the current exception.
+     */
+    public TransactionSignatureException(
+            @NotNull final SecureHash transactionId,
+            @NotNull final String message,
+            @Nullable final Throwable cause
```
lankydan: Can we remove the final

lankydan:
Remember to change the PR title / commit description later

ledger/ledger-common/src/main/kotlin/net/corda/v5/ledger/common/transaction/TransactionSignatureException.kt
vlajos: Thanks. Done.

ledger/ledger-consensual/src/main/kotlin/net/corda/v5/ledger/consensual/transaction/ConsensualTransactionVerificationException.kt
vlajos: I have updated it.

ledger/ledger-common/src/main/java/net/corda/v5/ledger/common/transaction/TransactionSignatureException.java
vlajos: Done.

lankydan (approved):

PR #871, opened 2023-02-16
CORE-9630 Finality takes a transaction builder

lankydan:
`UtxoLedgerService.finalize` and `ConsensualLedgerService.finalize` both
now take in their respective transaction builders instead of signed
transactions.

The respective finality flows now sign the builders.

`toSignedTransaction` has been made internal and removed from the API.

The "cant sign twice" code has been removed from the builders because
we control the signing of them, making the code redundant (as it was there
to help developers).

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -7,7 +7,7 @@ cordaProductVersion = 5.0.0
 # NOTE: update this each time this module contains a breaking change
 ## NOTE: currently this is a top level revision, so all API versions will line up, but this could be moved to
 ##   a per module property in which case module versions can change independently.
-cordaApiRevision = 647
+cordaApiRevision = 649
 
 # Main
 kotlinVersion = 1.8.10
--- a/ledger/ledger-consensual/src/main/kotlin/net/corda/v5/ledger/consensual/ConsensualLedgerService.kt
+++ b/ledger/ledger-consensual/src/main/kotlin/net/corda/v5/ledger/consensual/ConsensualLedgerService.kt
@@ -43,16 +43,16 @@ interface ConsensualLedgerService {
     fun findLedgerTransaction(id: SecureHash): ConsensualLedgerTransaction?
 
     /**
-     * Collects signatures, records and broadcasts to involved peers a [ConsensualSignedTransaction].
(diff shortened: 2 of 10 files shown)


vlajos:
The interop team mentioned that they will need a way to save draft transactions somehow. I think that means that they will need a way to get a hold on the transaction id somehow before signing it. I am not too sure about the details and its current priority, but if the transaction builder gets finalized instead of the signed transaction, this will be more tricky. 
@jmacmahonr3 @mnesbit

jmacmahonr3:
> am not too sure about the details and its current priority, but if the transaction builder gets finalized instead of the signed transaction, this will be more tric

Hi @vlajos, @lankydan & @mnesbit 
The ability to advertise an unsigned tx or a draft tx as we call it, is critical for alot of cross network patterns and will be our biggest ask for 5.1

lankydan:
Ok, @mnesbit, @blsemo what do we want to do in this case? If we want to provide the functionality that @jmacmahonr3 is asking for then we either: scrap this PR or introduce a callback that executes before signing the transaction. I'd say leaving the current code is probably best because the callback is just a different way of signing it before passing it into `finalize`. That being said, if most people are ok with the builder being passed in, then having the callback to cover @jmacmahonr3's scenario doesn't sound too bad.

blsemo:
> Hi @vlajos, @lankydan & @mnesbit The ability to advertise an unsigned tx or a draft tx as we call it, is critical for alot of cross network patterns and will be our biggest ask for 5.1

Unsigned as in completely unsigned? Or is signed by the initiator ok? In the latter case, just scrapping this PR would do it, in the former, we need to introduce a new intermediate format (or the callback Dan mentioned).

jmacmahonr3:
> 

Completely unsigned. We need something that represents a transaction proposal that a later notarized transaction (or fully signed transaction ) can be proven to be fulfilling. In classic c4's utxo ledger this has traditionally meant we have a notary's signature over the tx id

lankydan:
Closing this PR. Discussed with @blsemo @mnesbit to not do this as not to cause issues with interop.

PR #873, opened 2023-02-20
CORE-10461 Categorise notary errors and move notary exception to API

nkovacsx:
### Overview

Notary errors and exception has been merged. There will be only one main abstract class called `NotaryException`. This class has a private constructor so only the inner classes will be able to inherit from it, these are:
- `NotaryExceptionFatal`
- `NotaryExceptionUnknown`

These two are the main notary exception types and should be further extended in the implementation project.


ledger/notary-plugin/src/main/kotlin/net/corda/v5/ledger/notary/plugin/core/NotaryException.kt
```diff
@@ -0,0 +1,19 @@
...
+import net.corda.v5.base.exceptions.CordaRuntimeException
+import net.corda.v5.crypto.SecureHash
+
+/**
+ * Exception thrown by the notary plugin if any issues are encountered while trying to commit a transaction. The
+ * underlying [error] specifies the cause of failure.
+ */
+@CordaSerializable
+class NotaryException(
+    /** Cause of notarisation failure. */
```
vlajos: I think `@property` needs to be used.

vlajos:
I think new API structures should be written in java.

blsemo:
I have a question on the design - why do we add a NotaryError under the NotaryException? This seems like a lot of code just to say "This exception is retryable and that one isn't".
Wouldn't it just be easier to have an exception hierarchy, i.e. have `open class NotaryException: CordaRuntimeException` and then a `class FatalNotaryException: NotaryException` and `class UnkownNotaryException: NotaryException`?

nkovacsx:
@blsemo That sounds reasonable, I can't remember what was the reason to have a separate notary exception but I'll double check and make the change if possible.

relyafi:
To an extent, the concept of an exception with error is a carry-over from C4 I believe, in that it allows us to return the error with additional context, in this case, a transaction id.

The value would be if the client wanted to act upon the error, e.g. so that they could then easily resolve to a transaction in their vault to do something with it. But it looks like we are absorbing it within finality flow anyway, so we probably could just encode this info in the error message and not have a separate field for it.

nkovacsx:
I'm not sure what is the current state of classes in the `corda-api` project but if we want to have a `NotaryException` in the API we unfortunately have to make it a class because interfaces can't inherit from exception classes. I know it's not something we want to do regularly but maybe it's okay for exceptions. @mnesbit @blsemo ?

ledger/notary-plugin/src/main/java/net/corda/v5/ledger/notary/plugin/core/NotaryException.java
```diff
@@ -0,0 +1,58 @@
...
+import net.corda.v5.base.annotations.CordaSerializable;
+import net.corda.v5.base.exceptions.CordaRuntimeException;
+import net.corda.v5.crypto.SecureHash;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+/**
+ * Representation of errors that can be returned by the notary (plugin). This is only a marker interface, the plugins
+ * can define their own errors by implementing this interface. Please refer to the non-validating notary plugin for
+ * a more detailed example.
```
relyafi: Looks like a copy / paste from the `NotaryError` file, but it doesn't make sense - this isn't an interface. You probably need to start from scratch given the change in structure.

ledger/notary-plugin/src/main/java/net/corda/v5/ledger/notary/plugin/core/NotaryException.java
```diff
@@ -0,0 +1,58 @@
...
+        public NotaryExceptionFatal(@NotNull String notaryErrorMessage, @Nullable SecureHash txId) {
+            super(notaryErrorMessage, txId);
+        }
+    }
+    @CordaSerializable
+    public abstract static class NotaryExceptionUnknown extends NotaryException {
+        public NotaryExceptionUnknown(@NotNull String notaryErrorMessage, @Nullable SecureHash txId) {
+            super(notaryErrorMessage, txId);
+        }
+    }
```
relyafi: These should all have Java doc comments - especially the last two which are part of public API.

ledger/notary-plugin/src/main/java/net/corda/v5/ledger/notary/plugin/core/NotaryException.java
```diff
@@ -0,0 +1,58 @@
...
+
+
+    private NotaryException(@NotNull String notaryErrorMessage, @Nullable SecureHash txId) {
+        super("Unable to notarise transaction " + (txId != null ? txId : "<Unknown>:") + " " + notaryErrorMessage);
+        this.notaryErrorMessage = notaryErrorMessage;
+        this.txId = txId;
+    }
+
+    @CordaSerializable
+    public abstract static class NotaryExceptionFatal extends NotaryException {
```
relyafi: This and the unknown errors feel a bit clunky; callers now have to qualify the exception, e.g. `NotaryException.NotaryExceptionFatal`. If the intent is to replicate Kotlin sealed class functionality to prevent direct inheritance from `NotaryException`, you might want to consider making `NotaryException` "package private", i.e. specifying no qualifer. This will have a similar effect, in that whilst it can be extended, it could only be extended by R3 (only we should be using the `net.corda.v5.ledger.notary.plugin.core` package, and presumably sandboxing enforces this).

This would also allow you to define the fatal / unknown classes in their own files in this package and make things a bit cleaner.

If we do keep as-is, consider renaming to `Fatal` and `Unknown`, so at least then to the caller it appears as `NotaryException.Fatal`, which avoids duplication.

relyafi (changes requested):

ledger/notary-plugin/src/main/java/net/corda/v5/ledger/notary/plugin/core/NotaryException.java
nkovacsx: I do agree with it being clunky like this and happy to modify it to `Fatal` and `Unknown`. I'm not a big fan of using package private but if you feel like that would make it cleaner I can make the change.

relyafi: Maybe worth getting others opinions too - people like @lankydan and @MrMatthewLayton are probably better judges than me.

nkovacsx: I made the change but can make it back if the others feel like it is better. You have a point that static inner classes are not really java style and can look weird if making a static import.

Like let's say someone statically imports the `NotaryException` and then they only write `Fatal` and `Unknown` without the `NotaryException.` prefix that can lead to major confusion too.

ledger/notary-plugin/src/main/java/net/corda/v5/ledger/notary/plugin/core/NotaryException.java
```diff
@@ -0,0 +1,46 @@
...
+ * or 
+ * @see NotaryExceptionUnknown
+ */
+@CordaSerializable
+abstract class NotaryException extends CordaRuntimeException {
+
+    @NotNull
+    private final String notaryErrorMessage;
+    @Nullable
+    private final SecureHash txId;
```
chrisr3: I don't think you need to annotate private Java fields because the Kotlin compiler is never going to see these.

nkovacsx: Yeah that's a good point!

relyafi (dismissed):
LGTM

lankydan (approved):

PR #880, opened 2023-02-21
CORE-9980 send/receive transaction builder

vlajos:
Send and receive transaction builder functionality.

Example usage:
Initiator:
```kotlin
val updatedTxBuilder = utxoLedgerService.sendAndReceiveTransactionBuilder(txBuilder, session)
```

Receiver:
```kotlin
val proposedTxBuilder = utxoLedgerService.receiveTransactionBuilder(session)
proposedTxBuilder.add...(...)
proposedTxBuilder.add...(...)
proposedTxBuilder.add...(...)
utxoLedgerService.sendUpdatedTransactionBuilder(proposedTxBuilder, session)
```
proposedTxBuilder tracks diffs internally with a smarter transactionbuilder implementation.

Design summary: https://github.com/corda/platform-eng-design/pull/545

Runtime-os: https://github.com/corda/corda-runtime-os/pull/3195

### DIFF
--- a/ledger/ledger-utxo/src/main/java/net/corda/v5/ledger/utxo/UtxoLedgerService.java
+++ b/ledger/ledger-utxo/src/main/java/net/corda/v5/ledger/utxo/UtxoLedgerService.java
@@ -125,4 +125,62 @@ public interface UtxoLedgerService {
             @NotNull FlowSession session,
             @NotNull UtxoTransactionValidator validator
     );
+
+    /**
+     * Sends a transaction builder to another session, then it waits for other side to propose transaction builder components,
+     * then applies the proposed components to a copy of the original builder and returns that new builder.
+     * <p>
+     * It supports similar workflows:
+     * Initiator:
+     * <p>
+     * val updatedTxBuilder = utxoLedgerService.sendAndReceiveTransactionBuilder(txBuilder, session)
+     * <p>
+     * The proposal one's notary and time window will get discarded and the original will be kept if both the original and
+     * the proposal have these components set. Also, all duplications will be discarded.
+     * <p>
+     * Receiver:
+     * <p>
+     * val proposedTxBuilder = utxoLedgerService.receiveTransactionBuilder(session)
+     * proposedTxBuilder.add...(...)
+     * proposedTxBuilder.add...(...)
+     * proposedTxBuilder.add...(...)
+     * utxoLedgerService.replyTransactionBuilderProposal(proposedTxBuilder, session)
+     *
+     * @param transactionBuilder The {@link UtxoTransactionBuilder} to send.
+     * @param session The receiver {@link FlowSession]}.
+     *
+     * @return A new merged builder of the original and proposed components.
+     */
+    @NotNull
+    @Suspendable
+    UtxoTransactionBuilder sendAndReceiveTransactionBuilder(
+            @NotNull UtxoTransactionBuilder transactionBuilder,
+            @NotNull FlowSession session
+    );
+
+    /**
+     * Receives a transaction builder from another session.
+     *
+     * @param session The {@link FlowSession] to receive the {@link UtxoTransactionBuilder} from.
+     */
+    @NotNull
+    @Suspendable
+    UtxoTransactionBuilder receiveTransactionBuilder(
+            @NotNull FlowSession session
+    );
+
+    /**
+     * Sends the differences of transaction builders to another session with all dependent back chains.
+     * It works only with {@link #receiveTransactionBuilder(FlowSession)} created {@link UtxoTransactionBuilder}s
+     * which track the differences internally.
+     * If it is called with anything else, it throws [InvalidParameterException].
+     * <p>
+     * @param transactionBuilder The {@link UtxoTransactionBuilder} to send.
+     * @param session The receiver {@link FlowSession}.
+     */
+    @Suspendable
+    void sendUpdatedTransactionBuilder(
+            @NotNull UtxoTransactionBuilder transactionBuilder,
+            @NotNull FlowSession session
+    );
 }
--- a/ledger/ledger-utxo/src/main/java/net/corda/v5/ledger/utxo/UtxoLedgerService.java
+++ b/ledger/ledger-utxo/src/main/java/net/corda/v5/ledger/utxo/UtxoLedgerService.java
@@ -135,7 +135,7 @@ public interface UtxoLedgerService {
      * <p>
      * val updatedTxBuilder = utxoLedgerService.sendAndReceiveTransactionBuilder(txBuilder, session)
      * <p>
-     * The proposal one's notary and time window will get discarded and the original will be kept if both the original and
+     * The notary and time window from the proposal will get discarded and the original will be kept if both the original and
      * the proposal have these components set. Also, all duplications will be discarded.
      * <p>
      * Receiver:
@@ -171,7 +171,7 @@ public interface UtxoLedgerService {
 
     /**
      * Sends the differences of transaction builders to another session with all dependent back chains.
-     * It works only with {@link #receiveTransactionBuilder(FlowSession)} created {@link UtxoTransactionBuilder}s
+     * It works only with {@link UtxoTransactionBuilder}s created from {@link #receiveTransactionBuilder(FlowSession)}
      * which track the differences internally.
      * If it is called with anything else, it throws [InvalidParameterException].
      * <p>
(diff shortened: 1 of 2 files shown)


ledger/ledger-utxo/src/main/kotlin/net/corda/v5/ledger/utxo/UtxoLedgerService.kt
```diff
@@ -120,4 +121,79 @@ interface UtxoLedgerService {
...
+     * @param originalTransactionalBuilder The received parts will be appended to this [UtxoTransactionBuilder].
+     */
+    @Suspendable
+    fun receiveAndUpdateTransactionBuilder(
+        session: FlowSession,
+        originalTransactionalBuilder: UtxoTransactionBuilder
+    ): UtxoTransactionBuilder
+
+    /**
+     * Sends a transaction builders to another session, then it waits for other side returning an updated builder.
```
blsemo: This one needs to ensure that the returned tx builder is strictly an addition to the one passed in - this is probably the case, given that we send diffs, but should be documented as API promise.

I wonder whether we should also have a callback to check that we agree with the diff that is returned? Or would CorDapps somehow have to check the returned tx builder? Would this modify the one passed in or return a modified copy, so we can do a diff?

ledger/ledger-utxo/src/main/kotlin/net/corda/v5/ledger/utxo/UtxoLedgerService.kt
```diff
@@ -120,4 +121,79 @@ interface UtxoLedgerService {
...
+
+    /**
+     * Receives a transaction builder from another session.
+     *
+     * @param session The [FlowSession] to receive the [UtxoTransactionBuilder] from.
+     */
+    @Suspendable
+    fun receiveTransactionBuilder(
+        session: FlowSession,
+    ): UtxoTransactionBuilder
```
blsemo: These functions are just syntactic sugar over the diff ones below - I wonder whether they have to be on the API or could be just static helper functions? Maybe we could build the whole dealing with receiving/sending as helper functions, if we only have the diff-sending/receiving functions on the API?

ledger/ledger-utxo/src/main/kotlin/net/corda/v5/ledger/utxo/UtxoLedgerService.kt
```diff
@@ -120,4 +121,79 @@ interface UtxoLedgerService {
...
+    )
+
+    /**
+     * Receives a transaction builder from another session then appends its parts to an existing one.
+     *
+     * @param session The [FlowSession] to receive the [UtxoTransactionBuilder] from.
+     * @param originalTransactionalBuilder The received parts will be appended to this [UtxoTransactionBuilder].
+     */
+    @Suspendable
+    fun receiveAndUpdateTransactionBuilder(
```
blsemo: I don't like that name - I think it should be `receiveTransactionBuilderDiff()`, to be symmetrical with the sending function. It should still apply the diff to a passed in tx builder.

blsemo (commented):
I think we need to work on the API - it would be unfortunate all possible APIs on the API - remember we can never remove anything that has been released in 5.0.

Also, I think if we have that many functions, we need a separate service for sending/receiving tx builders (could be returned from the utxo ledger service).

vlajos:
> I think we need to work on the API - it would be unfortunate all possible APIs on the API - remember we can never remove anything that has been released in 5.0.

I plan to remove the unneded ones after we will have chosen the ones to keep.

ledger/ledger-utxo/src/main/kotlin/net/corda/v5/ledger/utxo/UtxoLedgerService.kt
vlajos: I have renamed it.

ledger/ledger-utxo/src/main/kotlin/net/corda/v5/ledger/utxo/UtxoLedgerService.kt
vlajos: > This one needs to ensure that the returned tx builder is strictly an addition to the one passed in - this is probably the case, given that we send diffs, but should be documented as API promise.

Yes. I have updated the doc to reflect this better.
It sends only the diff on sender side, but also does not apply again anything which is already in the original builder on the receiver side.

> 
> I wonder whether we should also have a callback to check that we agree with the diff that is returned? Or would CorDapps somehow have to check the returned tx builder? 
My original thinking was that the received builder can be used as a container only to let the cordapp cherry-pick its components into another builder, if the receiver side wants to control what it wants to accept and what not.
Then I just realized that only the `notary` is exposed on the `UtxoTransactionBuilder` interface.
So it is impossible to do any kind of diffing in the CorDapp.
If we want to give control, I would expose components on the `UtxoTransactionBuilder` and let the CorDapp decide how they want to cherry-pick. 

> Would this modify the one passed in or return a modified copy, so we can do a diff?

It creates a new builder from the original and received components.
Notary and TimeWindow gets overwritten by the new ones. The other components are appended.

ledger/ledger-utxo/src/main/kotlin/net/corda/v5/ledger/utxo/UtxoLedgerService.kt
vlajos: I have removed the discarded options.

mnesbit (dismissed):
LGTM

ledger/ledger-utxo/src/main/kotlin/net/corda/v5/ledger/utxo/UtxoLedgerService.kt
```diff
@@ -120,4 +121,59 @@ interface UtxoLedgerService {
...
+    /**
+     * Sends the differences of transaction builders to another session with all dependent back chains.
+     * It works only with [receiveTransactionBuilder] created [UtxoTransactionBuilder]s which track the differences internally.
+     * If it is called with anything else, it throws [InvalidParameterException].
+     *
+     * @param transactionBuilder The [UtxoTransactionBuilder] to send.
+     * @param session The receiver [FlowSession].
+     */
+    @Suspendable
+    fun replyTransactionBuilderProposal(
```
lankydan: I would prefer this was called `sendUpdatedTransactionBuilder`, but happy to see what others think.

lankydan (dismissed):

ledger/ledger-utxo/src/main/java/net/corda/v5/ledger/utxo/UtxoLedgerService.java
```diff
@@ -125,4 +125,62 @@ UtxoSignedTransaction receiveFinality(
...
+    /**
+     * Sends a transaction builder to another session, then it waits for other side to propose transaction builder components,
+     * then applies the proposed components to a copy of the original builder and returns that new builder.
+     * <p>
+     * It supports similar workflows:
+     * Initiator:
+     * <p>
+     * val updatedTxBuilder = utxoLedgerService.sendAndReceiveTransactionBuilder(txBuilder, session)
+     * <p>
+     * The proposal one's notary and time window will get discarded and the original will be kept if both the original and
```
blsemo:
```suggestion
     * The notary and time window from the proposal will get discarded and the original will be kept if both the original and
```

ledger/ledger-utxo/src/main/java/net/corda/v5/ledger/utxo/UtxoLedgerService.java
```diff
@@ -125,4 +125,62 @@ UtxoSignedTransaction receiveFinality(
...
+     */
+    @NotNull
+    @Suspendable
+    UtxoTransactionBuilder receiveTransactionBuilder(
+            @NotNull FlowSession session
+    );
+
+    /**
+     * Sends the differences of transaction builders to another session with all dependent back chains.
+     * It works only with {@link #receiveTransactionBuilder(FlowSession)} created {@link UtxoTransactionBuilder}s
```
blsemo:
```suggestion
     * It works only with {@link UtxoTransactionBuilder}s created from {@link #receiveTransactionBuilder(FlowSession)}
```

blsemo (commented):
Looks fine to me apart from the proposed wording changes on the KDocs.

blsemo (approved):
LGTM

PR #881, opened 2023-02-21
CORE-10336: port crypto API to Java

dickon:
CORE-10336: Rewrite crypto API in Java, and clean it up

This is part of a generation transition of all (non-test) Kotlin code to Java.

- Add NotNull and Nullable annotations.  
- Add equals/hashcode/toString for classes which used to be Kotlin data classes.
- Remove Kotlin extension methods since there is no Java equivalent.

Also, while we are making a significant API change:

- Remove limit constants, which we no longer regard as part of the API (but may still document).
- Remove all places we recommend a digest digest; it is up to the user to pick something appropriate to their needs.

Runtime OS PR: https://github.com/corda/corda-runtime-os/pull/3227


ledger/ledger-utxo/src/main/kotlin/net/corda/v5/ledger/utxo/StateRef.kt
```diff
@@ -17,7 +17,7 @@ data class StateRef(val transactionId: SecureHash, val index: Int) {
         /**
          * Parses the specified value into a new [StateRef] instance.
          *
-         * @param value The value to parse into a [StateRef] instance.
+ gi        * @param value The value to parse into a [StateRef] instance.
```
driessamyn: typo

crypto/src/main/java/net/corda/v5/crypto/CompositeKey.java
```diff
@@ -19,30 +20,41 @@
...
 
     /**
      * Takes single [PublicKey] and checks if [CompositeKey] requirements hold for that key.
+     * 
+     * @param key the public key
+     * @return true if the public key is a composite key, false otherwise
+     * 
      */
-    fun isFulfilledBy(key: PublicKey): Boolean
+    Boolean isFulfilledBy(PublicKey key);
```
chrisr3: Actually:
```
Boolean -> boolean
Boolean? -> @Nullable Boolean
```
Similarly:
```
Int -> int
Int? -> @Nullable Integer

Long -> long
Long? -> @Nullable Long

Short -> short
Short? -> @Nullable Short

Byte -> byte
Byte? -> @Nullable Byte
```

crypto/src/main/java/net/corda/v5/crypto/CompositeKey.java
```diff
@@ -19,30 +20,41 @@
...
-    fun isFulfilledBy(keysToCheck: Iterable<PublicKey>): Boolean
+    Boolean isFulfilledBy(Iterable<PublicKey> keysToCheck);
 
     /**
      * Set of all leaf keys of that [CompositeKey].
      */
-    val leafKeys: Set<PublicKey>
-}
\ No newline at end of file
+    Set<PublicKey> getLeafKeys();
```
chrisr3, line 57: You've forgotten the `@NotNull` annotation.

crypto/src/main/java/net/corda/v5/crypto/CordaOID.java
```diff
@@ -0,0 +1,34 @@
+package net.corda.v5.crypto;
+
+/**
+ * OIDs used for the Corda platform. All entries MUST be defined in this file only and they MUST NOT be removed.
+ * If an OID is incorrectly assigned, it should be marked deprecated and NEVER be reused again.
+ */
+
+public final class CordaOID {
```
chrisr3, line 9: Add a private no-argument constructor so that no-one can instantiate `CordaOID`.

crypto/src/main/java/net/corda/v5/crypto/SecureHash.java
```diff
@@ -0,0 +1,97 @@
+package net.corda.v5.crypto;
+
+import kotlin.jvm.internal.Intrinsics;
```
chrisr3: You cannot use any Kotlin functions.

crypto/src/main/kotlin/net/corda/v5/crypto/PublicKeyHash.kt
```diff
@@ -1,7 +1,7 @@
-@file:JvmName("PublicKeyUtils")
 package net.corda.v5.crypto
```
chrisr3: This is still a Kotlin file. And is this where `KeyUtils.java` came from? It looks like its original name was actually `PublicKeyUtils`.

There cannot be _any_ Kotlin references left in `main`.

chrisr3 (changes requested):
This isn't complete yet. And don't forget KDoc -> Javadoc too.

crypto/src/test/kotlin/net/corda/v5/crypto/SecureHashTests.kt
```diff
@@ -12,12 +12,9 @@ class SecureHashTests {
     @Test
     fun `toHexString Should HEX representation of digest`() {
         val data = "abc".toByteArray()
-        val algorithm = DigestAlgorithmName.SHA2_384.name
+        val algorithm = DigestAlgorithmName.SHA2_384.name;
```
anien: This is still kotlin file.

chrisr3: To be fair, it doesn't matter if the `test` files are still written in Kotlin. All that matters are the files in `main`.

Having said that, there's no point adding a semi-colon to a Kotlin file either :wink:.

anien: I should be more specific. I meant to remove `;` as it is still written in kotlin.

crypto/src/main/java/net/corda/v5/crypto/CompositeKey.java
dickon: fixed. At first I thought we'd remove them, and we've decided to keep them now.

crypto/src/test/kotlin/net/corda/v5/crypto/SecureHashTests.kt
dickon: yes, done

crypto/src/main/java/net/corda/v5/crypto/CompositeKey.java
dickon: fixed

crypto/src/main/java/net/corda/v5/crypto/CordaOID.java
dickon: why would you want to? `CordaOID` is just a bunch of string constants, there's no logic there and no need I can see to instantiate anything.

crypto/src/main/java/net/corda/v5/crypto/SecureHash.java
dickon: fixed

ledger/ledger-utxo/src/main/kotlin/net/corda/v5/ledger/utxo/StateRef.kt
dickon: removed

crypto/src/main/kotlin/net/corda/v5/crypto/PublicKeyHash.kt
dickon: Ported to Java now (since I'm continuning to work on the branch)

crypto/src/main/java/net/corda/v5/crypto/CordaOID.java
driessamyn: Let's just do it. No downside to it and if it's consistent with what we do elsewhere, that's a good enough reason.

crypto/src/main/java/net/corda/v5/crypto/KeyUtils.java
```diff
@@ -0,0 +1,48 @@
...
+
+public class KeyUtils {
+
+    /**
+     * Constant specifying the maximum number of key lookup by id items.
+     */
+    public static int KEY_LOOKUP_INPUT_ITEMS_LIMIT = 20;
+
+
+    public static boolean containsAny(PublicKey key, Iterable<PublicKey> otherKeys ) {
```
driessamyn: This is probably not a good name now that it isn't an extension function anymore, especially given the order of the arguments. With the arguments swapped the other way around, maybe it would work?

As an aside, I thought we were moving the interfaces/classes first, then extension methods in a second PR?

crypto/src/main/java/net/corda/v5/crypto/KeyUtils.java
```diff
@@ -0,0 +1,48 @@
...
+        } else {
+            for (PublicKey otherKey: otherKeys) {
+                if (otherKey == key) return true;
+            }
+        }
+        return false;
+    }
+
+    /** Return true if [otherKey] fulfils the requirements of [firstKey] */
+    public static boolean isFulfilledBy(PublicKey firstKey, PublicKey otherKey) {
```
driessamyn: As above. No longer a good name.

crypto/src/main/java/net/corda/v5/crypto/KeyUtils.java
```diff
@@ -0,0 +1,48 @@
...
+        return isFulfilledBy(firstKey, Collections.singleton(otherKey));
+    }
+
+    /** Return true if [otherKeys] fulfil the requirements of [firstKey]. */
+    public static boolean isFulfilledBy(PublicKey firstKey, Iterable<PublicKey> otherKeys) {
```
driessamyn: As above. No longer a good name.

driessamyn (changes requested):

crypto/src/main/java/net/corda/v5/crypto/CordaOID.java
dickon: Honestly this was just a trick I didn't know. Done.

crypto/src/main/java/net/corda/v5/crypto/KeyUtils.java
dickon: I renamed the methods in this class.

I wanted to address the high complexity/risk parts first.

crypto/src/main/java/net/corda/v5/crypto/KeyUtils.java
dickon: renamed

crypto/src/main/java/net/corda/v5/crypto/KeyUtils.java
dickon: changed

crypto-extensions/src/main/java/net/corda/v5/crypto/extensions/merkle/MerkleTreeHashDigestProviderWithSizeProofSupport.java
```diff
@@ -0,0 +1,10 @@
+package net.corda.v5.crypto.extensions.merkle;
+
+import net.corda.v5.crypto.merkle.MerkleProof;
+import org.jetbrains.annotations.NotNull;
+import java.util.List;
+
+public interface MerkleTreeHashDigestProviderWithSizeProofSupport extends MerkleTreeHashDigestProvider {
```
driessamyn, line 12: This needs a JDOC.

crypto-java-port.md
```diff
@@ -0,0 +1,62 @@
+# Introduction
```
driessamyn: This should be removed before merge.
Happy for this content to exist on the PR description.

crypto/src/main/java/net/corda/v5/crypto/PublicKeyHash.java
```diff
@@ -0,0 +1,142 @@
...
+// * Computes the public key hash from a given [PublicKey].
+// */
+//fun PublicKey.calculateHash() =
+//    PublicKeyHash.calculate(this)
+//
+///**
+// * Returns the id as the first 12 characters of an SHA-256 hash from a given [PublicKey].
+// */
+//fun PublicKey.publicKeyId(): String =
+//    PublicKeyHash.calculate(this).id
```
driessamyn: Does this need to be removed?

crypto/src/main/java/net/corda/v5/crypto/merkle/MerkleProof.java
```diff
@@ -18,11 +21,21 @@
...
+     * @return Type of the proof
+     */
+    @NotNull
+    MerkleProofType getProofType();
+
+    int getTreeSize();
+    
+     List<IndexedMerkleLeaf> getLeaves();
+    
+     List<SecureHash> getHashes();
```
driessamyn: Looks like a couple of stray spaces? (nitpicky)

driessamyn (commented):
Just a couple of minor doc/formatting comments.

crypto-java-port.md
dickon: Agreed; that was always my intention

crypto-extensions/src/main/java/net/corda/v5/crypto/extensions/DigestAlgorithm.java
```diff
@@ -40,29 +42,36 @@
  */
 interface DigestAlgorithm {
```
kyriathar: Do we want to add `public` here?

crypto-extensions/src/main/java/net/corda/v5/crypto/extensions/DigestAlgorithm.java
```diff
@@ -40,29 +42,36 @@
...
 
     /**
      * Computes the digest of the [InputStream] bytes. The computation must be stateless and thread safe.
      *
      * @param inputStream The [InputStream] to hash.
      * @return the hash value of the [digestLength]
      */
-    fun digest(inputStream: InputStream): ByteArray
+    @NotNull
+    byte[] digest(@NotNull InputStream var1);
```
kyriathar:
```suggestion
    byte[] digest(@NotNull InputStream inputStream);
```

crypto-extensions/src/main/java/net/corda/v5/crypto/extensions/merkle/MerkleTreeHashDigestProvider.java
```diff
@@ -0,0 +1,43 @@
...
+/**
+ * Different use cases require different [MerkleTree] calculations.
+ * [MerkleTreeHashDigestProvider]s let us specify the
+ *  - Leaf Nonce
+ *  - Leaf Hash
+ *  - Node Hash calculation methods
+ *  - Base Digest Algorithm
+ */
+
+interface MerkleTreeHashDigestProvider extends MerkleTreeHashDigest  {
```
kyriathar: Think we need public here.

crypto-extensions/src/main/java/net/corda/v5/crypto/extensions/merkle/MerkleTreeHashDigestProviderWithSizeProofSupport.java
kyriathar: This used to be in same file as `MerkleTreeHashDigestProvider`. Seems to be tightly related to  `MerkleTreeHashDigestProvider` should we keep it in same file as it used to be?

crypto-extensions/src/main/java/net/corda/v5/crypto/extensions/merkle/MerkleTreeHashDigestProviderWithSizeProofSupport.java
```diff
@@ -0,0 +1,10 @@
...
+    @NotNull
+    MerkleProof getSizeProof(@NotNull List<byte []> leaves);
```
kyriathar:
```suggestion
    MerkleProof getSizeProof(@NotNull List<byte[]> leaves);
```

crypto/src/main/java/net/corda/v5/crypto/CompositeKey.java
```diff
@@ -19,30 +22,42 @@
...
  * signatures required) to satisfy the subtree rooted at this node.
  */
-interface CompositeKey: PublicKey {
+public interface CompositeKey extends PublicKey {
+
+
+    /**
+     * Constant specifying the maximum number of children keys in the [CompositeKey].
+     */
+    int COMPOSITE_KEY_CHILDREN_LIMIT = 10;
```
kyriathar: I understand this constant is related to this class but bytecode diversion here I think.

crypto/src/main/java/net/corda/v5/crypto/DigestAlgorithmName.java
```diff
@@ -0,0 +1,83 @@
...
+    @NotNull
+    public final String getName() {
+        return this.name;
+    }
+
+    public DigestAlgorithmName(@NotNull String name) {
+        super();
+        this.name = name;
+        if (this.name.isBlank()) throw new IllegalArgumentException("Hash algorithm name unavailable or not specified");
+    }
```
anien: Just minor: what I came across is that constructor is usually the first function in the class.

crypto/src/main/java/net/corda/v5/crypto/ParameterizedSignatureSpec.java
```diff
@@ -14,12 +15,23 @@
...
+
+    @NotNull
+    public final AlgorithmParameterSpec getParams() {
+        return this.params;
+    }
+
+    public ParameterizedSignatureSpec(@NotNull String signatureName, @NotNull AlgorithmParameterSpec params) {
+        super(signatureName);
+        this.params = params;
+    }
```
anien: Same, constructor is usually the first function.

crypto/src/main/java/net/corda/v5/crypto/PublicKeyHash.java
dickon: deleted

crypto/src/main/java/net/corda/v5/crypto/merkle/MerkleProof.java
dickon: fixed

crypto/src/main/java/net/corda/v5/crypto/CompositeKeyNodeAndWeight.java
```diff
@@ -0,0 +1,29 @@
...
+import java.security.PublicKey;
+
+
+/**
+ * A simple data class for passing keys and weights into CompositeKeyGenerator
+ *
+ * @param node A public key
+ * @param weight The weight for that key, must be greater than zero.
+ */
+public final class CompositeKeyNodeAndWeight {
```
kyriathar, line 11: This used to be a data class. It will need that we override its `equals` `hashCode` and `toString`

crypto/src/main/java/net/corda/v5/crypto/CordaOID.java
```diff
@@ -0,0 +1,37 @@
...
+    private CordaOID() {
+        // this class is never constructed; it exists for the static strings only
+    }
+    
+    /** Assigned to R3, see http://www.oid-info.com/cgi-bin/display?oid=1.3.6.1.4.1.50530&action=display */
+    public static final String OID_R3_ROOT = "1.3.6.1.4.1.50530";
+
+    /** OIDs issued for the Corda platform. */
+    public static final String OID_CORDA_PLATFORM = "1.3.6.1.4.1.50530.1";
```
kyriathar: This used to use `OID_R3_ROOT`. Should we OID_R3_ROOT + ".1" instead?

crypto/src/main/java/net/corda/v5/crypto/CordaOID.java
```diff
@@ -0,0 +1,37 @@
...
+
+    public static final String OID_ALIAS_PRIVATE_KEY = "1.3.6.1.4.1.50530.1.2";
+
+    /** OID for CompositeKey. */
+
+    public static final String OID_COMPOSITE_KEY = "1.3.6.1.4.1.50530.1.3";
+
+    /** OID for CompositeSignature. */
+
+    public static final String OID_COMPOSITE_SIGNATURE = "1.3.6.1.4.1.50530.1.4";
```
kyriathar: Similarly these used to use `OID_CORDA_PLATFORM`.

crypto/src/main/java/net/corda/v5/crypto/DigestAlgorithmName.java
```diff
@@ -0,0 +1,83 @@
...
+import net.corda.v5.base.annotations.CordaSerializable;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+/**
+ * The digest algorithm name. This class is to be used in Corda hashing API.
+ *
+ * @param name The name of the digest algorithm to be used for the instance.
+ */
+
```
kyriathar: line break

crypto/src/main/java/net/corda/v5/crypto/DigestAlgorithmName.java
```diff
@@ -0,0 +1,83 @@
...
+@CordaSerializable
+public final class DigestAlgorithmName {
+    @NotNull private final String name;
+    /**
+     * Instance of SHA-256
+     */
+    @NotNull public static final DigestAlgorithmName SHA2_256 = new DigestAlgorithmName("SHA-256");
```
kyriathar: `NotNull` seems overkill to me here I think. Same for below cases.

crypto/src/main/java/net/corda/v5/crypto/DigestAlgorithmName.java
```diff
@@ -0,0 +1,83 @@
...
+     */
+    @NotNull public String toString() {
+        return this.name;
+    }
+    
+    /**
+     * Returns a hash code value for the object.
+     */
+    public int hashCode() {
+        return this.name.toUpperCase(Locale.ROOT).hashCode();
```
kyriathar: No sure what `Locale.ROOT` is. Is it needed?

crypto/src/main/java/net/corda/v5/crypto/DigestAlgorithmName.java
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+    /**
+     * Returns a Boolean to indicate if the two specified instances of the [DigestAlgorithmName] are the same based on their content.
+     */
+    public boolean equals(@Nullable Object other) {
+        if (other == null) return false;
+        if ((DigestAlgorithmName)this == other) return true;
```
kyriathar: No need to downcast at this point I think, only checking their references.

crypto/src/main/java/net/corda/v5/crypto/DigestAlgorithmName.java
```diff
@@ -0,0 +1,83 @@
...
+        return this.name.toUpperCase().equals(otherDigest.name.toUpperCase());
+    }
+
+    @NotNull
+    public final String getName() {
+        return this.name;
+    }
+
+    public DigestAlgorithmName(@NotNull String name) {
+        super();
```
kyriathar: This could be removed

crypto/src/main/java/net/corda/v5/crypto/DigestAlgorithmName.java
```diff
@@ -0,0 +1,83 @@
...
+        this.name = name;
+        if (this.name.isBlank()) throw new IllegalArgumentException("Hash algorithm name unavailable or not specified");
```
kyriathar: Check could be moved before assigning property and called on parameter.

crypto/src/main/java/net/corda/v5/crypto/DigestAlgorithmName.java
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+    static {
+        DEFAULT_ALGORITHM_NAME = SHA2_256;
+    }
```
kyriathar: Can't we inline this with static var definition further up?

crypto/src/main/java/net/corda/v5/crypto/merkle/MerkleProof.java
```diff
@@ -15,37 +18,39 @@
...
-     *
-     * @returns Root hash of the tree.
-     *
      * @throws MerkleProofRebuildFailureException if the calculation of the root hash failed.
+     * @returns Root hash of the tree.
      */
-    fun calculateRoot(
-        digest: MerkleTreeHashDigest
-    ): SecureHash
+    SecureHash calculateRoot(@NotNull MerkleTreeHashDigest var1);
```
anien: Any reason why did you use `var1` instead of `digest`? Just wondering.

kyriathar (commented):
A first review, till `DigestAlgorithmName.java`

crypto/src/main/java/net/corda/v5/crypto/CompositeKeyNodeAndWeight.java
dickon: done

crypto/src/main/java/net/corda/v5/crypto/CompositeKeyNodeAndWeight.java
```diff
@@ -1,29 +1,60 @@
...
- * @param weight The weight for that key, must be greater than zero.
  */
 public final class CompositeKeyNodeAndWeight {
+    /**
+     * Construct a key
+     *
+     * @param node   A public key
+     * @param weight The weight for that key, must be greater than zero.
+     */
+    public CompositeKeyNodeAndWeight(@NotNull PublicKey node, int weight) {
```
kyriathar, line 21: Let's place this constructor under the properties, looks better then.

crypto/src/main/java/net/corda/v5/crypto/CompositeKeyNodeAndWeight.java
```diff
@@ -1,29 +1,60 @@
...
+        super();
```
kyriathar: Let's remove this `super()` not needed.

crypto/src/main/java/net/corda/v5/crypto/CompositeKeyNodeAndWeight.java
```diff
@@ -1,29 +1,60 @@
...
+        this.node = node;
+        this.weight = weight;
+        if (this.weight <= 0)
```
kyriathar: Let's check parameter instead and move check prior to populating properties

crypto-extensions/src/main/java/net/corda/v5/crypto/extensions/merkle/MerkleTreeHashDigestProviderWithSizeProofSupport.java
dickon: I always thought one interface or class per file was strongly encouraged (or even required) as good style in Java, and that's something Kotlin changes.

crypto/src/main/java/net/corda/v5/crypto/merkle/MerkleProof.java
dickon: Good question! That's what the bytecode decompiler generates, and I went too fast and missed that when cut and pasting. Definetly don't want to call it var1 in the final change. @kyriathar has the baton at the moment.

crypto-extensions/src/main/java/net/corda/v5/crypto/extensions/DigestAlgorithm.java
dickon: fixed

crypto-extensions/src/main/java/net/corda/v5/crypto/extensions/merkle/MerkleTreeHashDigestProvider.java
dickon: yes, done

crypto/src/main/java/net/corda/v5/crypto/DigestAlgorithmName.java
dickon: Don't know, it was in the bytecode that the Kotlin compiler created, so I left it there to preserve behaviour. It's probably harmless to remove it.

crypto/src/main/java/net/corda/v5/crypto/DigestAlgorithmName.java
dickon: yeah, not sure what I was thinking there.

crypto/src/main/java/net/corda/v5/crypto/DigestAlgorithmName.java
dickon: fixed

crypto/src/main/java/net/corda/v5/crypto/ParameterizedSignatureSpec.java
dickon: fixed

crypto/src/main/java/net/corda/v5/crypto/merkle/MerkleProof.java
dickon: fixed

crypto-extensions/src/main/java/net/corda/v5/crypto/extensions/DigestAlgorithm.java
dickon: fixed

crypto-extensions/src/main/java/net/corda/v5/crypto/extensions/merkle/MerkleTreeHashDigestProviderWithSizeProofSupport.java
dickon: taken

crypto/src/main/java/net/corda/v5/crypto/DigestAlgorithmName.java
dickon: yes, although in this case I'll remove it since we don't want to endorse SHA256.

crypto/src/main/java/net/corda/v5/crypto/CompositeKeyNodeAndWeight.java
dickon: done

crypto/src/main/java/net/corda/v5/crypto/CompositeKeyNodeAndWeight.java
dickon: done

crypto/src/main/java/net/corda/v5/crypto/CordaOID.java
dickon: done

crypto/src/main/java/net/corda/v5/crypto/DigestAlgorithmName.java
```diff
@@ -1,53 +1,70 @@
...
+     *  * @param name The name of the digest algorithm to be used for the instance.
+     */
+    public DigestAlgorithmName(@NotNull String name) {
+        super();
+        this.name = name;
+        if (this.name.isBlank()) throw new IllegalArgumentException("Hash algorithm name unavailable or not specified");
+    }
+
+    @NotNull
+    private final String name;
```
kyriathar: Fine for constructor to go up here. Can we please revert `name` property to be on top please.

crypto/src/main/java/net/corda/v5/crypto/DigestAlgorithmName.java
kyriathar: Please let's keep for now to keep diff as smaller as possible in runtime-os PR too. We can decide to remove in some follow-up work.

crypto/src/main/java/net/corda/v5/crypto/DigestAlgorithmName.java
dickon: fixed

crypto/src/main/java/net/corda/v5/crypto/DigitalSignature.java
```diff
@@ -0,0 +1,58 @@
...
+        /**
+         * Public key which corresponding private key was used to sign the data (as if an instance
+         * of the [CompositeKey] is passed to the sign operation it may contain keys which are not actually owned by
+         * the member).
+         */
+        @NotNull
+        private final PublicKey by;
+
+        @NotNull
+        private final Map<String, String> context;
```
kyriathar: Can we please move properties above constructor.

crypto/src/main/java/net/corda/v5/crypto/DigitalSignature.java
```diff
@@ -0,0 +1,58 @@
...
+
+        public final PublicKey getBy() {
+            return this.by;
+        }
```
kyriathar: Not for this PR but this getter name doesn't look very nice, maybe we could consider renaming property and getter

crypto/src/main/java/net/corda/v5/crypto/DigestAlgorithmName.java
dickon: done

crypto/src/main/java/net/corda/v5/crypto/DigestAlgorithmName.java
dickon: done

crypto/src/main/java/net/corda/v5/crypto/CompositeKeyNodeAndWeight.java
dickon: done

crypto/src/main/java/net/corda/v5/crypto/CompositeKeyNodeAndWeight.java
dickon: What happened is that I started out by disassembling the bytecode, and Kotlin often initialises arguments before constructor logic.

crypto/src/main/java/net/corda/v5/crypto/DigestAlgorithmName.java
dickon: done

crypto/src/main/java/net/corda/v5/crypto/DigitalSignature.java
```diff
@@ -0,0 +1,58 @@
...
+         * of the [CompositeKey] is passed to the sign operation it may contain keys which are not actually owned by
+         * the member).
+         */
+        @NotNull
+        private final PublicKey by;
+
+        @NotNull
+        private final Map<String, String> context;
+
+        public final PublicKey getBy() {
```
kyriathar, line 54: I think it needs `NotNull`.

crypto/src/main/java/net/corda/v5/crypto/DigestAlgorithmName.java
dickon: It seems we do not need it, so removed.

crypto/src/main/java/net/corda/v5/crypto/DigestAlgorithmName.java
dickon: It's correct, though, and better to be consistent.

crypto/src/main/java/net/corda/v5/crypto/DigestAlgorithmName.java
dickon: fixed

crypto/src/main/java/net/corda/v5/crypto/CordaOID.java
dickon: fixed

crypto/src/main/java/net/corda/v5/crypto/KeySchemeCodes.java
```diff
@@ -0,0 +1,70 @@
...
+public final class KeySchemeCodes {
+
+    private KeySchemeCodes() {
+        // this class is never constructed; it exists for the static strings.
+    }
+    /**
+     * RSA key scheme code name.
+     * The key scheme can be used for signing only.
+     */
+    @NotNull public static final String RSA_CODE_NAME = "CORDA.RSA";
```
kyriathar: `NotNull`s in this file are redundant I believe.

crypto/src/main/java/net/corda/v5/crypto/MessageAuthenticationCode.java
```diff
@@ -0,0 +1,14 @@
+package net.corda.v5.crypto;
+
+public interface MessageAuthenticationCode {
```
kyriathar: This used to be part of `CryptoUtils.kt`, not really sure if the grouping needs revisiting.  The name could be without "Code" suffix?

crypto/src/main/java/net/corda/v5/crypto/ParameterizedSignatureSpec.java
```diff
@@ -0,0 +1,45 @@
...
+    public String toString() {
+        return this.getSignatureName() + ':' + this.params.getClass().getSimpleName();
+    }
+
+    @NotNull
+    public AlgorithmParameterSpec getParams() {
+        return this.params;
+    }
+
+
```
kyriathar: line breaks

crypto/src/main/java/net/corda/v5/crypto/PublicKeyHash.java
```diff
@@ -0,0 +1,130 @@
+package net.corda.v5.crypto;
```
kyriathar: I guess we are going to be deleting this file/ class when [#901](https://github.com/corda/corda-api/pull/901) gets merged, so not reviewing it

crypto/src/main/java/net/corda/v5/crypto/SecureHash.java
```diff
@@ -0,0 +1,97 @@
...
+        } else {
+            String algorithm = str.substring(0, idx);
+            String value = str.substring(idx + 1);
+            byte[] data = ByteArrays.parseAsHex(value);
+            return new SecureHash(algorithm, data);
+        }
+    }
+
+    @NotNull
+    public final String algorithm;
```
kyriathar: Let's put on top please

crypto/src/main/java/net/corda/v5/crypto/SecureHash.java
```diff
@@ -0,0 +1,97 @@
...
+
+    /**
+     * Creates a [SecureHash].
+     * <p>
+     * This function does not validate the length of the created digest. Prefer using
+     * [net.corda.v5.application.crypto.HashingService.parse] for a safer mechanism for creating [SecureHash]es.
+     *
+     * @see net.corda.v5.application.crypto.HashingService.parse
+     */
+
```
kyriathar: line break

crypto/src/main/java/net/corda/v5/crypto/SecureHash.java
```diff
@@ -0,0 +1,97 @@
...
+    public static final SecureHash parse(String str) {
```
kyriathar: I believe final is redundant here. Static function can't be overridden anyway

crypto/src/main/java/net/corda/v5/crypto/SecureHash.java
```diff
@@ -0,0 +1,97 @@
...
+        }
+    }
+
+    @NotNull
+    public final String algorithm;
+
+    /**
+     * Returns hexadecimal representation of the hash value.
+     */
+    public final String toHexString()  {
```
kyriathar: Let's remove `final` for this and below methods. Reduntant since we have marked class as `final`

crypto/src/main/java/net/corda/v5/crypto/SecureHash.java
```diff
@@ -0,0 +1,97 @@
...
+    public final String prefixChars() {
+        return prefixChars(6);
+    }
+
+    /**
+     * Compares the two given instances of the [SecureHash] based on the content.
+     */
+    public boolean equals(Object other) {
+        if (this == other) return true;
+        return other instanceof SecureHash &&  this.algorithm.equals(((SecureHash)other).algorithm) && super.equals(other);
```
kyriathar:
```suggestion
        return other instanceof SecureHash && this.algorithm.equals(((SecureHash)other).algorithm) && super.equals(other);
```

crypto/src/main/java/net/corda/v5/crypto/SignatureSpec.java
```diff
@@ -0,0 +1,172 @@
...
+     * SHA256withSM2 [SignatureSpec].
+     */
+    public static final SignatureSpec SM2_SHA256 = new SignatureSpec("SHA256withSM2");
+
+    /**
+     * GOST3411withGOST3410 [SignatureSpec].
+     */
+    public static final SignatureSpec GOST3410_GOST3411 = new SignatureSpec("GOST3411withGOST3410");
+
+    public SignatureSpec(@NotNull String signatureName) {
```
kyriathar: We could move on top right after property.

crypto/src/main/java/net/corda/v5/crypto/SignatureSpec.java
```diff
@@ -0,0 +1,172 @@
...
+        super();
+        this.signatureName = signatureName;
+        if (signatureName == null) {
+            throw new IllegalArgumentException("The signatureName must not be null.");
+        }
+        if (signatureName.isBlank()) {
+            throw new IllegalArgumentException("The signatureName must not be blank.");
+        }
```
kyriathar: We could move before properties init, check on params instead

crypto/src/main/java/net/corda/v5/crypto/SignatureSpec.java
```diff
@@ -0,0 +1,172 @@
...
+     */
+    public static final SignatureSpec SM2_SHA256 = new SignatureSpec("SHA256withSM2");
+
+    /**
+     * GOST3411withGOST3410 [SignatureSpec].
+     */
+    public static final SignatureSpec GOST3410_GOST3411 = new SignatureSpec("GOST3411withGOST3410");
+
+    public SignatureSpec(@NotNull String signatureName) {
+        super();
```
kyriathar: super can be removed

crypto/src/main/java/net/corda/v5/crypto/SignatureSpec.java
```diff
@@ -0,0 +1,172 @@
...
+
+    public final @NotNull String getSignatureName() {
+        return this.signatureName;
+    }
+    
+    public int hashCode() {
+        return this.signatureName.hashCode();
+    }
+
+
```
kyriathar: line breaks

crypto-java-port.md
dickon: deleted

crypto/src/main/java/net/corda/v5/crypto/CompositeKey.java
dickon: It was in CryptoUtils.kt before, so yeah, it's different bytecode, but this seemed as good a way to handle this as I could find.

crypto/src/main/java/net/corda/v5/crypto/merkle/HashDigestConstants.java
```diff
@@ -1,52 +1,65 @@
-@file:JvmName("HashDigestConstants")
-package net.corda.v5.crypto.merkle
+package net.corda.v5.crypto.merkle;
+
+import org.jetbrains.annotations.NotNull;
+
+
+public final class HashDigestConstants {
```
kyriathar, line 8: private constructor to disallow instantiations of this class

crypto/src/main/java/net/corda/v5/crypto/merkle/HashDigestConstants.java
```diff
@@ -1,52 +1,65 @@
...
 
 /**
  * An implementation of rfc6962 compatible merkle tree
  */
-const val HASH_DIGEST_PROVIDER_DEFAULT_NAME = "DefaultHashDigestProvider"
+    @NotNull
```
kyriathar, line 13: Again I believe `NotNull`s can be removed

crypto/src/main/java/net/corda/v5/crypto/merkle/IndexedMerkleLeaf.java
```diff
@@ -0,0 +1,64 @@
...
+/**
+ * [IndexedMerkleLeaf]s are building blocks of [MerkleProof]s.
+ * They contain the required information about a particular leaf which is needed for the verification.
+ *
+ * @property index The leaf's index.
+ * @property nonce The leaf's optional nonce.
+ * @property leafData The leaf's data.
+ *
+ */
+
```
kyriathar: line break

crypto/src/main/java/net/corda/v5/crypto/merkle/IndexedMerkleLeaf.java
```diff
@@ -0,0 +1,64 @@
...
+@CordaSerializable
+public final class IndexedMerkleLeaf {
+    private final int index;
+    @Nullable
+    private final byte[] nonce;
+    @NotNull
+    private final byte[] leafData;
+
```
kyriathar, line 20: line break

crypto/src/main/java/net/corda/v5/crypto/merkle/IndexedMerkleLeaf.java
```diff
@@ -0,0 +1,64 @@
...
+
+     public IndexedMerkleLeaf(int index, @Nullable byte[] nonce, @NotNull byte[] leafData) {
+        super();
```
kyriathar: super can go

crypto/src/main/java/net/corda/v5/crypto/merkle/IndexedMerkleLeaf.java
```diff
@@ -0,0 +1,59 @@
...
+        if (other == null) return false;
+        if (!(other instanceof IndexedMerkleLeaf)) return false;
+
+        IndexedMerkleLeaf otherLeaf = (IndexedMerkleLeaf) other;
+
+        if (index != otherLeaf.index) return false;
+        if (nonce != otherLeaf.nonce) {
+            if (otherLeaf.nonce == null) return false;
+            if (!Arrays.equals(nonce, otherLeaf.nonce)) return false;
+        } else if (otherLeaf.nonce != null) return false;
```
kyriathar, line 43: This was already there (so not for this PR), but doesn't seem very right to me I think. I think it means pointers to different bytearrays which could still be equal will fail?

kyriathar (commented):
A second review till `IndexedMerkleLeaf`

crypto/src/main/java/net/corda/v5/crypto/MessageAuthenticationCode.java
dickon: Message Authentication Code (MAC) is a term of art, we should keep it in full.

kyriathar: Should it be made a class with private constructor to host the below consts?

crypto/src/main/java/net/corda/v5/crypto/merkle/MerkleProofRebuildFailureException.java
```diff
@@ -0,0 +1,13 @@
...
+
+import net.corda.v5.base.exceptions.CordaRuntimeException;
+import org.jetbrains.annotations.NotNull;
+
+/**
+ * Indicates that the calculation of the root hash of a [MerkleProof] failed.
+ */
+public final class MerkleProofRebuildFailureException extends CordaRuntimeException {
+    public MerkleProofRebuildFailureException(@NotNull String message) {
+        super(message, (Throwable) null);
```
kyriathar: I think let's remove cast on null

crypto/src/main/java/net/corda/v5/crypto/merkle/MerkleTreeHashDigest.java
```diff
@@ -0,0 +1,20 @@
...
+
+import net.corda.v5.crypto.DigestAlgorithmName;
+import org.jetbrains.annotations.NotNull;
+
+/**
+ * Different use cases require different [MerkleTree] calculations.
+ * [MerkleTreeHashDigest]s let us specify such implementations.
+ *
+ */
+
```
kyriathar: line break

crypto/src/main/java/net/corda/v5/crypto/SecureHash.java
```diff
@@ -0,0 +1,99 @@
...
+        } else {
+            String algorithm = str.substring(0, idx);
+            String value = str.substring(idx + 1);
+            byte[] data = ByteArrays.parseAsHex(value);
+            return new SecureHash(algorithm, data);
+        }
+    }
+
+    @NotNull
+    public final String algorithm;
```
kyriathar: I think this needs to become private and add a getter for it

crypto/src/test/java/net/corda/v5/crypto/SecureHashJavaApiTest.java
```diff
@@ -14,7 +14,7 @@ public void shouldCreateInstance() {
             36, 45, 42, 121, 39, 35, -29, -3, 111, -48, 97, -7, -43, -49, -47, 59, -113, -106, 19, 88, -26, -83, -70, 74
         };
         var cut = SecureHash.parse(str);
-        assertEquals(DigestAlgorithmName.SHA2_384.getName(), cut.getAlgorithm());
+        assertEquals(DigestAlgorithmName.SHA2_384.getName(), cut.algorithm);
```
kyriathar: This shouldn't change if we make property private and add getter

kyriathar (commented):
Went through all classes I think other than `CryptoUtils.kt`.

crypto-extensions/src/main/java/net/corda/v5/crypto/extensions/DigestAlgorithm.java
```diff
@@ -0,0 +1,77 @@
...
+/**
+ * Digest extensions: Interface defining a custom digest calculation implementation. The interface should be implemented
+ * if a CPK developer wishes to provide support for digest algorithms beyond those supported by the Corda platform.
+ * The implementation of the interface must be coupled with the implementation of the [DigestAlgorithmFactory] which
+ * will be used to create the instances. For each algorithm there must be matching a pair of
+ * [DigestAlgorithmFactory] and [DigestAlgorithm] implementations.
+ * <p>
+ * Example of the algorithm and the factory implementations:
+ * <p>
+ * ```kotlin
```
driessamyn: Not a change for this PR, but maybe we should raise a JIRA for adding a JAVA example here?

crypto/src/main/java/net/corda/v5/crypto/KeySchemeCodes.java
dickon: The Kotlin version has them in the bytecode.

crypto/src/main/java/net/corda/v5/crypto/DigitalSignature.java
dickon: added

crypto/src/main/java/net/corda/v5/crypto/CompositeKey.java
driessamyn: Does it need to be in API?
I.e. will we ever want to change it without breaking the API?
I admit I don't know much about how this is used, but this feels more like an implementation constraint.

Maybe @mnesbit can opine?

crypto/src/main/java/net/corda/v5/crypto/DigitalSignature.java
dickon: yeah, but resolving for this PR.

crypto/src/main/java/net/corda/v5/crypto/DigitalSignature.java
dickon: done

crypto/src/main/java/net/corda/v5/crypto/merkle/IndexedMerkleLeaf.java
dickon: done

crypto/src/main/java/net/corda/v5/crypto/merkle/IndexedMerkleLeaf.java
dickon: done

crypto/src/main/java/net/corda/v5/crypto/KeyUtils.java
```diff
@@ -0,0 +1,63 @@
...
+public class KeyUtils {
+
+    private KeyUtils() {
+        // this class is never constructed; it exists for the static methods and data only
+    }
+
+    /**
+     * Constant specifying the maximum number of key lookup by id items.
+     */
+    public static int KEY_LOOKUP_INPUT_ITEMS_LIMIT = 20;
```
driessamyn: Similar to the comment above. 
Where/how is this used and is it really API?

crypto/src/main/java/net/corda/v5/crypto/merkle/IndexedMerkleLeaf.java
dickon: I don't see how.  Do you mean the nonce or leafData bytearrays?

crypto/src/main/java/net/corda/v5/crypto/merkle/IndexedMerkleLeaf.java
dickon: fixed

crypto/src/main/java/net/corda/v5/crypto/merkle/HashDigestConstants.java
dickon: They are in the Kotlin bytecode and by default I'm trying to stay close to that. And they are good documentation.

crypto-extensions/src/main/java/net/corda/v5/crypto/extensions/DigestAlgorithm.java
dickon: Feel free if you like, I'm not concerned enough to raise a ticket.

crypto/src/test/java/net/corda/v5/crypto/SecureHashJavaApiTest.java
dickon: yes, reverted.

crypto/src/main/java/net/corda/v5/crypto/SecureHash.java
dickon: done

crypto/src/main/java/net/corda/v5/crypto/merkle/MerkleTreeHashDigest.java
dickon: done

crypto/src/main/java/net/corda/v5/crypto/merkle/MerkleProofRebuildFailureException.java
dickon: done

crypto/src/main/java/net/corda/v5/crypto/MessageAuthenticationCode.java
dickon: good spot, yeah. Done.

crypto/src/main/java/net/corda/v5/crypto/KeyUtils.java
dickon: It's used for instance in the `net.corda.data.crypto.wire.ops.flow.queries.ByIdsFlowQuery` Avro messsage. The number 20 is in the Avro docstring, and it is used in both the implementations and test in `corda-runtime-os`. However, it is reasonable to have access to a numeric constant value here, rather than make me put the constant in to their code.

kyriathar: I think this should be internal. Nothing for the user to know how we batch the keys onto Kafka.

dickon: The Avro messages are not internal; the number 20 is already in `corda-api`. So I regard it as public information. The client-side (with respect to Kafka) stuff we offer is, as far as  I know, best regarded as sample code.

dickon: (The constant was originally defined in CryptoUitls)

crypto/src/main/java/net/corda/v5/crypto/SignatureSpec.java
dickon: fixed

crypto/src/main/java/net/corda/v5/crypto/merkle/HashDigestConstants.java
dickon: done

crypto/src/main/java/net/corda/v5/crypto/KeyUtils.java
driessamyn: I agree with @kyriathar on this. I don't think this belongs in the API bytecode.

AVRO changes we can change over time, but API bytecode we can't. 
If we choose to change this to 25 as part of 5.1, why would that impact somebody's CorDapp ?

crypto/src/main/java/net/corda/v5/crypto/merkle/IndexedMerkleLeaf.java
driessamyn: If there is some debate about this and it's an existing issue, I'm ok with that being treated as a separate JIRA/PR after beta2. The implementation doesn't really impact the CorDapp API.

crypto-extensions/src/main/java/net/corda/v5/crypto/extensions/DigestAlgorithm.java
chrisr3: Javadoc does not support markdown. You will need to wrap this code block inside:
```
<pre>{@code .... }</pre>
```
Don't forget to reapply formatting :wink:.

These comments apply throughout.

crypto-extensions/src/main/java/net/corda/v5/crypto/extensions/DigestAlgorithm.java
```diff
@@ -0,0 +1,77 @@
...
+import org.jetbrains.annotations.NotNull;
+
+import java.io.InputStream;
+
+/**
+ * Digest extensions: Interface defining a custom digest calculation implementation. The interface should be implemented
+ * if a CPK developer wishes to provide support for digest algorithms beyond those supported by the Corda platform.
+ * The implementation of the interface must be coupled with the implementation of the [DigestAlgorithmFactory] which
+ * will be used to create the instances. For each algorithm there must be matching a pair of
+ * [DigestAlgorithmFactory] and [DigestAlgorithm] implementations.
```
chrisr3: This are KDoc references, which need to be replaced with `{@link ...}` in Javadoc (e.g.)
```
{@link DigestAlgorithmFactory}
```

crypto-extensions/src/main/java/net/corda/v5/crypto/extensions/merkle/MerkleTreeHashDigestProvider.java
```diff
@@ -0,0 +1,45 @@
...
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+/**
+ * Different use cases require different [MerkleTree] calculations.
+ * [MerkleTreeHashDigestProvider]s let us specify the
+ * - Leaf Nonce
+ * - Leaf Hash
+ * - Node Hash calculation methods
+ * - Base Digest Algorithm
```
chrisr3: This will need to become:
```
<ul>
<li>Leaf Nonce</li>
...
</ul>
```

crypto-extensions/src/main/java/net/corda/v5/crypto/extensions/merkle/MerkleTreeHashDigestProvider.java
```diff
@@ -0,0 +1,45 @@
...
+    byte[] leafNonce(int index);
+
+    /**
+     * Calculates the hash of a leaf.
+     *
+     * @param index The leaf's index.
+     * @param nonce The leaf's nonce.
+     * @param bytes The leaf's content bytes.
+     */
+    SecureHash leafHash(int index, @Nullable byte[] nonce, @NotNull byte[] bytes);
```
chrisr3, line 39: Missing `@NotNull` or (`@Nullable`?) on return types.

crypto/src/main/java/net/corda/v5/crypto/CompositeKey.java
dickon: Let's move to `corda-runtime-os`. The implementation of `CompositeKey` is in corda-runtime-os. The CorDapp might call the factory function `CompositeKeyGeneration,.create` with too many keys, in which the factory function will  throw `IllegalArgumentFunction`. We can put the limit in some documentation if we want, and increase it in further version.

(Personally I'd have done this in a separate PR before GA but since others seem keen to do some API cleanup now then so be it).

crypto/src/main/java/net/corda/v5/crypto/KeyUtils.java
dickon: Again, seems a shame to delay this PR, but we can do this now so let's do that.

crypto-extensions/src/main/java/net/corda/v5/crypto/extensions/merkle/MerkleTreeHashDigestProvider.java
```diff
@@ -0,0 +1,45 @@
...
+
+    /**
+     * Calculates the hash of a node.
+     *
+     * @param depth Depth of the node.
+     * @param left  [SecureHash] of the left child of the node.
+     * @param right [SecureHash] of the right child of the node.
```
chrisr3: `[SecureHash` -> `{@link SecureHash}` etc.

crypto/src/main/java/net/corda/v5/crypto/CompositeKey.java
```diff
@@ -1,48 +1,60 @@
...
 
 /**
  * A tree data structure that enables the representation of composite public keys, which are used to represent
  * the signing requirements for multi-signature scenarios. A composite key is a list
  * of leaf keys and their contributing weight, and each leaf can be a conventional single key or a composite key.
  * Keys contribute their weight to the total if they are matched by the signature.
- *
+ * <p>
  * For complex scenarios, such as *"Both Alice and Bob need to sign to consume a state S"*, we can represent
  * the requirement by creating a tree with a root [CompositeKey], and Alice and Bob as children.
```
chrisr3: `*text*` -> `<b>text</b>`

crypto/src/main/java/net/corda/v5/crypto/CompositeKey.java
```diff
@@ -1,48 +1,60 @@
...
+    void checkValidity();
 
     /**
      * Takes single [PublicKey] and checks if [CompositeKey] requirements hold for that key.
+     *
+     * @param key the public key
+     * @return true if the public key is a composite key, false otherwise
      */
-    fun isFulfilledBy(key: PublicKey): Boolean
+    boolean isFulfilledBy(PublicKey key);
```
chrisr3: `@NotNull PublicKey key`

crypto/src/main/java/net/corda/v5/crypto/CompositeKey.java
```diff
@@ -1,48 +1,60 @@
...
 
     /**
      * Checks if the public keys corresponding to the signatures are matched against the leaves of the composite
      * key tree in question, and the total combined weight of all children is calculated for every intermediary node.
      * If all thresholds are satisfied, the composite key requirement is considered to be met.
      */
-    fun isFulfilledBy(keysToCheck: Iterable<PublicKey>): Boolean
+    boolean isFulfilledBy(Iterable<PublicKey> keysToCheck);
```
chrisr3: `@NotNull Iterable<PublicKey>`

crypto/src/main/java/net/corda/v5/crypto/CompositeKey.java
driessamyn: Thanks.

The main reason I think this is important is that this leaves the option of somebody referencing this value in their CorDapp, so taking it away later means that change will break the CorDapp.

So it's in a slightly different category compared to things we can fix up without impacting the CorDapp API.

crypto/src/main/java/net/corda/v5/crypto/DigestAlgorithmName.java
```diff
@@ -0,0 +1,79 @@
...
+
+    /**
+     * Construct a digest algorithm name.
+     * <p>
+     * * @param name The name of the digest algorithm to be used for the instance.
+     */
+    public DigestAlgorithmName(@NotNull String name) {
+        if (name.isBlank()) throw new IllegalArgumentException("Hash algorithm name unavailable or not specified");
```
chrisr3: You should probably check for name being `null` too, since that what Kotlin would have done.

crypto/src/main/java/net/corda/v5/crypto/DigitalSignature.java
```diff
@@ -0,0 +1,60 @@
...
+            this.by = by;
+            this.context = context;
+        }
+
+        /**
+         * Public key which corresponding private key was used to sign the data (as if an instance
+         * of the [CompositeKey] is passed to the sign operation it may contain keys which are not actually owned by
+         * the member).
+         */
+        @NotNull
```
chrisr3: Not sure that `@NotNull` makes much sense for `private` fields :thinking:.

crypto/src/main/java/net/corda/v5/crypto/KeyUtils.java
```diff
@@ -0,0 +1,60 @@
+package net.corda.v5.crypto;
+
+import java.security.PublicKey;
+import java.util.Collections;
+import java.util.Set;
+
+public class KeyUtils {
```
chrisr3: `public final class KeyUtils`?

crypto/src/main/java/net/corda/v5/crypto/KeyUtils.java
```diff
@@ -0,0 +1,60 @@
...
+    public static int KEY_LOOKUP_INPUT_ITEMS_LIMIT = 20;
+
+
+    /**
+     * Checks whether any of the given [keys] match a leaf on the [CompositeKey] tree or a single [PublicKey].
+     * <p>
+     * <i>Note that this function checks against leaves, which cannot be of type [CompositeKey]. Due to that, if any of the
+     * [otherKeys] is a [CompositeKey], this function will not find a match.</i>
+     */
+    public static boolean isKeyInSet(PublicKey key, Iterable<PublicKey> otherKeys) {
```
chrisr3: `@NotNull` on parameters.

crypto/src/main/java/net/corda/v5/crypto/SecureHash.java
```diff
@@ -0,0 +1,103 @@
...
+    static final char DELIMITER = ':';
+
+    /**
+     * Creates a [SecureHash].
+     * <p>
+     * This function does not validate the length of the created digest. Prefer using
+     * [net.corda.v5.application.crypto.HashingService.parse] for a safer mechanism for creating [SecureHash]es.
+     *
+     */
+    public static SecureHash parse(String str) {
```
chrisr3: Missing `@NotNull` annotations.

chrisr3 (dismissed):
You still need to finish your KDoc -> Javadoc conversions. And there are many missing `@NotNull` / `@Nullable` annotations still.

crypto/src/main/java/net/corda/v5/crypto/CompositeKey.java
dickon: Removed in my local work.

crypto/src/main/java/net/corda/v5/crypto/merkle/IndexedMerkleLeaf.java
dickon: okay

crypto/src/main/java/net/corda/v5/crypto/KeyUtils.java
dickon: removed locally

crypto-extensions/src/main/java/net/corda/v5/crypto/extensions/DigestAlgorithm.java
dickon: fixed in local copy

crypto-extensions/src/main/java/net/corda/v5/crypto/extensions/DigestAlgorithm.java
dickon: fixed in local copy

crypto-extensions/src/main/java/net/corda/v5/crypto/extensions/merkle/MerkleTreeHashDigestProvider.java
dickon: yes, done locally

crypto/src/main/java/net/corda/v5/crypto/DigitalSignature.java
dickon: why not though? Still useful for compilers and linters

crypto/src/main/java/net/corda/v5/crypto/SecureHash.java
dickon: fixed

crypto/src/main/java/net/corda/v5/crypto/KeyUtils.java
dickon: fixed

crypto/src/main/java/net/corda/v5/crypto/KeyUtils.java
dickon: fixed

crypto/src/main/java/net/corda/v5/crypto/DigestAlgorithmName.java
dickon: done

crypto-extensions/src/main/java/net/corda/v5/crypto/extensions/merkle/MerkleTreeHashDigestProvider.java
dickon: fixed

crypto-extensions/src/main/java/net/corda/v5/crypto/extensions/merkle/MerkleTreeHashDigestProvider.java
dickon: fixed

crypto-extensions/src/main/java/net/corda/v5/crypto/extensions/merkle/MerkleTreeHashDigestProvider.java
dickon: fixed

crypto-extensions/src/main/java/net/corda/v5/crypto/extensions/DigestAlgorithm.java
dickon: fixed

crypto-extensions/src/main/java/net/corda/v5/crypto/extensions/DigestAlgorithm.java
dickon: pushed

crypto/src/main/java/net/corda/v5/crypto/KeyUtils.java
dickon: pushed

crypto/src/main/java/net/corda/v5/crypto/CompositeKey.java
dickon: fixed

crypto/src/main/java/net/corda/v5/crypto/CompositeKey.java
dickon: fixed

crypto/src/main/java/net/corda/v5/crypto/CompositeKey.java
dickon: I used <em> in this case. Done.

crypto/src/main/java/net/corda/v5/crypto/CompositeKey.java
dickon: pushed

crypto/src/main/java/net/corda/v5/crypto/ParameterizedSignatureSpec.java
dickon: they are gone

crypto/src/main/java/net/corda/v5/crypto/PublicKeyHash.java
dickon: yeah, it's deleted now

crypto/src/main/java/net/corda/v5/crypto/SecureHash.java
dickon: done

crypto/src/main/java/net/corda/v5/crypto/SecureHash.java
dickon: no longer present

crypto/src/main/java/net/corda/v5/crypto/SecureHash.java
dickon: gone

crypto/src/main/java/net/corda/v5/crypto/SecureHash.java
dickon: gone

crypto/src/main/java/net/corda/v5/crypto/SecureHash.java
dickon: this was changed, no longer relevant

crypto/src/main/java/net/corda/v5/crypto/SignatureSpec.java
dickon: done

crypto/src/main/java/net/corda/v5/crypto/SignatureSpec.java
dickon: done

crypto/src/main/java/net/corda/v5/crypto/SignatureSpec.java
dickon: gone

crypto/src/main/java/net/corda/v5/crypto/merkle/IndexedMerkleLeaf.java
driessamyn: @kyriathar can you create a JIRA?

kyriathar: Raised [CORE-11305](https://r3-cev.atlassian.net/browse/CORE-11305) for it.

crypto/src/main/java/net/corda/v5/crypto/DigitalSignature.java
dickon: comiitted locally

driessamyn (approved):
LGTM

kyriathar (approved):

PR #888, opened 2023-02-23
CORE-10793: Refactor isRelevant

MrMatthewLayton:
It's been suggested that the `Contract.isRelevant` function is neither suitably named, nor suitably located.

The original function is defined as such:
```java
public interface Contract {
    default boolean isRelevant(ContractState state, Set<PublicKey> myKeys) {
        ...
    }
}
```

Problems:
1. What does _relevant_ mean in this context?
2. Given a transaction which may include _n > 1_ `ContractState` outputs, the function would need an if/else condition to check each one individually, which is not straightforward for developers.
3. It's not the _relevancy_ of the **contract** that is being determined, rather, the relevancy of each state within a transaction, therefore this implementation has broken the single responsibility principle, as it's not the contract's responsibility to decide who `state` is, or isn't relevant to.

I've refactored `isRelevant` to:

```java
public interface ContractState {

    ...

    default boolean isVisible(@NotNull UtxoLedgerTransaction transaction, @NotNull Iterable<PublicKey> myKeys) {
        ...
    }
}
```

The rationale behind this refactor:
1. Renamed to `isVisible` because _visibility_ better describes the intent of the function than _relevancy_ does, therefore from a naming perspective it is less ambiguous.
2. Rather than an overarching contract having to decide visibility for each state, instead, the state gets to decide only its visibility in relation to the transaction in which it is being issued, and the keys that are available to it. This should be far more logical to developers to implement.
3. As each `ContractState` is now responsible for its own visibility, this respects the Single Responsibility Principle, by allowing each state implementation to implement something that fundamentally belongs to its behaviour.

---
**UPDATE: 2023-02-27**
As per last week's meeting on this topic, we have instead decided to implement a _checker_ service, which will be implemented against `Contract` rather than `ContractState`, as this provides us with better flexibility for visibility checking in future.

```kotlin
public interface VisibilityChecker {
    boolean areAnyKeysOurs(@NotNull Iterable<PublicKey> keys);
}

public interface Contract {
    default boolean isVisible(@NotNull ContractState state, @NotNull VisibilityChecker checker) {
        return checker.areAnyKeysOurs(state.getParticipants());
    }
}
```

Should we desire visibility checking per `ContractState`, this is still achievable with this design:
```kotlin
public interface VisibleContractState extends ContractState {

    default boolean isVisible(@NotNull VisibilityChecker checker) {
        return checker.areAnyKeysOurs(getParticipants());
    }
}

public interface Contract {
    default boolean isVisible(@NotNull ContractState state, @NotNull VisibilityChecker checker) {
        if (state is VisibleContractState) {
            return state.isVisible(checker);
        }
        
        return checker.areAnyKeysOurs(state.getParticipants());
    }
}
```

The proposal above demonstrates that visibility checking can be delegated to a contract state, meaning that visibility concerns can be implemented on a per-state basis, as well as at the contract level.


MrMatthewLayton:
From Slack...

It feels wrong being on the contract. 

If we're trying to determine whether a **state** is visible to a particular node, then that's not the **contract's** responsibility, and it's really hard to reason about when you implement it. Given the current implementation, all you can really do is determine whether a state is visible based on its participants, or key properties if you cast the state...

```kotlin
override isVisible(ContractState state, myKeys: Set<PublicKey>): Boolean {
    // all that is accessible at this point is participants, so...
    return myKeys.any { it in state.participants }

    // Or you could cast it...
    if(state is Token) {
        return myKeys.any { it in state.participants } || state.issuer in myKeys
    }
}
```

That's really not very useful as the function doesn't even have access to the `UtxoLedgerTransaction` so you can't even determine visibility based on other stuff in the transaction. 

So, let's say we add that. Imagine a fake scenario where we have visibility rules, a token definition and a token all in the same transaction...

```kotlin
override isVisible(UtxoLedgerTransaction transaction, ContractState state, myKeys: Set<PublicKey>): Boolean {

    if(state is VisibilityRules)
    {
        return false
    }

    val visibilityRules = transaction.getOutputsOfType<VisibilityRules>().single()

    if(state is TokenDefinition)
    {
        return state.issuer in myKeys || visibilityRules.tokenDefinitionIsVisibleToEveryone
    }

    if(state is Token)
    {
        return state.bearer in myKeys || (state.issuer in myKeys && visibilityRules.tokenIsVisibleToIssuer)
    }
}
```

Now the function can get extra data from the transaction, but has to account for different rules based on different contract state types.

This just doesn't seem like a logical separation of visibility logic.

In contrast, if we put that function on the contract state (since it relates to the contract state anyway)...

```kotlin
class TokenDefinition(val issuer: PublicKey) : ContractState {
    
    override fun getParticipants(): List<PublicKey> {
        return listOf(issuer)
    }

    override fun isVisible(transaction: UtxoLedgerTransaction, service: SigningService): Boolean {
        val myKeys = service.findMyLedgerKeys()
        val visibilityRules = transaction.getOutputsOfType<VisibilityRules>().single()
        return issuer in myKeys || visibilityRules.tokenDefinitionIsVisibleToEveryone
    }
}

class Token(val issuer: PublicKey, val owner: PublicKey) : ContractState {

    override fun getParticipants(): List<PublicKey> {
        return listOf(owner)
    }

    override fun isVisible(transaction: UtxoLedgerTransaction, service: SigningService): Boolean {
        val myKeys = service.findMyLedgerKeys()
        val visibilityRules = transaction.getOutputsOfType<VisibilityRules>().single()
        return bearer in myKeys || (issuer in myKeys && visibilityRules.tokenIsVisibleToIssuer)
    }
}
```

ledger/ledger-utxo/src/main/java/net/corda/v5/ledger/utxo/VisibilityChecker.java
```diff
@@ -0,0 +1,19 @@
+package net.corda.v5.ledger.utxo;
+
+import org.jetbrains.annotations.NotNull;
+
+import java.security.PublicKey;
+
+/**
+ * Provides functionality to determine state visibility.
+ */
+public interface VisibilityChecker {
```
mnesbit, line 10: I think this might be clearer in same file as consumer API function

ledger/ledger-utxo/src/main/java/net/corda/v5/ledger/utxo/VisibilityChecker.java
```diff
@@ -0,0 +1,19 @@
...
+
+    /**
+     * Determines whether any of the specified keys belong to the current node.
+     *
+     * @param keys The keys for which to determine belong to the current node.
+     * @return Returns true if any of the specified keys belong to the current node; otherwise, false.
+     */
+    boolean areAnyKeysOurs(@NotNull Iterable<PublicKey> keys);
```
mnesbit: Need a better name really. My thoughts are: containsSigningKey

PR #890, opened 2023-02-24
CORE-9702 - Query for queued registration requests

nikinagy:
Modified command schema for querying registration requests to be able to limit the number of results.


data/avro-schema/src/main/resources/avro/net/corda/data/membership/db/request/query/QueryRegistrationRequests.avsc
```diff
@@ -16,6 +16,11 @@
         "type": "array",
         "items": "net.corda.data.membership.common.RegistrationStatus"
       }
+    },
+    {
+      "name": "limit",
```
yift-r3, line 21: If we have a limit, we should have order and paging (probably not for this ticket)...

nikinagy: I had a discussion with @dimosr before adding the limit and I think the conclusion was that we will add paging etc in the future. But we should keep this in mind, create a ticket for it.

dimosr: Yes, pagination has already been scoped as part of the large network support epic.

charlieR3 (dismissed):

charlieR3 (approved):

PR #893, opened 2023-02-24
INFRA-2046: seperate kdoc and java doc publsihing 

ronanbrowne:
Seperate javaDoc and Kdoc publishing as requested by docs team.

Builds now have two distinct targets (and artifacts published) for Kdoc / JavaDoc.

`generateMultiModuleKDocs` will generate a zip archive of all kdoc in the project

`generateMultiModuleJavaDocs` will generate a zip archive of all javaDocs in the project.

Refactored `corda.java-only.gradle` to a slightly different approach to remove the need to apply `org.jetbrains.kotlin.jvm` plugin to the java modules, as this causes the java modules to also be included in the Kdoc.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -334,17 +334,27 @@ wrapper {
     distributionType = Wrapper.DistributionType.BIN
 }
 
-tasks.register('generateMultiModuleDocs', Zip) {
-    description = 'Create mutli module docs Site Jar'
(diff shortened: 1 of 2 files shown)


ronanbrowne:
Actually, back to draft ... not yet ready here

ronanbrowne:
Closing for now will revisit post Kotlin migration if needed

PR #898, opened 2023-03-01
CORE-10336 Internalize non api types

kyriathar:


dickon (approved):
LGTM

Let's make these internal, until we are are we need them public. I want to revisit the throttling strategy, with custom logic where we need it for retries rather than a layer at the top of the crypto processor.

mnesbit (approved):
OK

PR #904, opened 2023-03-02
CORE-10678 Make MGM Registration Idempotent

williamvigorr3:
Change vnode_group_policy table to not have a generated version. So operations on the vnode_group_policy table can be made idempotent.


dimosr (dismissed):

ronanbrowne (approved):
lgtm (blt - review)

dimosr (approved):

PR #905, opened 2023-03-02
CORE-5533 - minimum config value setting not being adhered to when config change is input

emilybowe:
**Description:**
The issue here: [CORE-5533](https://r3-cev.atlassian.net/browse/CORE-5533) seemed like the minimum setting of 1000 milliseconds for _maxRetryDelay_ was not being used as 999 milliseconds was accepted. 

However, looking at the query that was tested actually the issue appeared to be that the _maxRetryDelay_ was not queried within the _processing_ object it was defined in. As a result of this the _maxRetryDelay_ was actually being set and accepted as an _additional property_. 

The fix here was to prevent additional properties from being accepted. This was done by adding additionalProperties: false to: properties object, event object, session object, processing object.

The issue raised was replicated and then the solution tested using manual testing. Curl command and response below. 

**Testing:**

```
curl -X 'PUT' \
  -u "admin:admin" \
  -k 'https://localhost:8888/api/v1/config' \
  -H 'accept: application/json' \
  -H 'Content-Type: application/json' \
  -d '{
  "request": {
    "config": "{ \"maxRetryDelay\": 999 }",
    "schemaVersion": {
      "major": 1,
      "minor": 0
    },
    "section": "corda.flow",
    "version": 0
    }
  }'
```

**Response:**

```
{
    "title": "Configuration \"{ \"maxRetryDelay\": 999 }\" could not be validated. Valid JSON or HOCON expected. Cause: Configuration failed to validate for key corda.flow at schema version null: $.maxRetryDelay: is not defined in the schema and the schema does not allow additional properties",
    "status": 400,
    "details": {}
}%    
```                   

[CORE-5533]: https://r3-cev.atlassian.net/browse/CORE-5533?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ben-millar (dismissed):

ben-millar:
LGTM, pending 🟢 build

data/config-schema/src/main/resources/net/corda/schema/configuration/flow/1.0/corda.flow.json
```diff
@@ -95,7 +97,10 @@
...
         }
-      }
-    }
-  }
+      },
+      "additionalProperties": false
+    },
+    "additionalProperties": false
+  },
+  "additionalProperties": false
```
LWogan: ive a feeling this last one on the root json node _might_ not be doing anything as its not defined within a properties block.

emilybowe:
> 
So I've looked into the failed gate here with Ronan. `CorDapp-downstream-compatibility-checks ` is a new non blocking PR gate added yesterday. Ronan is going to send comms on the new gate soon but in the meantime on this PR it seems here for some reason, a downstream job has not passed all the needed arguments to gradle when running. Ronan said either way it's build/infra related so fine to merge.

LWogan (approved):

ben-millar (approved):

PR #909, opened 2023-03-04
CORE-10416 : Maximize normalization in metaInf dir of Jars, minimize cache misses. 

ronanbrowne:
When calculating gradle cache, ignore changes in the metaInf dir of jars, any properties here that change from build to build will result in cache misses and sub-optimal performance of remote/local caching.

This only affects Gradles caching mechanism and how cache keys are inspected and has no bearing on published artifacts / what is produced in a jars metaInf.

When observing build-cache misses Dokka task was also adding overhead for each run due to the changing module version, set this to cordaVersion or RELEASE_VERSION if it exists, which does not add the dynamically injected suffix, this ensures these tasks are up-to-date if no dokka related change has occurred.

Similar runtime os where i have captured some build scan data in the description: https://github.com/corda/corda-runtime-os/pull/3304

### DIFF
--- a/.ci/JenkinsApiCompatibility
+++ b/.ci/JenkinsApiCompatibility
@@ -1,4 +1,4 @@
 // Check corda-api compatibility with downstream consumers which implement CordApps
-@Library('corda-shared-build-pipeline-steps@5.0.1') _
+@Library('corda-shared-build-pipeline-steps@ronanb/CORE-11304/c5-compatibility-issues') _
 
 cordaApiCompatibilityCheck()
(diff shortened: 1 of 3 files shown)


yarivus (approved):
LGTM
@ronanbrowne can you add the value-gain observed by this change ?

ronanbrowne:
> LGTM @ronanbrowne can you add the value-gain observed by this change ?

PR slipped through my radar, But the difference is  about 17 Gradle tasks approx 1 min saved in corda-api (varies per execution based on other factors also ) 

but some comparisons of buid scans here

post change https://gradle.dev.r3.com/c/tc3uv2gfjvipw/wuc6qsxixa2sw/task-inputs?cacheability=cacheable

pre-change: https://gradle.dev.r3.com/c/qjl53jx6ztmqa/i4gnsc5dd4bkc/task-inputs?cacheability=cacheable 

The main take away is is the same control scenario, we were able to get a build to take advantage of the case so that 44 tasks were executed (this is quite low)

PR #910, opened 2023-03-05
CORE-11304: follow up from branch creation, ensure we set correct downstream branch builds for api -> e2e compatibility 

ronanbrowne:
when building against e2e tests ensure we checkout `release/5.0-Beta2` in line with this beta branch


yarivus (approved):
LGTM

PR #911, opened 2023-03-06
CORE-9724: `SecureHash` constructors should be removed from public API

anien:
Change `SecureHash` to interface.
Move parse function to `DigestService`.
`StateRef's parse` function now also requires `DigestService`.
Fix some tests.

Follow up: move some tests to runtime-os, fix few more tests.


application/src/main/java/net/corda/v5/application/crypto/DigestService.java
```diff
@@ -40,4 +40,14 @@ public interface DigestService {
...
+
+
+    /**
+     * Creates a [SecureHash].
+     * <p>
+     * This function does not validate the length of the created digest.
+     */
+
+    @Suspendable
+    SecureHash parse(@NotNull String str);
```
blsemo: I don't like the function naming - it doesn't parse a `DigestService`, so it should say what it's parsing.
```suggestion
    SecureHash parseSecureHash(@NotNull String str);
```

crypto/src/test/java/net/corda/v5/crypto/SecureHashJavaApiTest.java
```diff
@@ -5,16 +5,17 @@
 import static org.junit.jupiter.api.Assertions.assertArrayEquals;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 
+// TODO: this needs a rewrite
```
blsemo: Please don't commit/merge large commented out chunks of code.
We have git to go back to what was there.

ledger/ledger-common/src/test/java/net/corda/v5/ledger/common/transaction/TransactionSignatureExceptionJavaApiTest.java
```diff
@@ -1,13 +1,18 @@
 package net.corda.v5.ledger.common.transaction;
 
+import net.corda.v5.application.crypto.DigestService;
```
blsemo: I believe all these tests have been deleted from the API repo now.

anien: Yep, just merging main branch.

crypto/src/test/java/net/corda/v5/crypto/SecureHashJavaApiTest.java
anien: I just wanted to share my branch so it can compile. At the end, we squash and merge anyway, so this won't be seen as a change.

anien: But I agree. I try not to do that.

PR #918, opened 2023-03-07
CORE-10383 Create the interfaces for custom query registration

nkovacsx:
### Overview

This PR creates the required interfaces for the custom query registration. These include:
- `VaultNamedQuery` -- a POJO representing a named query with all the properties. It can have a filter, mapper and collector (`VaultNamedQueryCollector`, `VaultNamedQueryFilter`, `VaultNamedQueryTransformer`)
-  `VaultNamedQueryBuilderFactory` -- Used to create and build custom queries
- `VaultNamedQueryFactory` -- The main interface that needs to be implemented in CPKs in order to create custom queries
- `VaultNamedQueryRegistry` -- Used to store custom queries

### TODOs / known issues

`VaultNamedQueryRegistry` currently have a function called `registerQuery` that should NOT be on the public API because it is only used internally. However, I couldn't create an internal interface that can be wired properly because I always got an OSGi exception, will need to investigate further.

### Test notes

Please refer to the implementation PR.


driessamyn (commented):
Just looking at thisy API, it's quite difficult to see how you'd use it. E.g. which are interfaces you must implement and which is the service that is injectable? 

On the surface of it, it looks very different from the `persistence` API. 
Should we make some effort to follow similar patterns and language? I know they are 2 different use-cases, but it may feel odd for an end-user to have 2 completely different ways of querying.

The Persistence API uses a `PersistenceService` with a `query` method that returns a `ParameterizedQuery` object, which also uses the fluent builder pattern, and which has an `execute` function.

e.g. 

```kotlin
val pagedQuery = persistenceService
   .query("find_by_name_and_age", Dog::class.java)
  .setParameter("name", "Felix")
  .setParameter("maxAge", 5)
  .setLimit(100)
  .setOffset(200)

val result1 = pagedQuery.execute()
```
Maybe it's worth having an example in the JDOC?

nkovacsx:
Thanks @driessamyn I've added example usages for the classes that needs to be implemented by the user.

ledger/ledger-common/src/main/java/net/corda/v5/ledger/common/query/VaultNamedQueryBuilderFactory.java
```diff
@@ -0,0 +1,62 @@
...
+@DoNotImplement
+public interface VaultNamedQueryBuilderFactory {
+
+    /**
+     * Sets the name of the named query
+     * @param queryName Name of the query
+     * @return A builder instance with the name set
+     */
+    @Suspendable
+    @NotNull VaultNamedQueryBuilderFactory create(@NotNull String queryName);
```
lankydan: We can remove `@Suspendable` because im 100% sure we'll never suspend in these.

ledger/ledger-common/src/main/java/net/corda/v5/ledger/common/query/VaultNamedQueryBuilderFactory.java
```diff
@@ -0,0 +1,62 @@
...
+
+    /**
+     * Sets the where clause of the named query
+     * @param json The json query representation
+     * @return A builder instance with the where clause set
+     */
+    @Suspendable
+    @NotNull VaultNamedQueryBuilderFactory whereJson(@NotNull String json);
```
lankydan: Can we have a separate `VaultNamedQueryBuilder` that `create` returns. Then the rest of the functions exist on that interface instead.

ledger/ledger-common/src/main/java/net/corda/v5/ledger/common/query/VaultNamedQueryCollector.java
```diff
@@ -0,0 +1,83 @@
...
+ *     }
+ * }
+ * }</pre></li>
+ * <li>Java:<pre>{@code
+ * public class MyVaultNamedQueryCollector implements VaultNamedQueryCollector<Object, Integer> {
+ *     @NotNull
+ *     @Override
+ *     public Result<Integer> collect(List<Object> resultSet, Map<String, Object> parameters) {
+ *         return new VaultNamedQueryCollector.Result<>(
+ *                 Collections.singletonList(
```
lankydan: You can do `List.of` as of Java 9

ledger/ledger-common/src/main/java/net/corda/v5/ledger/common/query/VaultNamedQueryFactory.java
```diff
@@ -0,0 +1,60 @@
...
+ * <li>Java:<pre>{@code
+ * public class MyVaultNamedQueryFactory implements VaultNamedQueryFactory {
+ *     @Override
+ *     public void create(@NotNull VaultNamedQueryBuilderFactory vaultNamedQueryBuilderFactory) {
+ *         vaultNamedQueryBuilderFactory.create("FIND_WITH_CORDA_COLUMNS")
+ *                 .whereJson(
+ *                         "WHERE custom ->> 'TestUtxoState.testField' = :testField " +
+ *                                 "AND custom ->> 'Corda.participants' IN :participants " +
+ *                                 "AND custom ? :contractStateType "
+ *                         "AND custom ->> 'Corda.createdTimestamp' > :created_timestamp"
```
lankydan: Formatting of the `AND`.

ledger/ledger-common/src/main/java/net/corda/v5/ledger/common/query/VaultNamedQueryTransformer.java
```diff
@@ -0,0 +1,38 @@
...
+ * }
+ * }</pre></li></ul>
+ *
+ * @param <S> Type of the state returned from the database
+ * @param <T> Type that the original state is transformed into
+ */
+public interface VaultNamedQueryTransformer<S, T> {
+    @Suspendable
+    @NotNull
+    T transform(@NotNull S state, @NotNull Map<String, Object> parameters);
```
lankydan: While I don't disagree with the idea of having common interfaces, for now, we're only concerned with the UTXO ledger and therefore we should put our code there.

Doing this also means we can have `S extends ContractState`.

Can we also rename the generics to `T` for the state and `R` for the return type.

ledger/ledger-common/src/main/java/net/corda/v5/ledger/common/query/VaultNamedQueryBuilderFactory.java
nkovacsx: I'm happy to make that change but do you think it's worth having one more interface just for a `create` function? I mean I don't have a strong opinion but not sure if it's worth it.

ledger/ledger-common/src/main/java/net/corda/v5/ledger/common/query/VaultNamedQueryTransformer.java
nkovacsx: Do you mean every interface that I've created should be under utxo?

lankydan:
> Just looking at thisy API, it's quite difficult to see how you'd use it. E.g. which are interfaces you must implement and which is the service that is injectable?
> 
> On the surface of it, it looks very different from the `persistence` API. Should we make some effort to follow similar patterns and language? I know they are 2 different use-cases, but it may feel odd for an end-user to have 2 completely different ways of querying.
> 
> The Persistence API uses a `PersistenceService` with a `query` method that returns a `ParameterizedQuery` object, which also uses the fluent builder pattern, and which has an `execute` function.
> 
> e.g.
> 
> ```kotlin
> val pagedQuery = persistenceService
>    .query("find_by_name_and_age", Dog::class.java)
>   .setParameter("name", "Felix")
>   .setParameter("maxAge", 5)
>   .setLimit(100)
>   .setOffset(200)
> 
> val result1 = pagedQuery.execute()
> ```
> 
> Maybe it's worth having an example in the JDOC?

I think its mainly because only the DB worker, query registration code is done atm. The calling side should follow the persistence service closer.

ledger/ledger-utxo/src/main/java/net/corda/v5/ledger/utxo/query/VaultNamedQuery.java
```diff
@@ -0,0 +1,43 @@
...
+import net.corda.v5.base.annotations.DoNotImplement;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+/**
+ * An interface representing a named ledger query. A named ledger query must have a name associated with it,
+ * the rest of the fields are optional.
+ */
+@DoNotImplement
+public interface VaultNamedQuery {
```
lankydan: I don't think we need this interface. Do we?

ledger/ledger-utxo/src/main/java/net/corda/v5/ledger/utxo/query/VaultNamedQueryCollector.java
```diff
@@ -0,0 +1,83 @@
...
+public interface VaultNamedQueryCollector<R, T> {
+
+    /**
+     * @param resultSet The original result set that was returned by the named query execution
+     * @param parameters Parameters that were present in the query
+     * @return A result that the original result set was collected into
+     */
+    @Suspendable
+    @NotNull
+    Result<T> collect(@NotNull List<R> resultSet, @NotNull Map<String, Object> parameters);
```
lankydan, line 50: Remove the suspendables from all the interfaces.

lankydan (commented):
2 small comments but looks good.

ledger/ledger-utxo/src/main/java/net/corda/v5/ledger/utxo/query/VaultNamedQuery.java
```diff
@@ -0,0 +1,43 @@
...
+
+    /**
+     * @return The name of the named ledger query
```
lankydan: Pedantic comment. Can you use `.` at the end of every sentence/single line to keep things consistent. Think its the same in other files.

ledger/ledger-utxo/src/main/java/net/corda/v5/ledger/utxo/query/VaultNamedQueryBuilder.java
```diff
@@ -0,0 +1,46 @@
...
+     * Sets the where clause of the named query
+     * @param json The json query representation
+     * @return A builder instance with the where clause set
+     */
+    @NotNull
+    VaultNamedQueryBuilder whereJson(@NotNull String json);
+
+    /**
+     * Sets the filter function of the named query
+     * @param filter A filter object
```
lankydan: Put a new line between the params and returns, something like:

```
* Sets the filter function of the named query.
* 
* @param filter A filter object.
* 
* @return A builder instance with the filter function set.
```

ledger/ledger-utxo/src/main/java/net/corda/v5/ledger/utxo/query/VaultNamedQuery.java
nkovacsx: good point

lankydan (dismissed):

lankydan (approved):

PR #922, opened 2023-03-07
CORE-10646: Add a new API for flow external messaging

owenstanford:
This change adds a new API that enables flows to send messages to an external system via a named Kafka channel.


driessamyn (dismissed):
LGTM

driessamyn (approved):
Still LGTM ;)

PR #923, opened 2023-03-08
CORE-11434: Case insensitive maintenance mode REST request

josephzunigadaly:
* Add new Avro type `VirtualNodeState` with values matching `OperationalStatus`
* Replace `String` status values with `VirtualNodeState` in Avro types

Runtime OS PR: https://github.com/corda/corda-runtime-os/pull/3333

### DIFF
--- a/data/avro-schema/src/main/resources/avro/net/corda/data/virtualnode/VirtualNodeState.avsc
+++ b/data/avro-schema/src/main/resources/avro/net/corda/data/virtualnode/VirtualNodeState.avsc
@@ -1,0 +1,10 @@
+{
+  "type": "enum",
+  "name": "VirtualNodeState",
+  "namespace": "net.corda.data.virtualnode",
+  "doc": "Virtual Node state.",
+  "symbols": [
+    "ACTIVE",
+    "INACTIVE"
+  ]
+}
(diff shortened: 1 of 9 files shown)


vkolomeyko (dismissed):
LGTM

data/avro-schema/src/main/resources/avro/net/corda/data/virtualnode/VirtualNodeState.avsc
```diff
@@ -0,0 +1,10 @@
+{
+  "type": "enum",
+  "name": "VirtualNodeState",
+  "namespace": "net.corda.data.virtualnode",
+  "doc": "Virtual Node state.",
```
sprintengine: I don't think naming this `VirtualNodeState` is quite correct, it's the status of an operational characteristic of a virtual node, for example, in the future a user may request to put their vnode into "ROLLING_UPGRADE" at the REST endpoint, which would result in flowP2pOperationalStatus and vaultDbOperationalStatus set to INACTIVE, but the others ACTIVE.

Probably a better name - `VirtualNodeOperationalState`

But this is just a minor internal thing, I'm happy to approve as generally this change is a good idea.

conal-smith (dismissed):

Tom-Fitzpatrick (dismissed):

data/avro-schema/src/main/resources/avro/net/corda/data/virtualnode/VirtualNodeState.avsc
josephzunigadaly: I have pushed a rename to `VirtualNodeOperationalState`.

conal-smith (approved):

vkolomeyko (approved):
LGTM

PR #925, opened 2023-03-08
CORE-9724 Make `SecureHash` an interface on public API

kyriathar:
- makes `SecureHash` an interface and decouples it on API level from `OpaqueBytes` and `ByteSequence`
- moves `SecureHash.parse` to `DigestService` to be available on public API
- adds `SecureHash.getBytes` (previously inherited from `OpaqueBytes`)
- adds `DigestService` to `StateRef.parse` parameters  for `SecureHash` parsing
- adds `DigestService` to `UtxoLedgerTokenStateObserver.onCommit`  parameters for `SecureHash` parsing 
- removes `SecureHash.prefixChars` APIs to not encourage use of short hashes in flows.
             
runtime-os PR: [#3338](https://github.com/corda/corda-runtime-os/pull/3338)


application/src/main/java/net/corda/v5/application/crypto/DigestService.java
```diff
@@ -33,6 +33,15 @@ public interface DigestService {
     @NotNull
     SecureHash hash(@NotNull InputStream inputStream, @NotNull DigestAlgorithmName digestName);
 
+    /**
+     * Creates a {@link SecureHash}.
+     * <p>
+     * This function does not validate the length of the created digest. Prefer using
+     * <code>HashingService#parse</code> for a safer mechanism for creating {@link SecureHash}es.
```
anien: Is there a HashingService? I couldn't find any!

kyriathar: Good catch, removed

anien (dismissed):
LGTM!

application/src/test/java/net/corda/v5/crypto/merkle/MerkleProofJavaApiTest.java
```diff
@@ -16,15 +16,15 @@ class MerkleProofJavaApiTest {
     void verify() {
         when(merkleProof.verify(any(), any())).thenReturn(true);
 
-        final SecureHash hash = new SecureHash("SHA-256", "123".getBytes());
+        final SecureHash hash = mock(SecureHash.class);
```
driessamyn, line 19: I believe these tests can actually be deleted since we converted the API to JAVA.

crypto/src/main/java/net/corda/v5/crypto/SecureHash.java
```diff
@@ -1,109 +1,55 @@
...
-     * @param algorithm Hashing algorithm which was used to generate the hash.
-     * @param bytes     Hash value.
+     * Hashing algorithm which was used to generate the hash.
      */
-    public SecureHash(@NotNull String algorithm, @NotNull byte[] bytes) {
-        super(bytes);
-        this.algorithm = algorithm;
-    }
-
-    static final char DELIMITER = ':';
```
driessamyn, line 33: I wonder if this needs to be still in the API as we basically prescribing a contract that the string representation must be `algo:hash`

And maybe that means that the function to convert a `SecureHash` interface into a string should be on the API as well as effectively that cannot be changed?

ledger/ledger-utxo/src/main/java/net/corda/v5/ledger/utxo/StateRef.java
```diff
@@ -67,12 +68,12 @@ public int getIndex() {
      * @return Returns a {@link StateRef} parsed from the specified {@link String} value.
      * @throws IllegalArgumentException if the specified value cannot be parsed.
      */
-    public static StateRef parse(@NotNull final String value) {
+    public static StateRef parse(@NotNull final String value, DigestService digestService) {
```
driessamyn, line 71: Maybe the `DigestService` should be injected instead?
This looks a bit odd.

And maybe this should be an interface instead too so we don't need implementation detail in the API?
Probably a question for @blsemo or @lankydan ?

ledger/ledger-utxo/src/main/java/net/corda/v5/ledger/utxo/observer/UtxoLedgerTokenStateObserver.java
```diff
@@ -82,5 +83,5 @@
      * @return Returns a {@link UtxoToken}.
      */
     @NotNull
-    UtxoToken onCommit(@NotNull T state);
+    UtxoToken onCommit(@NotNull T state, @NotNull DigestService digestService);
```
driessamyn, line 86: Similar to above, does this need to be an argument here or should the service just be injected into the implementation?

application/src/test/java/net/corda/v5/crypto/merkle/MerkleProofJavaApiTest.java
kyriathar: Yes and they should, however I thought maybe addressing under different PR to not mix work. Happy to address here too though.

crypto/src/main/java/net/corda/v5/crypto/SecureHash.java
kyriathar: Good point. Will put `DELIMITER` back to API.

Not sure how function to convert a `SecureHash` can be put back to API without using the constructor. Sounds like we should be introducing a `SecureHash` factory? Although not really sure how that much adds at this point. But yes `parseSecureHash` sounds like it needs to be put back to API as something to never be changed.

kyriathar: I think same applies for `SecureHash.toString`. Would we want it to be on API somehow?

driessamyn: You're right. `toString` could be added (e.g. `CryptoUtil.secureHashToString(hash: secureHash): String`, but not the other way around. But may be best to keep them both internal to the implementation. The implication is just that the format is implicit, so should be documentated in the toString and parse JDOC.

ledger/ledger-utxo/src/main/java/net/corda/v5/ledger/utxo/observer/UtxoLedgerTokenStateObserver.java
kyriathar: I think this can't be injected to implementation because CorDapp developers are to implement this one. @owenstanford could you please confirm?

crypto/src/main/java/net/corda/v5/crypto/SecureHash.java
```diff
@@ -1,109 +1,55 @@
...
-        return other instanceof SecureHash && this.algorithm.equals(((SecureHash) other).algorithm) && super.equals(other);
-    }
-
-    /**
-     * Returns a hash code value for the object.
-     */
-    public int hashCode() {
-        return ByteBuffer.wrap(this.getBytes()).getInt();
-    }
+    String prefixChars();
```
kyriathar: @driessamyn should we be removing this overload? Under the hood just calls `prefixChars(6)`

ledger/ledger-utxo/src/main/java/net/corda/v5/ledger/utxo/observer/UtxoLedgerTokenStateObserver.java
owenstanford: That's correct, this is user code executed in the DB sandbox which doesn't support service injection like the flow sandbox, we could add it of course, but this is probably a reasonable option.

ledger/ledger-utxo/src/main/java/net/corda/v5/ledger/utxo/StateRef.java
blsemo: `StateRef` is supposed to be a lightweight data class that any user can create/parse whenever they need one - I don't want that class itself to be hidden behind an interface/factory pattern.
I don't think we can inject the `DigestService` into `StateRef` as it stands right now, `@CordaInject` doesn't work in random classes, and as it is a concrete class, constructor injection would just shift the problem to the public constructor.
If we don't want to pass in the `DigestService` in this function, then we need to make it an interface, have a factory interface, get a factory off the `UtxoLedgerService` and serve up/parse `StateRef` instances via the factory - which seems like massive overkill for a small data class, and as people have pointed out on slack, would be bad user experience.

crypto/src/main/java/net/corda/v5/crypto/SecureHash.java
kyriathar: Added `SecureHash.DELIMITER` back to API.

mnesbit: I am fine with saying that algorithm name may only match the regex [a-zA-Z_][a-zA-Z_0-9\\-/]* so that : is always a safe separator. We need to also mandate a maximum algorithm name length as we store hash strings in DB columns and need to allow for SHA-512 data and sensible sized algorithm names eg. HmacPBESHA512/256 is a Mac algorithm name in JDK 17. From that we can mandate a suitable future proof column size

ledger/ledger-utxo/src/main/java/net/corda/v5/ledger/utxo/StateRef.java
mnesbit: I think this probably is best we can do right now

crypto/src/main/java/net/corda/v5/crypto/SecureHash.java
kyriathar: All `prefixChars` APIs removed after internal communication.

application/src/main/java/net/corda/v5/application/crypto/DigestService.java
```diff
@@ -33,6 +33,22 @@ public interface DigestService {
     @NotNull
     SecureHash hash(@NotNull InputStream inputStream, @NotNull DigestAlgorithmName digestName);
 
+    /**
+     * Parses a secure hash in string form into a {@link SecureHash}.
+     * <p>
+     * A valid secure hash string should be containing the algorithm and hexadecimal representation of the bytes
+     * separated by the colon character (':') ({@link net.corda.v5.crypto.SecureHash.DELIMITER}).
+     * <p>
+     * Please note that this function does not validate the length of the hex string of the digest
```
mnesbit: https://github.com/corda/corda-api/blob/release/os/5.0/crypto-extensions/src/main/java/net/corda/v5/crypto/extensions/DigestAlgorithm.java#L59 should ensure that we can carry out that check

kyriathar: Good catch. Since we have now moved to `SecureHash` parsing in `DigestService` means we now have control over digest algorithms available. Therefore `DigestService.parseSecureHash` implementation should be changed to now take digest algorithm into account for validations during secure hash parsing. Raised [CORE-11656](https://r3-cev.atlassian.net/browse/CORE-11656) for it. Will remove that note from kDoc.

driessamyn (dismissed):
LGTM

mnesbit (dismissed):
ok

mnesbit (approved):
OK

driessamyn (approved):
LGTM

PR #926, opened 2023-03-09
CORE-6564 Add functionality to allow counterparty flow information to  be exchanged

LWogan:
Exchange a map of properties as part of session initiation. 
move the protocol info into the session properties. 
Include a new SessionConfirm message which will be sent in response to SessionInits to pass the session properties from initiated to initiating sessions. 
add a waiting for handler to wait for session confirmation before trying to return FlowInfo to the Fiber.


mbrkic-r3 (dismissed):
LGTM

application/src/main/java/net/corda/v5/application/messaging/FlowInfo.java
```diff
@@ -0,0 +1,16 @@
...
+public interface FlowInfo {
+    /**
+     * @return The protocol name the flow is running.
+     */
+    String protocol();
+
+    /**
+     * @return The protocol version the flow is running.
+     */
+    Integer protocolVersion();
```
mbrkic-r3, line 18: Is this mandatory?

LWogan: yes. added NotNull annotation

mbrkic-r3 (approved):
LGTM

mnesbit (approved):
OK, but I will discuss with @JamesHR3 about clarifying role with regard to existing flow properties API

PR #929, opened 2023-03-09
CORE-11521 Remove `DigitalSignature.WithKey.context`

kyriathar:
`DigitalSignature.WithKey.context` gets returned as part of the signature. Callers of sign operation have no control over it so it's of no use to them (and also as such `DigitalSignature.WithKey.context` is not signed either).

- removes `DigitalSignature.WithKey.context`

runtime-os PR: [#3347](https://github.com/corda/corda-runtime-os/pull/3347)

### DIFF
--- a/data/avro-schema/src/main/resources/avro/net/corda/data/crypto/wire/CryptoSignatureWithKey.avsc
+++ b/data/avro-schema/src/main/resources/avro/net/corda/data/crypto/wire/CryptoSignatureWithKey.avsc
@@ -13,11 +13,6 @@
       "name": "bytes",
       "type": "bytes",
       "doc": "Byte array of the signature, exactly as returned by crypto signing operations"
-    },
-    {
-      "name": "context",
-      "type": "net.corda.data.KeyValuePairList",
-      "doc": "The optional key/value operation specific context."
     }
   ]
 }
--- a/application/src/test/java/net/corda/v5/application/crypto/SigningServiceJavaApiTest.java
+++ b/application/src/test/java/net/corda/v5/application/crypto/SigningServiceJavaApiTest.java
@@ -22,8 +22,7 @@ class SigningServiceJavaApiTest {
     void signWithByteArrayAndSignatureSpecTest() {
         final DigitalSignature.WithKey signatureWithKey = new DigitalSignature.WithKey(
             publicKey,
-            "test".getBytes(),
-            new HashMap<>()
+            "test".getBytes()
         );
         Mockito.when(signingService.sign(any(), any(), any(SignatureSpec.class))).thenReturn(signatureWithKey);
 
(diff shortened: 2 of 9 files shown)


data/avro-schema/src/main/resources/avro/net/corda/data/crypto/wire/CryptoSignatureWithKey.avsc
```diff
@@ -13,11 +13,6 @@
       "name": "bytes",
       "type": "bytes",
       "doc": "Byte array of the signature, exactly as returned by crypto signing operations"
-    },
-    {
```
charlieR3, line 17: Could `net.corda.data.crypto.wire.CryptoSignatureSpec` be added here? Are there cases we would have a signature and not need to know the signature spec to verify it? Seems like everywhere `CryptoSignatureWithKey` is referenced in other schemas, `CryptoSignatureSpec` has to be added also.

kyriathar: Valid comment! So the reasoning for removing it from here, is to not allow that exactly. I.e. this object will be created at the signature calculation site (having passed in also a signature spec). However, we shouldn't be allowing signature spec  to be populated there because it means the caller of the sign operation relies on whatever is returned by whoever signed. The caller of the sign operation should rather be the one adding the signature spec (which is the expected one) to the signature packet (Moreover, the signature spec should actually be serialized and added to the to be signed payload. Although in your case you have been using that internally but still that is the proper use. There is a nice example of it at `net.corda.ledger.common.flow.impl.transaction.TransactionSignatureServiceImpl`). @mnesbit could you please confirm. Also had a chat about this with @dimosr.

We could be creating some SignatureWithSignatureSpec avro type to be re-used, but that should be created and populated at sign call site.

kyriathar: So I believe `CryptoSignatureWithKey` needs to remain as is now.

mnesbit: The signature spec when creating a signature should be calculated from PublicKey and hash type you want to use. There are then two choices either
1. The hash type is implicit in which case it should not be serialized with the message. Instead receiving side should calculate the SignatureSpec in same way as signer via the service and check signature. This is in some ways more secure, but can be had with mixed software versions.
2. The alternative is to include the value inside the object being signed. Receiver uses this and checks it is on the allowed list of SignatureSpecs for the type of PublicKey and then verifies the signature. This is more version proof but it modifies the data being signed, which is problematic for the MemberInfo as we have two signers, though approach of transactions is to make a metadata structure with txid and SignatureSpec (and possibly timestamp, or other useful stuff) and sign that. Then serialize the metadata along with transaction (txid is implicit) and verifier checks signature over recalculated metadata rather than simply over the object hash.

charlieR3 (approved):

PR #932, opened 2023-03-10
CORE-10427 - Alias Identity Mapper

julia-filipczak:
Adding interop to hosted identities topic consumers to allow reading from the topic for work in https://github.com/corda/corda-runtime-os/pull/3299


jmacmahonr3 (approved):
LGTM

PR #933, opened 2023-03-10
CORE-11034: Add default row for Sandbox config and update the schema 

fowlerrr:
Default row in the config table was missing and the schema has been changed to support a cache for each sandbox type

### DIFF
--- a/data/config-schema/src/main/resources/net/corda/schema/configuration/sandbox/1.0/corda.sandbox.json
+++ b/data/config-schema/src/main/resources/net/corda/schema/configuration/sandbox/1.0/corda.sandbox.json
@@ -5,16 +5,63 @@
   "description": "Configuration schema for the sandbox subsection.",
   "type": "object",
   "properties": {
-    "cache": {
-      "description": "Settings for sandbox caching",
+    "flow": {
+      "description": "Settings for flow sandbox",
       "type": "object",
       "default": {},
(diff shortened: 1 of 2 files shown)


data/config-schema/src/main/resources/net/corda/schema/configuration/sandbox/1.0/corda.sandbox.json
```diff
@@ -5,16 +5,63 @@
...
+        "cache": {
+          "description": "Settings for flow sandbox caching",
+          "type": "object",
+          "default": {},
+          "properties": {
+            "size": {
+              "description": "The maximum number cached flow sandboxes.",
+              "type": "integer",
+              "minimum": 0,
+              "default": 2
```
driessamyn: Sorry to be a pain.
Can we set all defaults to 10? It's a bit arbitrary, but I hope it's a more appropriate level and we can review again after we see the impact on the soaktests and/or perf testing.
FYI @JamesHR3 @davidcurrie

fowlerrr: Cool, I've changed that.  (sorry, I acidentally amended the commit so a bit of history is lost of the PR)

driessamyn (approved):
LGTM

PR #936, opened 2023-03-13
CORE-9586 Allow Custom Fields in Registration schema

williamvigorr3:
Allow extra string fields in the registration context. These will be used to allow registering member to add custom fields to their member info. Enforce that the custom fields start with a "ext." prefix, all other fields are reserved for future platform usage. Custom fields can have 256 characters in the value.

### DIFF
--- a/data/membership-schema/src/main/resources/net/corda/schema/membership/member/dynamic/registration/1.0/corda.member.dynamic.registration.json
+++ b/data/membership-schema/src/main/resources/net/corda/schema/membership/member/dynamic/registration/1.0/corda.member.dynamic.registration.json
@@ -38,7 +38,7 @@
       "type": "string"
     },
     "(?=^(?!corda\\.).+)(?=^(?=.{1,128}$)[a-zA-Z0-9]*[^$%^&*;:,<>?()\"']*$)": {
-      "description": "Optional. A user can specify additional properties, they must not start with corda or be more than 128 characters.",
+      "description": "Optional. A user can specify additional properties, they must not start with corda. or be more than 128 characters.",
       "type": "string",
       "maxLength": 128
     }
@@ -77,6 +77,5 @@
   "required": [
     "corda.session.key.id"
   ],
-  "additionalProperties": false,
-  "maxProperties": 128
+  "additionalProperties": false
 }
--- a/data/membership-schema/src/main/resources/net/corda/schema/membership/member/dynamic/registration/1.0/corda.member.dynamic.registration.json
+++ b/data/membership-schema/src/main/resources/net/corda/schema/membership/member/dynamic/registration/1.0/corda.member.dynamic.registration.json
@@ -37,7 +37,7 @@
       "description": "Optional. The ledger public key signature spec. Can be multiple specified by index.",
       "type": "string"
     },
-    "^ext\\.[a-zA-Z0-9\\.]+$": {
+    "^ext\\.[a-zA-Z0-9.]+$": {
       "description": "Optional. A user can specify additional properties, they must be prefixed with `ext.`.",
       "type": "string",
       "maxLength": 256
--- a/data/membership-schema/src/main/resources/net/corda/schema/membership/member/dynamic/registration/1.0/corda.member.dynamic.registration.json
+++ b/data/membership-schema/src/main/resources/net/corda/schema/membership/member/dynamic/registration/1.0/corda.member.dynamic.registration.json
@@ -37,10 +37,10 @@
       "description": "Optional. The ledger public key signature spec. Can be multiple specified by index.",
       "type": "string"
     },
-    "(?=^(?!corda\\.).+)(?=^(?=.{1,128}$)[a-zA-Z0-9]*[^$%^&*;:,<>?()\"']*$)": {
-      "description": "Optional. A user can specify additional properties, they must not start with corda. or be more than 128 characters.",
+    "^ext\\.[a-zA-Z0-9\\.]+$": {
+      "description": "Optional. A user can specify additional properties, they must be prefixed with `ext.`",
       "type": "string",
-      "maxLength": 128
+      "maxLength": 256
     }
   },
   "properties": {
--- a/data/membership-schema/src/main/resources/net/corda/schema/membership/member/dynamic/registration/1.0/corda.member.dynamic.registration.json
+++ b/data/membership-schema/src/main/resources/net/corda/schema/membership/member/dynamic/registration/1.0/corda.member.dynamic.registration.json
@@ -36,6 +36,11 @@
     "^corda.ledger.keys.[0-9]+.signature.spec$": {
       "description": "Optional. The ledger public key signature spec. Can be multiple specified by index.",
       "type": "string"
+    },
+    "(?=^(?!corda\\.).+)(?=^(?=.{1,128}$)[a-zA-Z0-9]*[^$%^&*;:,<>?()\"']*$)": {
+      "description": "Optional. A user can specify additional properties, they must not start with corda or be more than 128 characters.",
+      "type": "string",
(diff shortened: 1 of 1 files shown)


data/membership-schema/src/main/resources/net/corda/schema/membership/member/dynamic/registration/1.0/corda.member.dynamic.registration.json
```diff
@@ -36,6 +36,11 @@
     "^corda.ledger.keys.[0-9]+.signature.spec$": {
       "description": "Optional. The ledger public key signature spec. Can be multiple specified by index.",
       "type": "string"
+    },
+    "(?=^(?!corda\\.).+)(?=^(?=.{1,128}$)[a-zA-Z0-9]*[^$%^&*;:,<>?()\"']*$)": {
```
dimosr: This regex is a bit scary.. I am a bit worried it might be a maintenance pain in the future. It made me wonder; should we reconsider and instead limit custom attributes under a specific prefix, say `ext.`?

data/membership-schema/src/main/resources/net/corda/schema/membership/member/dynamic/registration/1.0/corda.member.dynamic.registration.json
```diff
@@ -36,6 +36,11 @@
...
+      "description": "Optional. A user can specify additional properties, they must not start with corda. or be more than 128 characters.",
```
nikinagy:
```suggestion
      "description": "Optional. A user can specify additional properties, they must not start with 'corda.' prefix or be more than 128 characters.",
```

data/membership-schema/src/main/resources/net/corda/schema/membership/member/dynamic/registration/1.0/corda.member.dynamic.registration.json
charlieR3: Maybe we should be more restrictive so that it is something like `[a-zA-Z0-9\.]*` instead of `[a-zA-Z0-9]*[^$%^&*;:,<>?()\"']*`

I tested out a couple keys and these are all valid with this regex
```
my.value=
my.value+
my.value_
my.value-
my.value#
my.value!
my.value@
my.value€
my.value|
my.value\
```

williamvigorr3: Yeah regex is not nice or easy to read. However I think the main problem is the: `(?=^(?=.{1,128}$)[a-zA-Z0-9]*[^$%^&*;:,<>?()\"']*$)` which limits the key size to 128 characters.

dimosr: that's true. I see we have the checks on the verifier performed on the member side as well, so we could consider removing the validation from the schema if it makes things a bit easier to understand and maintain. 

Having a specific namespace for custom properties sounds attractive anyway though, since we'll be able to reason better about where they are. For example, if we want to evolve it further in the future (e.g. add extra category of fields), we won't have to adjust the logic that filters them out. It feels a bit more robust. I think we've discussed that in the past, but I can't recall if there was an important reason we didn't go for it.

charlieR3: Not sure there was an important reason. I was thinking previously that using the `corda` namespace as our own protected namespace, and allowing the cordapp developers to define their own namespace separate to that made sense rather than having the cordapp developer namespace as a subset of the corda namespace. So they could have keys like `company.myKey` (where `company` could be the name of the company developing the cordapp) instead of `corda.ext.company.myKey`. I'm not strongly opposed to using `corda.ext` though.

dimosr: Sorry, I might not have been clear. My proposal was about using a new "root" namespace outside of `corda.`. I agree that it's good to preserve the `corda.` namespace for everything defined by the platform.

So, that would be:
* everything under `ext.` is custom application properties. In your example, it would be `ext.companyA.keyA`, `ext.companyB.keyB` etc.
* everything under `corda.` is defined & used by the platform.

williamvigorr3: Let's go with that (unless Charlie objects). I don't think it matters too much really, as long as we reserve some namespace for ourselves, which we can evolve in the future.

williamvigorr3: One thing to note is this schema is only checked on the member side and we can have infinite many `corda.` fields. Maybe we should restrict the total number of fields or the total size in the member context.

charlieR3: No objection from me 👍

dimosr: > Maybe we should restrict the total number of fields or the total size in the member context.

That's a good point. We can introduce an extra pair of limits for `corda.` fields, one for max number of fields and another one on max value size (this needs to be quite conservative). But, I think we can do that later, no need to spend too much time on it now as part of this ticket.

williamvigorr3: Should I add a ticket under this EPIC: https://r3-cev.atlassian.net/browse/CORE-11226?

dimosr: yes, please - thanks

williamvigorr3: I added https://r3-cev.atlassian.net/browse/CORE-11768

data/membership-schema/src/main/resources/net/corda/schema/membership/member/dynamic/registration/1.0/corda.member.dynamic.registration.json
williamvigorr3: I reworded this.

data/membership-schema/src/main/resources/net/corda/schema/membership/member/dynamic/registration/1.0/corda.member.dynamic.registration.json
```diff
@@ -36,6 +36,11 @@
     "^corda.ledger.keys.[0-9]+.signature.spec$": {
       "description": "Optional. The ledger public key signature spec. Can be multiple specified by index.",
       "type": "string"
+    },
+    "^ext\\.[a-zA-Z0-9\\.]+$": {
```
dimosr: Is this supposed to be
```suggestion
    "^ext\.[a-zA-Z0-9\.]+$": {
```

williamvigorr3: We need to use `\\` because you have to escape `\` inside json. I think it should be `^ext\\.[a-zA-Z0-9.]+$` (escaping the `.` inside json is redundant).

dimosr (approved):

PR #938, opened 2023-03-14
DOC-4884: Tech Writer review of messaging config descriptions

nadinequinr3:
Tech Writing review of descriptions. These descriptions are pulled in the docs portal. (Draft preview currently here: https://nadine.preview.docs.r3.com/en/platform/corda/5.0-beta/operating/configuration/messaging.html)

### DIFF
--- a/data/config-schema/src/main/resources/net/corda/schema/configuration/messaging/1.0/kafka-properties.json
+++ b/data/config-schema/src/main/resources/net/corda/schema/configuration/messaging/1.0/kafka-properties.json
@@ -6,7 +6,7 @@
   "type": "object",
   "properties": {
     "common": {
-      "description": "Kafka properties that are common between consumers and producers. For example, connection details.",
+      "description": "Kafka properties that are common to both consumers and producers. For example, connection information such as SSL details.",
       "type": "object"
     }
   }
--- a/data/config-schema/src/main/resources/net/corda/schema/configuration/messaging/1.0/kafka-properties.json
+++ b/data/config-schema/src/main/resources/net/corda/schema/configuration/messaging/1.0/kafka-properties.json
@@ -6,7 +6,7 @@
   "type": "object",
   "properties": {
     "common": {
-      "description": "Kafka properties that are common between consumers and producers, e.g. connection details.",
+      "description": "Kafka properties that are common between consumers and producers. For example, connection details.",
       "type": "object"
     }
   }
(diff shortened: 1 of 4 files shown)


data/config-schema/src/main/resources/net/corda/schema/configuration/messaging/1.0/kafka-properties.json
```diff
@@ -6,7 +6,7 @@
   "type": "object",
   "properties": {
     "common": {
-      "description": "Kafka properties that are common between consumers and producers, e.g. connection details.",
+      "description": "Kafka properties that are common between consumers and producers. For example, connection details.",
```
nadinequinr3: This description seems like it needs more information. Examples maybe?

thiagoviana: @LWogan do you see any example we could have here?

LWogan: "For example, connection information such as SSL details and bootstrap server address."

or something like that. There's lots of connection details which are common.

some real examples for context:
https://kafka.apache.org/documentation/#consumerconfigs_bootstrap.servers
https://kafka.apache.org/documentation/#producerconfigs_bootstrap.servers

https://kafka.apache.org/documentation/#producerconfigs_ssl.key.password
https://kafka.apache.org/documentation/#consumerconfigs_ssl.key.password

thiagoviana: @nadinequinr3 what do you think? 

"Kafka properties that are common between consumers and producers. For example, connection information such as SSL details and/or bootstrap server address."

LWogan: actually in hindsight best not to mention bootstrap server address as we actually manually override it with the one the node operator uses to deploy corda. it is one of the things we dont allow to be set. stick to SSL details as an example if we want to include one

LWogan (approved):
LGTM

data/config-schema/src/main/resources/net/corda/schema/configuration/messaging/1.0/kafka-properties.json
nadinequinr3: Sorry, I meant more an example of an actual value for this field. Or would the expected format be obvious to users? Would it make sense to refer to the Kafka documentation? Something like:

`Kafka properties that are common to both consumers and producers. For example, connection information, such as SSL details. For information about all possible consumer and producer configuration values, see the Apache Kafka documentation.`

@LWogan

PR #940, opened 2023-03-14
CORE-11422: Allow membership to produce requests to itslef

yift-r3:
I tested a few times using:
```
cd ~/corda-non-functional-test/
scripts/build_cpi.sh
./gradlew test -Dcorda.namespace="yift-cluster-a" -Dtest.duration=PT3M --tests="*MembershipWorkerFailureTests*"
```


dimosr (dismissed):
LGTM - can't we merge this without an api version bump though? Seems like a non-breaking change.

dimosr (approved):

PR #941, opened 2023-03-14
CORE-11065 Custom ledger queries liquibase changes

lankydan:
Add `custom_representation` and repurpose the `consumed` column
in the `utxo_relevant_transaction_state` table.

The `custom_representation` column is of type `JSONB` and is
therefore Postgres  specific.

`ledger-utxo-relevant-transaction-state-creation-v1.0.xml` has been 
separated to create the table. For now it has been restricted to 
creating the table only for Postgres databases, meaning that the table
will not be made for anything else. In the future, we can add the table
with the correct columns when we support more databases.

### DIFF
--- a/data/db-schema/src/main/resources/net/corda/db/schema/vnode-vault/migration/ledger-utxo-creation-v1.0.xml
+++ b/data/db-schema/src/main/resources/net/corda/db/schema/vnode-vault/migration/ledger-utxo-creation-v1.0.xml
@@ -156,32 +156,6 @@
                                  referencedColumnNames="transaction_id,group_idx,leaf_idx"
                                  referencedTableName="utxo_transaction_component"/>
 
-        <createTable tableName="utxo_relevant_transaction_state">
-            <column name="transaction_id" type="VARCHAR(160)">
-                <constraints nullable="false"/>
-            </column>
-            <column name="group_idx" type="INT">
-                <constraints nullable="false"/>
-            </column>
-            <column name="leaf_idx" type="INT">
-                <constraints nullable="false"/>
-            </column>
-            <column name="consumed" type="BOOLEAN">
-                <constraints nullable="false"/>
-            </column>
-            <column name="created" type="TIMESTAMP">
-                <constraints nullable="false"/>
-            </column>
-        </createTable>
-        <addPrimaryKey columnNames="transaction_id, group_idx, leaf_idx"
-                       constraintName="utxo_relevant_transaction_state_pkey"
-                       tableName="utxo_relevant_transaction_state"/>
-        <addForeignKeyConstraint baseColumnNames="transaction_id,group_idx,leaf_idx"
-                                 baseTableName="utxo_relevant_transaction_state"
-                                 constraintName="fk_utxo_relevant_transaction_state"
-                                 referencedColumnNames="transaction_id,group_idx,leaf_idx"
-                                 referencedTableName="utxo_transaction_component"/>
-
         <createTable tableName="utxo_cpk">
             <column name="file_checksum" type="VARCHAR(255)">
                 <constraints nullable="false" primaryKey="true"/>
@@ -217,4 +191,6 @@
                        constraintName="utxo_transaction_cpk_pkey"
                        tableName="utxo_transaction_cpk"/>
     </changeSet>
+
+    <include file="net/corda/db/schema/vnode-vault/migration/ledger-utxo-relevant-transaction-state-creation-v1.0.xml"/>
 </databaseChangeLog>
--- a/data/db-schema/src/main/resources/net/corda/db/schema/vnode-vault/migration/ledger-utxo-relevant-transaction-state-creation-v1.0.xml
+++ b/data/db-schema/src/main/resources/net/corda/db/schema/vnode-vault/migration/ledger-utxo-relevant-transaction-state-creation-v1.0.xml
@@ -1,0 +1,40 @@
+<?xml version="1.1" encoding="UTF-8" standalone="no"?>
+<databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+                   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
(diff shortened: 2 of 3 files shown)


vlajos (dismissed):

data/db-schema/src/main/resources/net/corda/db/schema/vnode-vault/migration/ledger-utxo-relevant-transaction-state-creation-v1.0.xml
```diff
@@ -0,0 +1,40 @@
...
+            <column name="transaction_id" type="VARCHAR(160)">
+                <constraints nullable="false"/>
+            </column>
+            <column name="group_idx" type="INT">
+                <constraints nullable="false"/>
+            </column>
+            <column name="leaf_idx" type="INT">
+                <constraints nullable="false"/>
+            </column>
+            <column name="consumed" type="BOOLEAN">
```
vlajos: I guess this field could be computed from the `consumed_timestamp`.
Although I am not too sure if it makes any real differences.

nkovacsx (dismissed):

data/db-schema/src/main/resources/net/corda/db/schema/vnode-vault/migration/ledger-utxo-relevant-transaction-state-creation-v1.0.xml
blsemo: Yes, I agree that this is redundant information. Having this and `consumed_timestamp` could lead to inconsistent data.

data/db-schema/src/main/resources/net/corda/db/schema/vnode-vault/migration/ledger-utxo-relevant-transaction-state-creation-v1.0.xml
```diff
@@ -0,0 +1,40 @@
...
+                <constraints nullable="false"/>
+            </column>
+            <column name="custom" type="JSONB">
```
blsemo: I am not thrilled by this name - could we have something like "custom_representation" or "custom_data" etc?

data/db-schema/src/main/resources/net/corda/db/schema/vnode-vault/migration/ledger-utxo-creation-v1.0.xml
```diff
@@ -217,4 +191,6 @@
                        constraintName="utxo_transaction_cpk_pkey"
                        tableName="utxo_transaction_cpk"/>
     </changeSet>
+
+    <include file="net/corda/db/schema/vnode-vault/migration/ledger-utxo-relevant-transaction-state-creation-v1.0.xml"/>
```
blsemo, line 195: Good idea to split this out into a separate file!

data/db-schema/src/main/resources/net/corda/db/schema/vnode-vault/migration/ledger-utxo-relevant-transaction-state-creation-v1.0.xml
lankydan: Renamed to `custom_representation`.

data/db-schema/src/main/resources/net/corda/db/schema/vnode-vault/migration/ledger-utxo-relevant-transaction-state-creation-v1.0.xml
lankydan: Removed the original `consumed` column and renamed `consumed_timestamp` to `consumed` to match the `created` column.

vlajos: Thanks.

vlajos (dismissed):

mnesbit (approved):
OK

data/db-schema/src/main/resources/net/corda/db/schema/vnode-vault/migration/ledger-utxo-creation-v1.0.xml
driessamyn: I don't think it makes technically a difference, but wouldn't it be more consistent to have this include in the `db.changelog-master.xml` file?
Where it is here may get a bit confusing when a new version is introduced.

data/db-schema/src/main/resources/net/corda/db/schema/vnode-vault/migration/ledger-utxo-relevant-transaction-state-creation-v1.0.xml
```diff
@@ -0,0 +1,37 @@
+<?xml version="1.1" encoding="UTF-8" standalone="no"?>
+<databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+                   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-4.3.xsd">
+
+    <changeSet author="R3.Corda" id="ledger-utxo-relevant-transaction-state-creation-postgres-v1.0" dbms="postgresql">
```
driessamyn, line 6: For reference, where we've had to deal with DB specific types elsewhere, we have used properties.

E.g.: https://github.com/corda/corda-api/blob/5169296760cc10fbdd07ba3fee2a68227dd6adb6/data/db-schema/src/main/resources/net/corda/db/schema/config/migration/config-creation-v1.0.xml#L6

This would let you do something like
```
<property name="json-type" value="VARCHAR(MAX)" dbms="hsqldb"/>
<property name="json-type" value="JSONP" dbms="postgresql"/>

...

<column name="custom_representation" type="${json-type}">
</column>

```

PR #948, opened 2023-03-15
CORE-11639: Make `SerializedBytes` an interface on public API

anien:
- Makes `SerializedBytes` an interface and decouples it on API level from `OpaqueBytes` and `ByteSequence`
- Adds `SerializedBytes.getBytes` (previously inherited from `OpaqueBytes`)

runtime-os PR: [#3376](https://github.com/corda/corda-runtime-os/pull/3376)


serialization/src/main/java/net/corda/v5/serialization/SerializedBytes.java
```diff
@@ -1,29 +1,22 @@
...
 
     /**
      * Returns a text description of this SerializedBytes object
      * @return Text description of this SerializedBytes object.
      */
     @NotNull
-    public String getSummary() {
-        return "SerializedBytes(size = " + getSize() + ')';
-    }
+    String getSummary();
```
kyriathar: Don't really see much useful text coming from this one, perhaps we should kill?

(Related to this one:) Perhaps we should be providing a more meaningful `toString` than the current `ByteSequence.toString` I think.  Maybe something like: "SerializedBytes(bytes = \<hex string bytes\>)";"

kyriathar: (This PR is meant to only convert it to interface, so maybe above comment could be done under different work/ PR)

anien: Let's do it now, it's not a big change and it would be better to do it all together.

anien: When thinking about it, I would expect `toString()` to only return the hex value of the bytes, not some other information/sentence. `getSummary()` seems more appropriate for me to return custom details.

kyriathar: I think let's just remove then, doesn't offer much, seems more like unneeded overhead to the API to me

anien: Done.

kyriathar (approved):
LGTM

driessamyn (approved):
LGTM

PR #949, opened 2023-03-15
CORE-11523: Boot parameters for REST worker

vkolomeyko:
This is a non-breaking change.

### DIFF
--- a/data/config-schema/src/main/java/net/corda/schema/configuration/BootConfig.java
+++ b/data/config-schema/src/main/java/net/corda/schema/configuration/BootConfig.java
@@ -5,6 +5,7 @@ import static net.corda.schema.configuration.MessagingConfig.MAX_ALLOWED_MSG_SIZ
 /**
  * Configuration paths for values used to bootstrap the worker
  */
+@SuppressWarnings("unused")
 public final class BootConfig {
     private BootConfig() {
     }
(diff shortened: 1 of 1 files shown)


data/config-schema/src/main/java/net/corda/schema/configuration/BootConfig.java
```diff
@@ -26,4 +27,9 @@ private BootConfig() {
     public static final String BOOT_DIR = "dir";
     public static final String BOOT_WORKSPACE_DIR = BOOT_DIR + ".workspace";
     public static final String BOOT_TMP_DIR = BOOT_DIR + ".tmp";
+
+    public static final String BOOT_REST = "rest";
+    public static final String BOOT_REST_PARAMS = BOOT_REST + ".params";
```
driessamyn, line 32: Does it need `params`?
It seems maybe a bit un-necessary. Maybe `rest.tls.keystore...` is good enough?

vkolomeyko: I am flexible either way. But I was trying to be consistent with DB, i.e.: `BOOT_DB_PARAMS` which is `db.params....`.

Either way command line argument will look like: `-rtls.keystore.file.path=...` so use of `params` is purely internal and not visible to the C5 operator using REST Worker images.

driessamyn: Ah. Yes. Got it 👍

driessamyn (approved):
LGTM

PR #950, opened 2023-03-15
CORE-9704 - Attach serial to registration requests

nikinagy:
Attach serial number to registration requests. Users should be able to specify their current info's serial in the registration context as well.


charlieR3 (dismissed):

data/membership-schema/src/main/resources/net/corda/schema/membership/member/dynamic/registration/1.0/corda.member.dynamic.registration.json
```diff
@@ -67,6 +67,11 @@
       "description": "Optional. A UUID string that uniquely identifies the pre-auth token.",
       "type": "string",
       "pattern": "^[a-fA-F0-9]{8}-[a-fA-F0-9]{4}-[1-5][a-fA-F0-9]{3}-[89aAbB][a-fA-F0-9]{3}-[a-fA-F0-9]{12}$"
+    },
+    "corda.serial": {
+      "description": "Optional. The serial on the member's latest member info.",
+      "type": "string",
+      "maxLength": "100"
```
dimosr: Is there a reason we can't use an integer here, instead?

Also, it would be good to not have a limit (or have an extremely large one) to support scenarios with a large number of re-registrations.

nikinagy: The thing is that the whole context is a map of <String, String> to begin with and we want to validate that by using this schema validation, which is a json validation so we need to convert the whole context to a json string by using the object mapper (see validateRegistrationContext() in MembershipSchemaValidatorImpl).

I can increase the limit if we want that. What could be a suitable value as limit?

charlieR3: If I understand right, `"maxLength": "100"` means it can be a number up to 100 digits since it's a string length, right? Not that the limit of the number value is `100`. So I think it's already extremely high.

nikinagy: Yes, it is in string length.

YashNabar: Should be possible to use integer here if we want I think, we're doing that for the group policy file format version

nikinagy: The group policy is not a map of string, string, it is actually a json file, hence it is a much easier case there.
We can add integer for the serial for sure (and define a min value as 0 and a max value), but then we need to convert back and forth, between the schema and the registration context.
But let me try it and dig into it, maybe I am overcomplicating it in my head.

dimosr: I see - ok, if that means 100 digits, we should be safe as-is.

dimosr (dismissed):
Fine by me - if you want to switch to integer, no objection from me, but if it's too much effort and you prefer to stick with string I think that's also ok for now

data/membership-schema/src/main/resources/net/corda/schema/membership/member/dynamic/registration/1.0/corda.member.dynamic.registration.json
```diff
@@ -67,6 +67,12 @@
       "description": "Optional. A UUID string that uniquely identifies the pre-auth token.",
       "type": "string",
       "pattern": "^[a-fA-F0-9]{8}-[a-fA-F0-9]{4}-[1-5][a-fA-F0-9]{3}-[89aAbB][a-fA-F0-9]{3}-[a-fA-F0-9]{12}$"
+    },
+    "corda.serial": {
+      "description": "Optional. The serial on the member's latest member info.",
+      "type": "integer",
+      "minimum": 0,
+      "maximum": 10000000
```
dimosr: Do we need to enforce an explicit max value if it's an integer now?

nikinagy: Good point. I'll remove it.

nikinagy: addressed in [d19ee48](https://github.com/corda/corda-api/pull/950/commits/d19ee48faa88719b573def61af1aa5a2bf3e393b)

dimosr (dismissed):

dimosr (approved):

PR #954, opened 2023-03-16
CORE-7994 Replace key with key id in digital signature packet 

kyriathar:
WIP


kyriathar:
Closing as [#956](https://github.com/corda/corda-api/pull/956) was used to address this work.

PR #956, opened 2023-03-16
CORE-7994 Replace key with key id in digital signature packet

vlajos:
At the moment, the `SigningService.sign` returned signature packet includes the signing key (public key) and the signature packet will be sent to over the wire to counter parties. This means that currently, the receiver of that signature packet can grab that key and use it to verify the signature. However, they should be checking that this key exists in the locally known/ trusted counter parties keys. By changing the key to key id it means the consumer of the signature has to to resolve the key out of the key id to proceed with signature verification. 

- replaces `DigitalSignature.WithKey` with `DigitalSignature.WithKeyId`
- adds signing key to `TransactionSignatureService.verifySignature` parameters, since the mapping from key to key id takes place in `TransactionSignatureService.verifySignature` callers sites

runtime-os PR: [#3400](https://github.com/corda/corda-runtime-os/pull/3400)


ledger/ledger-common/src/main/java/net/corda/v5/ledger/common/transaction/TransactionSignatureService.java
```diff
@@ -57,12 +58,21 @@ List<List<DigitalSignatureAndMetadata>> signBatch(
...
-            @NotNull final DigitalSignatureAndMetadata signatureWithMetadata
+            @NotNull final DigitalSignatureAndMetadata signatureWithMetadata,
+            @NotNull final PublicKey publicKey
+    );
+
+    // TODO Not too sure how good place is this for this function, but since Transactionsignature service is already available where this will be used
+    // in signed txs etc, probable good enough for now.
+    SecureHash getIdOfPublicKey(
+            @NotNull final PublicKey publicKey,
+            @NotNull final String digestAlgorithmName
```
mnesbit: should be nullable in which case platform default digest (as given by DigestService, but we seem to still be missing the ability to get any digest algorithm names anywhere!)

mnesbit: Speaking to @kyriathar he notes that the algorithm should match that of the id you are comparing to. So perhaps the method should be a matchesId(id: SecureHash) on PublicKey

ledger/ledger-common/src/main/java/net/corda/v5/ledger/common/transaction/TransactionSignatureService.java
```diff
@@ -57,12 +58,22 @@ List<List<DigitalSignatureAndMetadata>> signBatch(
...
      * @throws RuntimeException if the signature could not be verified.
      */
     void verifySignature(
             @NotNull final TransactionWithMetadata transaction,
-            @NotNull final DigitalSignatureAndMetadata signatureWithMetadata
+            @NotNull final DigitalSignatureAndMetadata signatureWithMetadata,
+            @NotNull final PublicKey publicKey
+    );
+
+    // TODO this needs internally in a few places, but maybe it could be kept internally? TransactionSignatureServiceInternal maybe?
```
kyriathar: So you have exposed this, because it needs to be also used outside `TransactionSignatureServiceImpl`. However, I think it feels like an internal utility.

ledger/ledger-common/src/main/java/net/corda/v5/ledger/common/transaction/TransactionSignatureService.java
```diff
@@ -57,12 +58,22 @@ List<List<DigitalSignatureAndMetadata>> signBatch(
...
+     * @param transaction           The original transaction.
      * @param signatureWithMetadata The signature to be verified.
+     * @param publicKey             The public key to verify against. It should match with signatureWithMetadata's keyId.
      * @throws RuntimeException if the signature could not be verified.
      */
     void verifySignature(
             @NotNull final TransactionWithMetadata transaction,
-            @NotNull final DigitalSignatureAndMetadata signatureWithMetadata
+            @NotNull final DigitalSignatureAndMetadata signatureWithMetadata,
+            @NotNull final PublicKey publicKey
```
kyriathar, line 68: Should we keep this API clean of key id, to only be aware of key from this point on? (Essentially keep it as before)

It should be more clean conceptually then? Although that means two distinct types for `DigitalSignatureAndMetadata` then.

If we for whatever reason want to add verify by key id, that should come with an overload then if the need arises?

(Maybe not for this PR though)

kyriathar: Although on second thought we need to have signature only with key id so not expose one with key.

ledger/ledger-common/src/main/java/net/corda/v5/ledger/common/transaction/TransactionSignatureService.java
vlajos: Yes. In the meantime I moved it into a `TransactionSignatureServiceInternal` interface in runtime-os.
The impl part stayed the same and that same impl, implements this new interface too.
I am not 100% sure how this plays with OSGi...
If it does not we can still keep it on the public API, on another internal service or somewhere else.

ledger/ledger-common/src/main/java/net/corda/v5/ledger/common/transaction/TransactionSignatureService.java
kyriathar: Should we rename `publicKey` to something more meaningful like `signatureKey`?

vlajos: I think I would keep `publicKey` since that has a well known meaning in signature verification context.

crypto/src/main/java/net/corda/v5/crypto/DigitalSignature.java
```diff
@@ -16,19 +14,19 @@ public DigitalSignature(@NotNull byte[] bytes) {
...
-    public static class WithKey extends DigitalSignature {
+    public static class WithKeyId extends DigitalSignature {
 
         /**
-         * Construct WithKey
+         * Creates a new {@code WithKeyId} using the specified key id and the signature ({@code bytes}).
          *
          * @param by      The public key of the corresponding private key used to sign the data (as if an instance
          *                of the {@link CompositeKey} is passed to the sign operation it may contain keys which are not actually owned by
          *                the member).
```
blsemo:
```suggestion
         * @param by      The id of the key pair used to sign the data (if this resolves to a
         *                {@link CompositeKey}, not all keys might be owned by the member).
```

This needs to be updated, as it doesn't take a key anymore. Also, the wording around composite keys is a bit confusing.

ledger/ledger-common/src/main/java/net/corda/v5/ledger/common/transaction/TransactionSignatureService.java
```diff
@@ -57,12 +57,14 @@ List<List<DigitalSignatureAndMetadata>> signBatch(
      * The underlying verification service signals the verification failures with different exceptions.
      * {@link DigitalSignatureVerificationService}
      *
-     * @param transaction The original transaction.
+     * @param transaction           The original transaction.
      * @param signatureWithMetadata The signature to be verified.
+     * @param publicKey             The public key to verify against. It should match with signatureWithMetadata's keyId.
```
blsemo: It should match or it must match?

blsemo (commented):
Looks good to me, but a few questions around JDocs.

crypto/src/main/java/net/corda/v5/crypto/DigitalSignature.java
kyriathar: I don't think that's very relevant to be here. Maybe more relevant to say that if composite key used to sign it will sign with the firstly found key leaf.

ledger/ledger-common/src/main/java/net/corda/v5/ledger/common/transaction/TransactionSignatureService.java
vlajos: I think _must_ is the correct.

mnesbit: must

kyriathar: Applied

mnesbit (dismissed):
OK

blsemo (approved):
LGTM

PR #960, opened 2023-03-17
CORE-11523: Property name change

vkolomeyko:
This is a non-breaking change as this property is not used anywhere yet.

Usage illustration in: https://github.com/corda/corda-runtime-os/pull/3377


driessamyn (approved):
LGTM

PR #962, opened 2023-03-17
CORE-11769 - simplify boot config

driessamyn:
- Remove params from config structure so that it becomes a cleaner structure to support in file based configuration.
- Add secrets boot config

Needed for: https://github.com/corda/corda-runtime-os/pull/3409


vkolomeyko (approved):
LGTM

PR #964, opened 2023-03-20
CORE-11786 Remove `SecureHash.getBytes`

kyriathar:
`SecureHash` API should be enough to users to just provide the hex string form of the hash.

- removes public API `SecureHash.getBytes`

runtime-os PR: [#3412](https://github.com/corda/corda-runtime-os/pull/3412)

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -7,7 +7,7 @@ cordaProductVersion = 5.0.0
 # NOTE: update this each time this module contains a breaking change
 ## NOTE: currently this is a top level revision, so all API versions will line up, but this could be moved to
 ##   a per module property in which case module versions can change independently.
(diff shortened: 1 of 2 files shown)


gradle.properties
```diff
@@ -7,7 +7,7 @@ cordaProductVersion = 5.0.0
 # NOTE: update this each time this module contains a breaking change
 ## NOTE: currently this is a top level revision, so all API versions will line up, but this could be moved to
 ##   a per module property in which case module versions can change independently.
-cordaApiRevision = 721
+cordaApiRevision = 722
```
driessamyn: This will need to be bumped again. Otherwise LGTM

driessamyn (approved):
LGTM

PR #970, opened 2023-03-20
CORE-11845: add columns to crypto_wrapping_key for key rotation

dickon:
We do not intend to implement key rotation at this time, but we have got far enough with the design that we can see making some schema changes now before we need to supply migrations will make implementation at a later date easier, and deploying releases that introduce those changes less disruptive.

Move out signing key material into a new table table crypto_signing_key_material,  so that it is possible to store the same private key wrapped by mutliple differnet wrapping keys. This allows us to support incremental key rotation.  To facilitate this, add a synthetic primary UUID key to the signing and wrapping key tables.

Add a rotation date column to wrapping keys. Currently, this can be populated to the far future, but if it is set to a an upcoming time would tirgger rotation. 

Include a reference to parent wrapping key from signing keys, a flag to indicate whether the wrapping key is stored in the crypto processor.

Rename timestamp in crypto tables to be created, which is more specific.

Rename cryptoLibrary in the top level of the configuration to be simply crypto instead, to improve consistency with other config node names.

### DIFF
--- a/data/db-schema/src/main/resources/net/corda/db/schema/crypto/migration/crypto-creation-v1.0.xml
+++ b/data/db-schema/src/main/resources/net/corda/db/schema/crypto/migration/crypto-creation-v1.0.xml
@@ -2,12 +2,27 @@
 <databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
                    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
                    xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-4.3.xsd">
+    <!--
+     For now, the schema must be identical between here and the 
+     virtual node schema (in ../../vnode-crypto/migration/crypto-creation-v1.0.xml)
+     since the crypto processor uses the same code irrespective of where the table resides.
+     
+     We could just include the same changeset files in both cases, but that might confuse readers
+     and liquibase, so instead we have to keep the content in sync.  
+     -->
     <changeSet author="R3.Corda" id="crypto-creation-v1.0">
         <createTable tableName="crypto_wrapping_key">
+            <column name="id" type="UUID">
+                <constraints nullable="false"/>>
+            </column>
             <column name="alias" type="VARCHAR(64)">
                 <constraints nullable="false"/>
             </column>
-            <column name="created" type="TIMESTAMP">
+            <!-- generation of the wrapping key; higher generation numbers to be used to replaced expiring wrapping keys -->
+            <column name="generation" type="INT">
+                <constraints nullable="false"/>
+            </column>
+            <column name="created" type="TIMESTAMP" defaultValueComputed="CURRENT_TIMESTAMP">
                 <constraints nullable="false"/>
             </column>
             <column name="encoding_version" type="INT">
@@ -19,11 +34,29 @@
             <column name="key_material" type="VARBINARY(1048576)">
                 <constraints nullable="false"/>
             </column>
+            <column name="rotation_date" type="TIMESTAMP">
+                <constraints nullable="false"/>
+            </column>
+            <!-- If true, the parent key is a wrapping key in this table, and 
+               parent_key_reference is the id of that wrapping key. If false, the parent_key
+               is a corda config path. -->  
+            <column name="is_parent_key_managed" type="BOOLEAN">
+                <constraints nullable="false"/>
+            </column>
+            <!-- the identity of the key used for key_material; may reference a root wrapping key outside the database, 
+                 so there is no foreign key constraint -->
+            <column name="parent_key_reference" type="VARCHAR(256)">
+                <constraints nullable="false"/>
+            </column>
         </createTable>
-        <addPrimaryKey  columnNames="alias"
+        <addPrimaryKey  columnNames="id"
                         tableName="crypto_wrapping_key"
                         constraintName="crypto_wrapping_key_pk"/>
-
+        <addUniqueConstraint columnNames="alias, generation"
+                             tableName="crypto_wrapping_key"
+                             constraintName="alias_and_generation_are_unique"/>
+        <!-- TODO FK constraint back to crypto_wrapping_key on parent_key_reference when is_parentkey_managed=true, if possible
+             (or remodel so we can) -->
         <createTable tableName="crypto_hsm_association">
             <column name="id" type="VARCHAR(36)">
                 <constraints nullable="false"/>
@@ -88,16 +121,21 @@
         and liquibase, so instead we have to keep the content in sync.  
         -->
         <createTable tableName="crypto_signing_key">
+            <column name="id" type="UUID">
+                <constraints nullable="false"/>
+            </column>
             <column name="tenant_id" type="VARCHAR(12)">
                 <constraints nullable="false"/>
             </column>
+            <!-- short hash (first 12 hex chars of SHA256) of the public key --> 
             <column name="key_id" type="CHAR(12)">
                 <constraints nullable="false"/>
             </column>
+            <!-- full SHA256hash of the public key -->
             <column name="full_key_id" type="VARCHAR(255)">
                 <constraints nullable="false"/>
             </column>
-            <column name="timestamp" type="TIMESTAMP">
+            <column name="created" type="TIMESTAMP" defaultValueComputed="CURRENT_TIMESTAMP">
                 <constraints nullable="false"/>
             </column>
             <column name="category" type="VARCHAR(64)">
@@ -109,13 +147,7 @@
             <column name="public_key" type="VARBINARY(1048576)">
                 <constraints nullable="false"/>
             </column>
-            <column name="key_material" type="VARBINARY(1048576)">
-                <constraints nullable="true"/>
-            </column>
             <column name="encoding_version" type="INT">
-                <constraints nullable="true"/>
-            </column>
-            <column name="master_key_alias" type="VARCHAR(64)">
                 <constraints nullable="true"/>
             </column>
             <column name="alias" type="VARCHAR(64)">
@@ -135,6 +167,9 @@
             </column>
         </createTable>
         <addPrimaryKey constraintName="crypto_signing_key_pk"
+                       tableName="crypto_signing_key"
+                       columnNames="id"/>
+        <addUniqueConstraint constraintName="crypto_signing_key_unique"
                        tableName="crypto_signing_key"
                        columnNames="tenant_id, key_id"/>
         <createIndex indexName="crypto_signing_key_full_key_idx"
@@ -156,5 +191,25 @@
             <column name="tenant_id"/>
             <column name="alias"/>
         </createIndex>
+        <createTable tableName="crypto_signing_key_material">
+            <column name="wrapping_key_id" type="UUID">
+                <constraints nullable="false"/>
+            </column>
+            <column name="signing_key_id" type="UUID">
+                <constraints nullable="false"/>
+            </column>
+            <column name="signing_key_material" type="VARBINARY(1048576)">
+                <constraints nullable="false"/>
+            </column>
+            <!-- when the key material was generated -->
+            <column name="created" type="TIMESTAMP" defaultValueComputed="CURRENT_TIMESTAMP">
+                <constraints nullable="false"/>
+            </column>
+        </createTable>
+        <addPrimaryKey constraintName="crypto_signing_key_material_pk"
+                       tableName="crypto_signing_key_material"
+                       columnNames="signing_key_id, wrapping_key_id"/>
+        <!-- TODO FK constraint to crypto_wrapping_key on (wrapping_key_alias, wrapping_key_generation) -->
+        <!-- TODO FK constraint to crypto_signing_key on (signing_key_alias, signing_key_generation) -->
     </changeSet>
 </databaseChangeLog>
--- a/data/db-schema/src/main/resources/net/corda/db/schema/vnode-crypto/migration/vnode-crypto-creation-v1.0.xml
+++ b/data/db-schema/src/main/resources/net/corda/db/schema/vnode-crypto/migration/vnode-crypto-creation-v1.0.xml
@@ -2,26 +2,140 @@
 <databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
                    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
                    xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-4.3.xsd">
+    <!--
+    For now, the crypto_signing_key table schema must be identical between here and the 
+    cluster schema (in ../../crypto/migration/crypto-creation-v1.0.xml)
+    since the crypto processor uses the same code irrespective of where the table resides.
+    
+    We could just include the same changeset files in both cases, but that might confuse readers
+    and liquibase, so instead we have to keep the content in sync.  
+    -->
     <changeSet author="R3.Corda" id="vnode-crypto-creation-v1.0">
+        <createTable tableName="crypto_wrapping_key">
+            <column name="id" type="UUID">
+                <constraints nullable="false"/>>
+            </column>
+            <column name="alias" type="VARCHAR(64)">
+                <constraints nullable="false"/>
+            </column>
+            <!-- generation of the wrapping key; higher generation numbers to be used to replaced expiring wrapping keys -->
+            <column name="generation" type="INT">
+                <constraints nullable="false"/>
+            </column>
+            <column name="created" type="TIMESTAMP" defaultValueComputed="CURRENT_TIMESTAMP">
+                <constraints nullable="false"/>
+            </column>
+            <column name="encoding_version" type="INT">
+                <constraints nullable="false"/>
+            </column>
+            <column name="algorithm_name" type="VARCHAR(64)">
+                <constraints nullable="false"/>
+            </column>
+            <column name="key_material" type="VARBINARY(1048576)">
+                <constraints nullable="false"/>
+            </column>
+            <column name="rotation_date" type="TIMESTAMP">
+                <constraints nullable="false"/>
+            </column>
+            <!-- If true, the parent key is a wrapping key in this table, and 
+               parent_key_reference is the id of that wrapping key. If false, the parent_key
+               is a corda config path. -->
+            <column name="is_parent_key_managed" type="BOOLEAN">
+                <constraints nullable="false"/>
+            </column>
+            <!-- the identity of the key used for key_material; may reference a root wrapping key outside the database, 
+                 so there is no foreign key constraint -->
+            <column name="parent_key_reference" type="VARCHAR(256)">
+                <constraints nullable="false"/>
+            </column>
+        </createTable>
+        <addPrimaryKey  columnNames="id"
+                        tableName="crypto_wrapping_key"
+                        constraintName="crypto_wrapping_key_pk"/>
+        <addUniqueConstraint columnNames="alias, generation"
+                             tableName="crypto_wrapping_key"
+                             constraintName="alias_and_generation_are_unique"/>
+        <!-- TODO FK constraint back to crypto_wrapping_key on parent_key_reference when is_parentkey_managed=true, if possible
+             (or remodel so we can) -->
+        <createTable tableName="crypto_hsm_association">
+            <column name="id" type="VARCHAR(36)">
+                <constraints nullable="false"/>
+            </column>
+            <column name="tenant_id" type="VARCHAR(12)">
+                <constraints nullable="false"/>
+            </column>
+            <column name="hsm_id" type="VARCHAR(36)">
+                <constraints nullable="false"/>
+            </column>
+            <column name="timestamp" type="TIMESTAMP">
+                <constraints nullable="false"/>
+            </column>
+            <column name="master_key_alias" type="VARCHAR(30)">
+                <constraints nullable="true"/>
+            </column>
+        </createTable>
+        <addPrimaryKey  columnNames="id"
+                        tableName="crypto_hsm_association"
+                        constraintName="crypto_hsm_association_pk"/>
+        <addUniqueConstraint columnNames="tenant_id, hsm_id"
+                             tableName="crypto_hsm_association"
+                             constraintName="crypto_hsm_association_uc"/>
+
+        <createTable tableName="crypto_hsm_category_association">
+            <column name="id" type="VARCHAR(36)">
+                <constraints nullable="false"/>
+            </column>
+            <column name="tenant_id" type="VARCHAR(12)">
+                <constraints nullable="false"/>
+            </column>
+            <column name="category" type="VARCHAR(20)">
+                <constraints nullable="false"/>
+            </column>
+            <column name="hsm_association_id" type="VARCHAR(36)">
+                <constraints nullable="false"/>
+            </column>
+            <column name="timestamp" type="TIMESTAMP">
+                <constraints nullable="false"/>
+            </column>
+            <column name="deprecated_at" type="BIGINT">
+                <constraints nullable="false"/>
+            </column>
+        </createTable>
+        <addPrimaryKey  columnNames="id"
+                        tableName="crypto_hsm_category_association"
+                        constraintName="crypto_hsm_category_association_pk"/>
+        <addUniqueConstraint columnNames="tenant_id, category, deprecated_at"
+                             tableName="crypto_hsm_category_association"
+                             constraintName="crypto_hsm_category_association_uc"/>
+        <addForeignKeyConstraint  baseColumnNames="hsm_association_id"
+                                  baseTableName="crypto_hsm_category_association"
+                                  referencedColumnNames="id"
+                                  referencedTableName="crypto_hsm_association"
+                                  constraintName="crypto_hsm_category_association_fk1"/>
         <!--
         For now, the crypto_signing_key table schema must be identical between here and the 
         virtual node schema (in ../../crypto/migration/crypto-creation-v1.0.xml)
         since the crypto processor uses the same code irrespective of where the table resides.
-        
+       
         We could just include the same changeset files in both cases, but that might confuse readers
         and liquibase, so instead we have to keep the content in sync.  
         -->
         <createTable tableName="crypto_signing_key">
+            <column name="id" type="UUID">
+                <constraints nullable="false"/>
+            </column>
             <column name="tenant_id" type="VARCHAR(12)">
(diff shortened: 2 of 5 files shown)


data/db-schema/src/main/resources/net/corda/db/schema/crypto/migration/crypto-creation-v1.0.xml
```diff
@@ -4,9 +4,17 @@
                    xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-4.3.xsd">
     <changeSet author="R3.Corda" id="crypto-creation-v1.0">
         <createTable tableName="crypto_wrapping_key">
+            <column name="id" type="VARCHAR(64)">
```
driessamyn:
```suggestion
            <column name="id" type="UUID">
```

data/db-schema/src/main/resources/net/corda/db/schema/crypto/migration/crypto-creation-v1.0.xml
```diff
@@ -4,9 +4,17 @@
...
+                <constraints nullable="false"/>
+            </column>
             <column name="alias" type="VARCHAR(64)">
                 <constraints nullable="false"/>
             </column>
+            <!-- generation of the wrapping key, as a UUID if created by us or a user supplied value for root wrapping keys
```
driessamyn: Maybe?
```suggestion
            <!-- generation of the wrapping key, as a INT with the highest value one considered the latest
```

data/db-schema/src/main/resources/net/corda/db/schema/crypto/migration/crypto-creation-v1.0.xml
```diff
@@ -4,9 +4,17 @@
...
         <createTable tableName="crypto_wrapping_key">
+            <column name="id" type="VARCHAR(64)">
+                <constraints nullable="false"/>
+            </column>
             <column name="alias" type="VARCHAR(64)">
                 <constraints nullable="false"/>
             </column>
+            <!-- generation of the wrapping key, as a UUID if created by us or a user supplied value for root wrapping keys 
+                 where we do not store the key material -->
+            <column name="generation" type="VARCHAR(64)">
```
driessamyn:
```suggestion
            <column name="generation" type="INT">
```

data/db-schema/src/main/resources/net/corda/db/schema/crypto/migration/crypto-creation-v1.0.xml
```diff
@@ -4,9 +4,17 @@
...
+                <constraints nullable="false"/>
+            </column>
             <column name="created" type="TIMESTAMP">
```
driessamyn:
```suggestion
            <column name="created" type="TIMESTAMP" defaultValueComputed="CURRENT_TIMESTAMP">
```

data/db-schema/src/main/resources/net/corda/db/schema/crypto/migration/crypto-creation-v1.0.xml
```diff
@@ -16,14 +24,34 @@
...
             <column name="key_material" type="VARBINARY(1048576)">
-                <constraints nullable="false"/>
+                <constraints nullable="true"/>
+            </column>
+            <!-- if null, there should be automated rotation -->
+            <column name="rotation_date" type="TIMESTAMP">
+                <constraints nullable="true"/>
+            </column>
+            <!-- FK to the other key that was used to encrypt this wrapping key.  -->
+            <column name="parent_key_id" type="VARCHAR(64)">
```
driessamyn:
```suggestion
            <column name="parent_key_id" type="UUID">
```

data/db-schema/src/main/resources/net/corda/db/schema/crypto/migration/crypto-creation-v1.0.xml
```diff
@@ -16,14 +24,34 @@
...
+                <constraints nullable="true"/>
             </column>
         </createTable>
-        <addPrimaryKey  columnNames="alias"
+        <addPrimaryKey  columnNames="alias, version"
```
driessamyn:
```suggestion
        <addPrimaryKey  columnNames="id"
```

data/db-schema/src/main/resources/net/corda/db/schema/crypto/migration/crypto-creation-v1.0.xml
```diff
@@ -16,14 +24,34 @@
...
                         tableName="crypto_wrapping_key"
                         constraintName="crypto_wrapping_key_pk"/>
-
+        <addUniqueConstraint columnNames="alias, generation"
+                             tableName="crypto_wrapping_key"
+                             constraintName="Alias and generation are unique"/>
+        <addForeignKeyConstraint baseColumnNames="parent_key_id"
+                                 baseTableName="crypto_wrapping_key"
+                                 onUpdate="RESTRICT"
+                                 onDelete="RESTRICT"
```
driessamyn: Are these needed?

driessamyn (commented):
Couple of minor things.

Not convinced by the nullable key material column.
I think maybe it would be simpler to have this as non-nullable and have the "complete" root keys in config rather than shadow them here.

But happy to go with this proposal and see the implementation.

data/db-schema/src/main/resources/net/corda/db/schema/crypto/migration/crypto-creation-v1.0.xml
dickon: change made

data/db-schema/src/main/resources/net/corda/db/schema/crypto/migration/crypto-creation-v1.0.xml
dickon: maybe not, le'ts try without. Deleted.

data/db-schema/src/main/resources/net/corda/db/schema/crypto/migration/crypto-creation-v1.0.xml
dickon: done

data/db-schema/src/main/resources/net/corda/db/schema/crypto/migration/crypto-creation-v1.0.xml
dickon: Liquibase docs are a bit limited, so I'm not sure what that would mean. My expectation is that some of the generation values are user supplied, for root wrapping keys, and that would mean the user configuration would have to be acceptable to the database.

https://xenovation.com/blog/development/java/java-professional-developer/liquibase-related-sql-java-types appears to show that UUID is mapped differently across databases.

data/db-schema/src/main/resources/net/corda/db/schema/crypto/migration/crypto-creation-v1.0.xml
dickon: well, going with INT does solve the problem of the exact type of a UUID, but it does expose us to various risks around collisions.

data/db-schema/src/main/resources/net/corda/db/schema/crypto/migration/crypto-creation-v1.0.xml
driessamyn: It's fine. That's how we do it elsewhere. There was an issue with this in old versions of Liquibase.

data/db-schema/src/main/resources/net/corda/db/schema/crypto/migration/crypto-creation-v1.0.xml
driessamyn: > well, going with INT does solve the problem of the exact type of a UUID, but it does expose us to various risks around collisions.

I don't think there is a risk of collision here. We're controlling what is inserted and when.
And it would be nice to be able to see which one is newest, so an INT makes sense.

dickon:
> Not convinced by the nullable key material column. I think maybe it would be simpler to have this as non-nullable and have the "complete" root keys in config rather than shadow them here.
> 
> But happy to go with this proposal and see the implementation.

The thing is that both @mnesbit and I would like to support, in the long run, arbitrary wrapping key hierarchies. This structure supports that, without much special casing in the code. Anyway I'll proceed and see how a POC looks.

driessamyn:
> > Not convinced by the nullable key material column. I think maybe it would be simpler to have this as non-nullable and have the "complete" root keys in config rather than shadow them here.
> > But happy to go with this proposal and see the implementation.
> 
> The thing is that both @mnesbit and I would like to support, in the long run, arbitrary wrapping key hierarchies. This structure supports that, without much special casing in the code. Anyway I'll proceed and see how a POC looks.

I appreciate that, but I'm always warry about implicit knowledge. I.e. because it's nullable has a special meaning here. 
Regardless of the hierarchy, you have to have a root somewhere, and I think that root cannot have its material in this table, so I think it should have nothing in this table.
Otherwise you'll end up with `if  ... == null` -> assume it's a root key sort of logic, followed by casting the thing to a non-nullable, which is just messy.
I'm not seeing what the benefit is about not being explicit about it. If anything I would rather then see a `keyFromConfigId` type of column that is also nullable then you can at least reason "if the key material is from config, grab it there, otherwise expect material here, which I appreciate would still leave this as a nullable column, but at least it would be explicite and allow any key to have it's material in config (possibly supporting an extension point for HSMs for certain wrapping keys)

data/config-schema/src/main/resources/net/corda/schema/configuration/cryptoLibrary/1.0/corda.cryptoLibrary.json
driessamyn, line 1: small thing, but can we take the opportunity to rename this file to `corda.cryto.json` to make it consistent with the others?

data/config-schema/src/main/resources/net/corda/schema/configuration/cryptoLibrary/1.0/corda.cryptoLibrary.json
```diff
@@ -5,6 +5,41 @@
   "description": "Configuration schema for the crypto library subsection.",
   "type": "object",
   "properties": {
+    "unmanagedWrappingKeys": {
+      "description" : "Key derivation parameters for unmanaged wrapping keys",
+      "type": "object",
```
driessamyn: Should this not be a list?

data/config-schema/src/main/resources/net/corda/schema/configuration/cryptoLibrary/1.0/corda.cryptoLibrary.json
```diff
@@ -5,6 +5,41 @@
   "description": "Configuration schema for the crypto library subsection.",
   "type": "object",
   "properties": {
+    "unmanagedWrappingKeys": {
```
driessamyn: Bit nitpicky, but is `unmanaged` appropriate here? 
I think I know you mean not managed by Corda, but it may sound wrong.
Maybe we should just call it `wrappingKeys` to keep it simple?

I know naming is difficult ;)

data/config-schema/src/main/resources/net/corda/schema/configuration/cryptoLibrary/1.0/corda.cryptoLibrary.json
```diff
@@ -5,6 +5,41 @@
...
+      "description" : "Key derivation parameters for unmanaged wrapping keys",
+      "type": "object",
+      "default": {},
+      "patternProperties": {
+        ".*": { 
+          "type" :  "object",
+          "properties": {
```
driessamyn: Do we need a type (with the default our only supported type `AES-...`)?
Maybe that's premature, on the other hand, maybe it helps being explicit? We can even have it as the only allowed value in the schema?

driessamyn: It this were to be a list, then we'd need an id property as well.

data/config-schema/src/main/resources/net/corda/schema/configuration/cryptoLibrary/1.0/corda.cryptoLibrary.json
```diff
@@ -5,6 +5,41 @@
...
+            },
+            "passphrase": {
+              "description": "Passphrase for the key derivation function",
+              "type": "string"
+            }
+          }
+        }
+      }
+    },
+    "defaultUnmanagedWrappingKey": {
```
driessamyn: I think maybe it would be clearer if this was called `defaultRootWrappingKeyId`.

And what' the thinking about the rotation when it's added? that we add a `previousRootWrappingKeyId`?

data/config-schema/src/main/resources/net/corda/schema/configuration/cryptoLibrary/1.0/corda.cryptoLibrary.json
```diff
@@ -5,6 +5,41 @@
...
+      "description": "The default unmanaged wrapping key, which must be in unmanagedWrappingKeys.",
+      "type": "string",
+      "default" : "unmanagedWrappingKey1"
```
driessamyn: I don't think a default makes sense here.

data/config-schema/src/main/resources/net/corda/schema/configuration/cryptoLibrary/1.0/corda.cryptoLibrary.json
```diff
@@ -5,6 +5,41 @@
...
+    "wrappingKeyRotationPeriod": {
+      "description": "The number of days before a managed wrapping key is considered to have expired.",
+      "type": "integer",
+      "default" : 365
+    },
+    "wrappingKeyRotationWindow": {
+      "description": "The number of days before a deadline when a managed wrapping key rotation should start.",
+      "type": "integer",
+      "default" : 180
+    },
```
driessamyn: I think we need to leave these out until we add key rotation (i.e. this should be on a separate branch).
We need some more design around this. I don't know having the 2 is valuable. We just need to know when to start rotating. I think Matthew was also keen that auto rotation takes into account the number of times a key have been used, so let's not settle the config schema for that.
At the moment, the only concern is that we can do a manual rotation of the root keys.

data/db-schema/src/main/resources/net/corda/db/schema/crypto/migration/crypto-creation-v1.0.xml
```diff
@@ -156,5 +179,27 @@
             <column name="tenant_id"/>
             <column name="alias"/>
         </createIndex>
+        <createTable tableName="crypto_signing_key_material">
+            <column name="wrapping_key_alias" type="VARCHAR(64)">
+                <constraints nullable="false"/>
+            </column>
+            <column name="wrapping_key_generation" type="VARCHAR(64)">
+                <constraints nullable="false"/>
+            </column>
```
driessamyn: Why not reference by `id` given that is the PK on the other table?

data/db-schema/src/main/resources/net/corda/db/schema/crypto/migration/crypto-creation-v1.0.xml
```diff
@@ -156,5 +179,27 @@
...
+            <column name="signing_key_id" type="CHAR(12)">
+                <constraints nullable="false"/>
+            </column>
+            <column name="signing_key_material" type="VARBINARY(1048576)">
+                <constraints nullable="false"/>
+            </column>
+        </createTable>
+        <addPrimaryKey constraintName="wrapping_key_alias, wrapping_key_generation, signing_key_tenant_id, signing_key_id"
+                       tableName="crypto_signing_key_material"
+                       columnNames="tenant_id, key_id"/>
```
driessamyn: This is a bit odd.
Why not make `signing_key_id` the PK and use a UUID type for that?

The mapping etc will be much simpler and even things like retrieving a key in queries.

data/db-schema/src/main/resources/net/corda/db/schema/crypto/migration/crypto-creation-v1.0.xml
```diff
@@ -156,5 +179,27 @@
...
+            <column name="wrapping_key_alias" type="VARCHAR(64)">
+                <constraints nullable="false"/>
+            </column>
+            <column name="wrapping_key_generation" type="VARCHAR(64)">
+                <constraints nullable="false"/>
+            </column>
+            <column name="signing_key_tenant_id" type="VARCHAR(12)">
+                <constraints nullable="false"/>
+            </column>
+            <column name="signing_key_id" type="CHAR(12)">
```
driessamyn: Why `CHAR(12)`? I would just make this UUID.

data/config-schema/src/main/resources/net/corda/schema/configuration/cryptoLibrary/1.0/corda.cryptoLibrary.json
dickon: Possibly this should be called "wrappingKeys" or "externalWrappingKeys". TBD more widely.

data/config-schema/src/main/resources/net/corda/schema/configuration/cryptoLibrary/1.0/corda.cryptoLibrary.json
dickon: Yes, let's switch to a list. I could have gone either way.

The JS object/map duality of confusing.

data/config-schema/src/main/resources/net/corda/schema/configuration/cryptoLibrary/1.0/corda.cryptoLibrary.json
dickon: Probably a symbolic string for key derivation function and key type. Sounds good.

dickon: :+1:

data/config-schema/src/main/resources/net/corda/schema/configuration/cryptoLibrary/1.0/corda.cryptoLibrary.json
dickon: agreed

data/config-schema/src/main/resources/net/corda/schema/configuration/cryptoLibrary/1.0/corda.cryptoLibrary.json
dickon: yes, config isn't needed for the POC.

data/db-schema/src/main/resources/net/corda/db/schema/crypto/migration/crypto-creation-v1.0.xml
dickon: Let's go to UUID.

data/db-schema/src/main/resources/net/corda/db/schema/crypto/migration/crypto-creation-v1.0.xml
dickon: Let's go to synthetic keys.

data/config-schema/src/main/resources/net/corda/schema/configuration/cryptoLibrary/1.0/corda.cryptoLibrary.json
dickon: Let's rename to `defaultRootWrappingKey`

data/db-schema/src/main/resources/net/corda/db/schema/crypto/migration/crypto-creation-v1.0.xml
dickon: This is out of date, we have a different approach now.

driessamyn (commented):
LGTM. Will wait for full approval till the runtime-os changes are there to review.

data/db-schema/src/main/resources/net/corda/db/schema/crypto/migration/crypto-creation-v1.0.xml
```diff
@@ -4,10 +4,20 @@
...
             </column>
-            <column name="created" type="TIMESTAMP">
+            <!-- generation of the wrapping key; higher generation numbers to be used to replaced expiring wrapping keys -->
+            <column name="generation" type="INT">
+                <constraints nullable="false"/>
+            </column>
+            <column name="created" type="TIMESTAMP" defaultValueComputed="CURRENT_TIMESTAMP">
+                <constraints nullable="false"/>
+            </column>
+            <column name="accessed" type="TIMESTAMP" defaultValueComputed="CURRENT_TIMESTAMP">
```
driessamyn: What's the accessed timestamp?
I don't think we should update a column on each read if that is what the intention was.

data/db-schema/src/main/resources/net/corda/db/schema/crypto/migration/crypto-creation-v1.0.xml
```diff
@@ -19,11 +29,30 @@
             <column name="key_material" type="VARBINARY(1048576)">
                 <constraints nullable="false"/>
             </column>
+            <!-- if specified, date when there should be automated rotation -->
+            <column name="rotation_date" type="TIMESTAMP">
+                <constraints nullable="true"/>
```
driessamyn: It may be simpler to keep this non-nullable and set the default to MAX in the future?

Not a big deal either way, but that would let you treat it as non-nullable in Kotlin.

data/db-schema/src/main/resources/net/corda/db/schema/crypto/migration/crypto-creation-v1.0.xml
```diff
@@ -88,16 +117,24 @@
...
             </column>
+            <!-- full SHA256hash of the public key -->
             <column name="full_key_id" type="VARCHAR(255)">
                 <constraints nullable="false"/>
             </column>
-            <column name="timestamp" type="TIMESTAMP">
+            <column name="created" type="TIMESTAMP" defaultValueComputed="CURRENT_TIMESTAMP">
+                <constraints nullable="false"/>
+            </column>
+            <column name="accessed" type="TIMESTAMP" defaultValueComputed="CURRENT_TIMESTAMP">
```
driessamyn: Similar to above. What is `accessed`?

data/db-schema/src/main/resources/net/corda/db/schema/crypto/migration/crypto-creation-v1.0.xml
```diff
@@ -134,7 +165,7 @@
                 <constraints nullable="false"/>
             </column>
         </createTable>
-        <addPrimaryKey constraintName="crypto_signing_key_pk"
+        <addUniqueConstraint constraintName="crypto_signing_key_pk"
```
driessamyn: This table still needs a pk (column `id`)

data/db-schema/src/main/resources/net/corda/db/schema/crypto/migration/crypto-creation-v1.0.xml
```diff
@@ -156,5 +187,30 @@
...
+            <column name="signing_key_material" type="VARBINARY(1048576)">
+                <constraints nullable="false"/>
+            </column>
+            <!-- when the key material was generated -->
+            <column name="created" type="TIMESTAMP" defaultValueComputed="CURRENT_TIMESTAMP">
+                <constraints nullable="false"/>
+            </column>
+
+            <!-- when the key material was generated -->
+            <column name="accessed" type="TIMESTAMP" defaultValueComputed="CURRENT_TIMESTAMP">
```
driessamyn: comment says `generated` but column says `accessed`. As above, what is the column for?

data/db-schema/src/main/resources/net/corda/db/schema/vnode-crypto/migration/vnode-crypto-creation-v1.0.xml
```diff
@@ -2,26 +2,147 @@
 <databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
                    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
                    xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-4.3.xsd">
-    <changeSet author="R3.Corda" id="vnode-crypto-creation-v1.0">
+    <changeSet author="R3.Corda" id="crypto-creation-v1.0">
```
driessamyn: Is there any reason `vnode` was removed from the ID? 
I'm guessing because it's a copy from the other on?

It may be clearer to have it on there to avoid confusion with the cluster one.

Otherwise, same comments as above.

driessamyn (changes requested):
Broadly looking good apart from the missing PK.

Let's chat about the `accessed` column.

data/db-schema/src/main/resources/net/corda/db/schema/crypto/migration/crypto-creation-v1.0.xml
dickon: removed

data/db-schema/src/main/resources/net/corda/db/schema/crypto/migration/crypto-creation-v1.0.xml
dickon: removed

data/db-schema/src/main/resources/net/corda/db/schema/crypto/migration/crypto-creation-v1.0.xml
dickon: Trying this.

data/db-schema/src/main/resources/net/corda/db/schema/crypto/migration/crypto-creation-v1.0.xml
dickon: removed

data/db-schema/src/main/resources/net/corda/db/schema/vnode-crypto/migration/vnode-crypto-creation-v1.0.xml
dickon: Yeah just a mistake syncing it up.

Having two separate copies is a pain, maybe I should have resisted this more

data/db-schema/src/main/resources/net/corda/db/schema/crypto/migration/crypto-creation-v1.0.xml
dickon: Done

driessamyn (commented):
LGTM

driessamyn (approved):
LGTM

PR #971, opened 2023-03-20
CORE-11412 Added virtual node status topic

owenstanford:
Adds support for async virtual node creation status tracking

`corda-runtime-os` PR: https://github.com/corda/corda-runtime-os/pull/3423

### DIFF
--- a/data/topic-schema/src/main/resources/net/corda/schema/VirtualNode.yaml
+++ b/data/topic-schema/src/main/resources/net/corda/schema/VirtualNode.yaml
@@ -86,8 +86,9 @@ topics:
   VirtualNodeOperationStatusTopic:
     name: virtual.node.operation.status
     consumers:
-      - db
+      - rest
     producers:
+      - db
       - rest
     config:
       cleanup.policy: compact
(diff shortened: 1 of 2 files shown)


data/topic-schema/src/main/resources/net/corda/schema/VirtualNode.yaml
```diff
@@ -82,4 +82,17 @@ topics:
...
       - rest
-    config:
\ No newline at end of file
+    config:
+  VirtualNodeOperationStatusTopic:
+    name: virtual.node.operation.status
+    consumers:
+      - db
+    producers:
+      - rest
```
davidcurrie: Are these not the wrong way round?

driessamyn: I think producer needs to be both `rest` and `db` and consumer `rest`

owenstanford: fixed

vkolomeyko (approved):
LGTM

PR #973, opened 2023-03-21
CORE-10422 Interop responder 

szymonsztuka:
New value for `flow initiator`, so any downstream processors can check if this was an Interop started flow.
Merging this tiny code change before the full functionality of the ticket is finises in `corda-runtime-os` greatly simplifies juggling with branches and builds.


ac101m (approved):
LGTM!

PR #974, opened 2023-03-21
CORE-11837: Remove sessionInitiationKeys from the member info

yift-r3:
Remove the sessionInitiationKeys from the member info public API. It should be used only by the membership and P2P layers. 

Runtime pull request is in https://github.com/corda/corda-runtime-os/pull/3492
E2E pull request in E2E pull request in https://github.com/corda/corda-e2e-tests/pull/52

### DIFF
--- a/membership/src/main/java/net/corda/v5/membership/MemberInfo.java
+++ b/membership/src/main/java/net/corda/v5/membership/MemberInfo.java
@@ -24,7 +24,6 @@ import java.util.List;
  * List<PublicKey> ledgerKeys = memberInfo.getLedgerKeys();
  * Long serial = memberInfo.getSerial();
  * int platformVersion = memberInfo.getPlatformVersion();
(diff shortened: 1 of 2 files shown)


membership/src/main/java/net/corda/v5/membership/MemberInfo.java
```diff
@@ -24,7 +24,6 @@
  * List<PublicKey> ledgerKeys = memberInfo.getLedgerKeys();
  * Long serial = memberInfo.getSerial();
  * int platformVersion = memberInfo.getPlatformVersion();
- * var sessionKeys = memberInfo.getSessionInitiationKeys();
```
charlieR3, line 27: There's a similar kdoc sample below: `val sessionKeys: PublicKey = memberInfo.sessionInitiationKeys`

charlieR3 (dismissed):

PR #977, opened 2023-03-21
CORE-9981 Added property keys accessible via flow context properties

pmck91:
please see linked pr in runtime https://github.com/corda/corda-runtime-os/pull/3397


data/avro-schema/src/main/java/net/corda/data/flow/CpiPropertyKeys.java
```diff
@@ -0,0 +1,10 @@
+package net.corda.data.flow;
```
sprintengine: The package sounds right but the module "data/avro-schema" doesn't sound right, sounds more like we should have a module inside "data" called "flow-schema" or something. Wdyt @mnesbit ? These are properties that will be used in flows to get from `flowEngine.getContextProperties(key)`

conal-smith (dismissed):

conal-smith (commented):
lgtm

conal-smith (dismissed):

fowlerrr (dismissed):
I'm happy with the platform version keys

mnesbit (approved):
LGTM

PR #979, opened 2023-03-21
CORE-5630 - Added the json schema for the external messaging

filipesoliveira:
Added the JSON schema for the external messaging.
Added tests.
Refactored the code in `src/main/java/net/corda/schema/configuration/` so it can be reused for the external messaging too.
Updated the API revision number.


vkolomeyko (dismissed):
LGTM

data/config-schema/src/main/java/net/corda/schema/common/provider/impl/SchemaProviderCommonImpl.java
```diff
@@ -0,0 +1,57 @@
+package net.corda.schema.common.provider.impl;
+
+import net.corda.schema.common.provider.SchemaProvider;
+import net.corda.v5.base.versioning.Version;
+import org.jetbrains.annotations.NotNull;
+import org.slf4j.Logger;
+
+import java.io.InputStream;
+
+public abstract class SchemaProviderCommonImpl implements SchemaProvider {
```
vkolomeyko:
```suggestion
public abstract class AbstractSchemaProvider implements SchemaProvider {
```

vkolomeyko (commented):
Given that we now have 2 classes extending from `SchemaProviderCommonImpl`, are we going to have the 3rd one for External Messages?

data/config-schema/src/main/resources/net/corda/schema/cordapp/configuration/external.messaging/1.0/corda.external.messaging.json
```diff
@@ -0,0 +1,41 @@
+{
```
vkolomeyko, line 1: Possibly, it would be good to keep this file out of this PR such that this PR will only be focused on refactoring for Config and CorDappConfig.

Then, PR to follow will do everything necessary to fit External Messages in.

filipesoliveira:
> Given that we now have 2 classes extending from `SchemaProviderCommonImpl`, are we going to have the 3rd one for External Messages?

The schema provider `SchemaProviderCordappConfigImpl` is intended to be used to fetch the schema for external messages. The class name is generic to allow people to add new schemas if they need to in the future.

data/config-schema/src/main/resources/net/corda/schema/cordapp/configuration/external.messaging/1.0/corda.external.messaging.json
filipesoliveira: Splitting this PR into two seems to require a lot more effort due to the process of merging changes to the api repository. I would suggest not to split this PR due to time constraints. I hope this is fine with you, @vkolomeyko.

vkolomeyko: Sure, let's leave it as one PR.

vkolomeyko (dismissed):
LGTM

vkolomeyko (dismissed):
LGTM

vkolomeyko (commented):
LGTM

vkolomeyko (dismissed):
LGTM

owenstanford (approved):

vkolomeyko (approved):
LGTM

PR #983, opened 2023-03-22
CORE-11905: REST Worker argument for certificates in PEM format

vkolomeyko:
This is a non-breaking `corda-api` change that adds few constants.


davidcurrie (approved):
👍

PR #984, opened 2023-03-22
CORE-9955: Add mutiple server definitions to gateway configuration

yift-r3:
Runtime pull request in https://github.com/corda/corda-runtime-os/pull/3466

Allow the gateway to listen to more than one `host:port` pair simultaneously.


data/config-schema/src/main/resources/net/corda/schema/configuration/p2p.gateway/1.0/corda.p2p.gateway.json
```diff
@@ -5,24 +5,36 @@
...
-    "urlPath": {
-      "description": "The url path the gateway server will be listening to.",
-      "type": "string",
-      "default": "/",
-      "pattern": "^\\/[\\/.a-zA-Z0-9-]*$"
+    "servers": {
+      "type": "array",
+      "minItems": 1,
+      "description": "The HTTP servers. The pair hostAddress and hostPort must be unique for each server.",
+      "default": [{}],
```
dimosr: Depends a bit on how the defaulting logic works for JSON schema.

But, should we instead have a default object in the array here that uses the defaults below (0.0.0.0, 8080, /) so that the 
setup does not require the user to specify config?

data/config-schema/src/main/resources/net/corda/schema/configuration/p2p.gateway/1.0/corda.p2p.gateway.json
```diff
@@ -5,24 +5,36 @@
...
+          "hostAddress": {
+            "description": "The host name or IP address where the HTTP server will bind.",
+            "type": "string",
+            "format": "hostname",
+            "default": "0.0.0.0"
+          },
+          "hostPort": {
+            "description": "The HTTP port.",
+            "type": "integer",
+            "default": 8080,
```
dimosr, line 30: Related to the comment above on defaulting. If we add the default object at the array level, should we remove the default at the type level for ports?

I am thinking of the scenario where someone provides multiple entries, then the default won't make a lot of sense since it will clash.

data/config-schema/src/main/resources/net/corda/schema/configuration/p2p.gateway/1.0/corda.p2p.gateway.json
williamvigorr3: Have you tested this, validates as expected if you specify different levels of schema (I had to put in defaults on both levels for the link manager but this might be different)?  You can adapt the `ConfigurationValidationIntegrationTest`, to check manually without deploying.

data/config-schema/src/main/resources/net/corda/schema/configuration/p2p.gateway/1.0/corda.p2p.gateway.json
```diff
@@ -5,24 +5,36 @@
...
-      "default": 8080,
-      "minimum": 0,
-      "maximum": 65535
-    },
-    "urlPath": {
-      "description": "The url path the gateway server will be listening to.",
-      "type": "string",
-      "default": "/",
-      "pattern": "^\\/[\\/.a-zA-Z0-9-]*$"
+    "servers": {
```
```suggestion
    "serversConfiguration": {
```

data/config-schema/src/main/resources/net/corda/schema/configuration/p2p.gateway/1.0/corda.p2p.gateway.json
yift-r3: I think that the default host and path make sense. I will remove the port to the list...

data/config-schema/src/main/resources/net/corda/schema/configuration/p2p.gateway/1.0/corda.p2p.gateway.json
```diff
@@ -5,24 +5,35 @@
...
-    "urlPath": {
-      "description": "The url path the gateway server will be listening to.",
-      "type": "string",
-      "default": "/",
-      "pattern": "^\\/[\\/.a-zA-Z0-9-]*$"
+    "serversConfiguration": {
+      "type": "array",
+      "minItems": 1,
+      "description": "The HTTP servers. The pair hostAddress and hostPort must be unique for each server.",
+      "default": [{"hostPort": 8080}],
```
dimosr: maybe it doesn't matter depending on how default takes effect (since if it's at least one entry, I would assume defaults for that type apply), but should we have all fields defined at the default at this level anyway? e.g.
```
"default": [{"hostAddress": "0.0.0.0", "hostPort": 8080, "urlPath": "/"}]
``` 
I think it would be a nice way to document the full default, instead of the reader having to jump to the various types to piece it together.

yift-r3: Annoyingly, if we put the default just in the array, it will not default any additional element in the group so I had to put them twice :(

dimosr (dismissed):

dimosr (approved):

PR #985, opened 2023-03-23
CORE-10385 Custom query execution (part 1)

nkovacsx:
### Overview

- Change `PagedQuery` to return a `ResultSet` instead of a `List<R>`, this will keep the APIs aligned (Persistence/Ledger-persistence)
- Modify `EntityResponse` avro schema to return similar properties as in `ResultSet` (`newOffset`, `hasNextPage`, `size`, `results`)
- Create `ExecuteVaultNamedQueryRequest` avro schema for custom query execution
- Add new `VaultNamedParameterizedQuery` to represent vault named queries 
- Fix `cordapp-configuration.properties` to align with the replaced `VaultNamedQueryFactory`


application/src/main/java/net/corda/v5/application/persistence/PagedQuery.java
```diff
@@ -49,5 +49,33 @@
...
+
+        /**
+         * @return The results contained in the result set.
+         */
+        List<T> getResults();
+
+        /**
+         * @return The size of the result set.
+         */
+        Integer getSize();
```
lankydan: Should be `int`? Same for the other ones as well.

application/src/main/java/net/corda/v5/application/persistence/PagedQuery.java
```diff
@@ -49,5 +49,33 @@
...
+
+        /**
+         * @return The new offset from where the result set continues or 0 if the result set is finished.
+         */
+        Integer getNewOffset();
```
lankydan: I think the new offset when the result set is finished should be the last record offset. In other words, it behaves as it normally would as if there was another page. At least this means that if they don't use `hasNextPage` for some reason, they wouldn't reprocess records they already went through.

data/avro-schema/src/main/resources/avro/net/corda/data/ledger/persistence/ExecuteVaultNamedQueryRequest.avsc
```diff
@@ -0,0 +1,31 @@
+{
+  "type": "record",
+  "name": "ExecuteVaultNamedQueryRequest",
+  "doc": "Retrieve and execute a named query with the given parameters.",
+  "namespace": "net.corda.data.ledger.persistence",
+  "fields": [
```
lankydan: Do we need this? Can `FindWithNamedQuery` be used instead?

application/src/main/java/net/corda/v5/application/persistence/PagedQuery.java
```diff
@@ -49,5 +49,33 @@
...
+
+        /**
+         * @return Whether the result set has a next page or not.
+         */
+        Boolean hasNextPage();
```
lankydan: `boolean`

application/src/main/java/net/corda/v5/application/persistence/PagedQuery.java
```diff
@@ -49,5 +49,33 @@
...
+     * A representation of a result set returned by a paged query when executed.
+     *
+     * @param <T> Type of the results contained in the result set.
+     */
+    interface ResultSet<T> {
+
+        /**
+         * @return The results contained in the result set.
+         */
+        List<T> getResults();
```
lankydan: Will need a `@NotNull`.

lankydan:
Surely this needs a version bump right?

application/src/main/java/net/corda/v5/application/persistence/PagedQuery.java
nkovacsx: I see, yeah that makes sense.

application/src/main/java/net/corda/v5/application/persistence/PagedQuery.java
nkovacsx: good point

application/src/main/java/net/corda/v5/application/persistence/PagedQuery.java
nkovacsx: good point

application/src/main/java/net/corda/v5/application/persistence/PagedQuery.java
nkovacsx: good point

data/avro-schema/src/main/resources/avro/net/corda/data/ledger/persistence/ExecuteVaultNamedQueryRequest.avsc
nkovacsx: hmm true, I think it should work

application/src/main/java/net/corda/v5/application/persistence/PagedQuery.java
```diff
@@ -49,5 +49,33 @@
      */
     @Suspendable
     @NotNull
-    List<R> execute();
+    ResultSet<R> execute();
```
driessamyn, line 52: The JavaDoc needs to be updated to reflect this change.

application/src/main/java/net/corda/v5/application/persistence/PagedQuery.java
```diff
@@ -49,5 +49,33 @@
...
+
+    /**
+     * A representation of a result set returned by a paged query when executed.
+     *
+     * @param <T> Type of the results contained in the result set.
+     */
+    interface ResultSet<T> {
```
driessamyn: What is the purpose of adding this intermediate interface?

And what is the expectation about when the query actually executes? I'm guessing still the `execute` method? Specifically, if you call `getResults` twice, does it just return the same thing twice or execute the query twice? 
I think this may just make the API more confusing.

application/src/main/java/net/corda/v5/application/persistence/PagedQuery.java
driessamyn: So does this just return the same offset that was passed in the `setOffset` method?

I'm struggling a bit to understand the value in this.
I think the existing API was modeled on the JPA API, which is well understood. I think it would be a shame to diverge.

application/src/main/java/net/corda/v5/application/persistence/PagedQuery.java
driessamyn: I'm really not sure about this. When you run a DB query, for example, you don't know if there is a "next page", until you run the next query.
Having this on the API would mean you always have to run 2 DB queries to find out. That would be inefficient, and also be different to otherwise well understood APIs.

It's also confusing to talk about pages when we don't in the main api. There it was a conscious decision to use limits and offsets, not pages, as that again aligns with the JPA API.

application/src/test/java/net/corda/v5/application/persistence/PersistenceServiceFindAllFlowJavaExample.java
```diff
@@ -45,7 +44,7 @@ public String call(@NotNull ClientRequestBody requestBody) {
                 .setOffset(200);
 
         // execute the query and return the results as a List
-        List<Dog> result1 = pagedQuery.execute();
+        PagedQuery.ResultSet<Dog> result1 = pagedQuery.execute();
```
driessamyn, line 47: comment are wrong.

This also shows how confusing the API is.
Does `result1` hold the result data or do you only get those once you call `getResults()` on it?

application/src/test/java/net/corda/v5/application/persistence/PersistenceServiceFindAllFlowJavaExample.java
```diff
@@ -55,7 +54,7 @@ public String call(@NotNull ClientRequestBody requestBody) {
                 .setOffset(200);
 
         // execute the query and return the results as a List
-        List<Dog> result2 = pagedQuery.execute();
+        PagedQuery.ResultSet<Dog> result2 = pagedQuery.execute();
```
driessamyn, line 57: As above, comment is wrong.

application/src/test/kotlin/net/corda/v5/application/persistence/PersistenceServiceFindAllFKotlinExample.kt
```diff
@@ -23,7 +24,7 @@ val pagedQuery = persistenceService
     .setOffset(200)
 
 // execute the query and return the results as a List
-val result1: List<Dog> = pagedQuery.execute()
+val result1: ResultSet<Dog> = pagedQuery.execute()
```
driessamyn: These examples should show the full API (i.e. how you would use Resultset)

data/avro-schema/src/main/resources/avro/net/corda/data/persistence/EntityResponse.avsc
```diff
@@ -11,6 +11,21 @@
...
+    },
+    {
+      "name": "newOffset",
+      "doc": "The new offset the result set continues from.",
+      "type": "int"
+    },
+    {
+      "name": "hasNextPage",
+      "doc": "A flag whether the result set has a next page or not.",
+      "type": "boolean"
```
driessamyn: Similar comments to above.
newOffset is confusing, and reference to a "page" too.

application/src/main/java/net/corda/v5/application/persistence/PagedQuery.java
nkovacsx: As far as I understand, this will return the offset from which the result set continues.

For example, if you do a query with `offset=0`, `limit=100` then the `newOffset` will be 100. That means to get the next "page" of the result set you need to set your offset to `offset=100`. Obviously you can only do that if there's a next "page" (i.e. result set still has elements)

@lankydan correct me if I'm wrong though.

driessamyn: But this is information you know, right? It's whichever offset you provided plus the limit.

application/src/main/java/net/corda/v5/application/persistence/PagedQuery.java
```diff
@@ -43,11 +43,25 @@
     /**
      * Executes the {@link PagedQuery}
      *
-     * @return List of entities found. Empty list if none were found.
+     * @return A {@link ResultSet} with the entities found.
```
lankydan: `A {@link ResultSet} containing the results of the query.` It won't hold entities for the results of the vault custom querying.

ledger/ledger-utxo/src/main/java/net/corda/v5/ledger/utxo/UtxoLedgerService.java
```diff
@@ -183,4 +184,53 @@ void sendUpdatedTransactionBuilder(
...
+     *     .setParameter("contractStateType", ContractState::class.java.name)
+     *     .setParameter("in-memory-filter-parameter", "parameter")
+     *     .setCreatedTimestampLimit(Instant.now())
+     *     .setOffset(0)
+     *     .setLimit(100)
+     * do {
+     *     val resultSet = query.execute()
+     *     processResultsWithApplicationLogic(resultSet.results)
+     *     query.setOffset(resultSet.newOffset)
+     * } while (resultSet.hasNextPage)
```
lankydan: We need to update the code docs due to the changes in the API.

```
do {
     *     val resultSet = query.execute()
     *     processResultsWithApplicationLogic(resultSet.results)
     *     query.setOffset(query.offset + 100)
     * } while (resultSet.results.isNotEmpty())
     *
```

Or something like that.

application/src/main/java/net/corda/v5/application/persistence/PagedQuery.java
nkovacsx: I'm not sure if I follow, do you mean the results will be transformed or ?

application/src/test/java/net/corda/v5/application/persistence/PagedQueryJavaApiTest.java
blsemo, line 1: These tests are completely pointless after moving the API to JAVA - can we delete it, please?

application/src/main/java/net/corda/v5/application/persistence/PagedQuery.java
blsemo: This needs a very good description what the result set will contain:
- what are the contents (e.g. transformed states/TransactionStates/StateRefs?)
- when is the query executed (i.e. is the query executed and returns a `ResultSet` with a data snapshot, or will the query be run when you call `getResults()`?


It is not returning Entities because that is a DB/JDBC concern that we don't deal with on the Corda API.

application/src/test/java/net/corda/v5/application/persistence/ParameterizedQueryJavaApiTest.java
blsemo, line 1: Also pointless, delete please.

application/src/test/java/net/corda/v5/application/persistence/PersistenceServiceJavaApiTest.java
blsemo, line 1: Can we get rid of this one as well?

ledger/ledger-utxo/src/main/java/net/corda/v5/ledger/utxo/UtxoLedgerService.java
```diff
@@ -183,4 +184,53 @@ void sendUpdatedTransactionBuilder(
...
+     *     .setParameter("participants", listOf("something"))
+     *     .setParameter("contractStateType", ContractState::class.java.name)
+     *     .setParameter("in-memory-filter-parameter", "parameter")
+     *     .setCreatedTimestampLimit(Instant.now())
+     *     .setOffset(0)
+     *     .setLimit(100)
+     * do {
+     *     val resultSet = query.execute()
+     *     processResultsWithApplicationLogic(resultSet.results)
+     *     query.setOffset(query.offset + 100)
```
blsemo, line 208: Is that `offset + 100` business correct? I seem to remember a discussion about new offset being returned? Just checking that we don't accidentally send mixed messages...

nkovacsx: If you have a look at the discussion above you can see that the new offset was confusing to @driessamyn so we (@lankydan) decided to remove that

application/src/test/java/net/corda/v5/application/persistence/PagedQueryJavaApiTest.java
nkovacsx: done

application/src/test/java/net/corda/v5/application/persistence/ParameterizedQueryJavaApiTest.java
nkovacsx: done

application/src/test/java/net/corda/v5/application/persistence/PersistenceServiceJavaApiTest.java
nkovacsx: done

application/src/main/java/net/corda/v5/application/persistence/PagedQuery.java
lankydan: Its a general interface so the documentation has to be general as well.

lankydan: I think it should look like this:

`PagedQuery`:

```java
/**
 * Used to build a Query that supports limit and offset.
 *
 * @param <R> The type of the results.
 */
public interface PagedQuery<R> {

    /**
     * Sets the maximum number of results to return.
     * <p>
     * If no limit is set, all records will be returned.
     *
     * @param limit The maximum number of results to return.
     *
     * @return The same {@link PagedQuery} instance.
     *
     * @throws IllegalArgumentException If {@code limit} is negative.
     */
    @NotNull
    PagedQuery<R> setLimit(int limit);

    /**
     * Sets the index of the first result in the query to return.
     * <p>
     * A default of `0` will be used if it is not set.
     *
     * @param offset The index of the first result in the query to return.
     *
     * @return The same {@link PagedQuery} instance.
     *
     * @throws IllegalArgumentException If {@code offset} is negative.
     */
    @NotNull
    PagedQuery<R> setOffset(int offset);

    /**
     * Executes the {@link PagedQuery}.
     *
     * @return A {@link ResultSet} containing the results of executing the query.
     *
     * @throws CordaPersistenceException If there is an error executing the query.
     */
    @Suspendable
    @NotNull
    ResultSet<R> execute();

    /**
     * A result set containing the results of calling {@link PagedQuery#execute()}.
     *
     * @param <R> The type of the results contained in the result set.
     */
    interface ResultSet<R> {

        /**
         * Extracts the results of a {@link ResultSet} from a previously executed query.
         * <p>
         * This method does not execute a query itself, call {@link PagedQuery#execute()} to execute a query and
         * generate a {@link ResultSet}.
         *
         * @return The results contained in the result set.
         */
        @NotNull
        List<R> getResults();
    }
}
```

`VaultNamedParameterizedQuery`:

```java
/**
 * Representation of a vault named query.
 * </p>
 * Ensure that {@link R} matches the type returned from the {@link VaultNamedQueryTransformer} and/or
 * {@link VaultNamedQueryCollector} (both optional) for the vault named query this object represents.
 * </p>
 * If no {@link VaultNamedQueryTransformer} or {@link VaultNamedQueryCollector} is applied to the vault named query,
 * the {@link ResultSet} from executing this query will contain {@link StateAndRef}s.
 *
 * @param <R> The type of results this query will fetch.
 *
 * @see UtxoLedgerService For thorough documentation of how to create and execute a vault named query.
 */
public interface VaultNamedParameterizedQuery<R> extends ParameterizedQuery<R> {

    /**
     * Sets the timestamp limit for the query. This will influence which results are returned.
     * @param timestampLimit The timestamp limit the query should enforce.
     * @return A {@link VaultNamedParameterizedQuery} object with the timestamp limit set.
     */
    @NotNull
    VaultNamedParameterizedQuery<R> setCreatedTimestampLimit(@NotNull Instant timestampLimit);

    /**
     * Executes the {@link VaultNamedParameterizedQuery} and creates a {@link PagedQuery.ResultSet} contains the results
     * of the query.
     * <p>
     * The results of the query depends on the executed named query and the {@link VaultNamedQueryTransformer} and/or
     * {@link VaultNamedQueryCollector} (both optional) applied to it.
     * </p>
     * If no {@link VaultNamedQueryTransformer} or {@link VaultNamedQueryCollector} is applied to the vault named query,
     * the {@link ResultSet} will contain {@link StateAndRef}s.
     *
     * @return A {@link ResultSet} containing the results of executing the query.
     *
     * @throws CordaPersistenceException If there is an error executing the query.
     */
    @Suspendable
    @NotNull
    @Override
    ResultSet<R> execute();
}
```

ledger/ledger-utxo/src/main/java/net/corda/v5/ledger/utxo/UtxoLedgerService.java
lankydan: Yeah, me and @driessamyn agreed to remove the `newOffset` concept and instead to introduce a `ResultSet.getNext()` in the future when we complete the optimisation on the db worker side to re-execute queries if the page size is not filled up. This would then understand how to manage the correct offset internally.

ledger/ledger-utxo/src/main/java/net/corda/v5/ledger/utxo/UtxoLedgerService.java
```diff
@@ -185,4 +186,53 @@ void sendUpdatedTransactionBuilder(
             @NotNull UtxoTransactionBuilder transactionBuilder,
             @NotNull FlowSession session
     );
+
+
+    /**
+     * Creates a query object for a named ledger query with the given name - if it exists. This query can be executed later.
```
lankydan: Remove the `if it exists` because it implies that it does nothing if it doesn't exist, instead it will error later on and throw in the user code.

ledger/ledger-utxo/src/main/java/net/corda/v5/ledger/utxo/query/VaultNamedQueryTransformer.java
```diff
@@ -28,10 +27,10 @@
  * }
  * }</pre></li></ul>
  *
- * @param <T> Type of the state returned from the database.
- * @param <R> Type that the original state is transformed into.
+ * @param <T> Type of the data returned from the database.
```
lankydan, line 31: Update the Javadocs for this and the filter, just changing the `state` in the argument to `data` is enough for me now. Everything else should still hold.

ledger/ledger-utxo/src/main/java/net/corda/v5/ledger/utxo/query/registration/VaultNamedQueryBuilder.java
```diff
@@ -28,11 +33,13 @@ public interface VaultNamedQueryBuilder extends VaultNamedQueryBuilderBase {
 
     /**
      * Sets the mapper function of the named query.
-     * @param mapper A transformer object.
+     * Please note that mapping function will always be applied after filtering.
```
lankydan: "Note that the transformation will always be applied after filtering."

blsemo (dismissed):
Looks good to me

lankydan (dismissed):

blsemo (dismissed):
LGTM

lankydan (approved):

PR #988, opened 2023-03-23
CORE-11895 Added support for message headers

owenstanford:
Adding support for headers to the messaging API, this is required to support messaging to external applications.


filipesoliveira (approved):

data/db-schema/src/main/resources/net/corda/db/schema/messagebus/migration/db-message-bus-tables-v1.0.xml
```diff
@@ -13,6 +13,9 @@
             <column name="record_value" type="VARBINARY(1048576)">
                 <constraints nullable="true"/>
             </column>
+            <column name="record_headers" type="TEXT">
```
vkolomeyko, line 16: Since there may be multiple headers, it probably would be worth putting a comment how they are going to be represented. E.g.:
`HeaderKey1=Value1,HeaderKey2=Value2,...`

PR #992, opened 2023-03-24
CORE-10575 Add a membership actions processor to the member worker

williamvigorr3:
Added new topic for membership actions. Changed the avro schema to move, distributing the member info from the registration processor to the membership actions processor.


data/avro-schema/src/main/resources/avro/net/corda/data/membership/actions.request/DistributeMemberInfo.avsc
```diff
@@ -0,0 +1,28 @@
...
+    },
+    {
+      "name": "minimumGroupParametersEpoch",
+      "doc": "The minimum group parameter epoch to be distribute (if null we use the latest version). If the group parameters with this epoch hasn't been published, then the membership actions processor will requeue this request to retry later.",
+      "type": ["null", "int"]
+    },
+    {
+      "name": "minimumUpdatedMemberSerial",
+      "doc": "The serial number of the minimum version of the updated members' memberInfo to distribute (if null we use the latest version). If the member info with this serial hasn't been published, then the membership actions processor will requeue this request to retry later.",
+      "type": ["null", "long"]
```
nikinagy, line 25: I don't really see how or in which case this can be null. As far a Is understand we get this epoch from the member info, where serial is non-nullable. Or is this nullability part of some future plans which I missed?
Actually, same goes for epoch. Are we sure there is a case when it can be null? Because I think we are throwing an exception in case it ends up being null.

williamvigorr3: The idea is that you set it too null if you don't care which version you distribute (but there has to be a version to distribute). This probably is only useful if you updated the member info and not the group parameters.

nikinagy: As I can see in ApproveRegistrationHandler we will still use the current epoch, so it technically won't be null. This is not a problem I think, but I feel like we are doing null checks on a field which can only be null in theory, but actually never be.

williamvigorr3: Actually what I mean is: This probably is only useful if you updated the group parameters and not the member info.

williamvigorr3: I think we can set the `minimumGroupParametersEpoch` to null in the ApproveRegistrationHandler, if the member if not a notary.

nikinagy: If it's a non-notary it will just get the latest epoch from the latest group params via the group reader (and throw an exception if it's null - can't find the group policy)

nikinagy: Discussed offline and I'm happy to go with nullables.

nikinagy (dismissed):
LGTM

dimosr (approved):

PR #998, opened 2023-03-27
CORE-10049 - Added secrets bootstrap config

simon-johnson-r3:
More bootstrap Config is required to integrate with HashiCorp Vault. These are for CORE-10049 and also CORE-9674.

Optional extra params which can be passed at bootstrapping:
- SECRET_REFRESH_PERIOD: A refresh period for secrets from Vault
- SECRET_RETRY_BACKOFF: A backoff period for automatic retries when Vault is not reachable

Keys and properties which define what a secret in the Config looks like. These were previously hard coded at multiple points of use, or defined outside of the api, which was incorrect as they are public facing.
- SECRET_KEY: The parent name of a secret in the Config
- SECRET_KEY_VAULT_PATH: The name of the property under SECRET_KEY which defines the path of the secret to locate it in the remote source
- SECRET_KEY_VAULT_KEY: The name of the property under SECRET_KEY which defines the key of the secret in the remote source at the specified SECRET_KEY_VAULT_PATH

This is a non-breaking change, only new values are declared.

### DIFF
--- a/data/config-schema/src/main/java/net/corda/schema/configuration/ConfigKeys.java
+++ b/data/config-schema/src/main/java/net/corda/schema/configuration/ConfigKeys.java
@@ -42,7 +42,7 @@ public final class ConfigKeys {
     //
     // For instance:
     //   - EncryptionSecretsService in corda-runtime-os needs SECRETS_PASSPHRASE and SECRETS_SALT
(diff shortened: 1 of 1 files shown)


driessamyn (approved):
LGTM

dickon (approved):
LGTM

PR #1003, opened 2023-03-28
CORE-9662: Make `SignatureSpec` an interface on public API

anien:
* makes `SignatureSpec` an interface
* internalizes `ParameterizedSignatureSpec` (ports it to runtime-os repo)

runtime-os PR: [#3494](https://github.com/corda/corda-runtime-os/pull/3494)


crypto/src/main/java/net/corda/v5/crypto/SignatureSpec.java
```diff
@@ -2,195 +2,22 @@
...
-     */
-    public @NotNull String toString() {
-        return this.signatureName;
-    }
+@DoNotImplement
+public interface SignatureSpec {
 
     /**
      * Alternative access getter for the signature name for completeness; same result as [toString].
      *
```
blsemo: This comment seems weird now  - this is the only function on the interface, so it can't be _alternative_. This probably needs a bit of JDoc how to use it?

anien: Updated. PTAL.

PR #1005, opened 2023-03-28
CORE-11255: Add initial platform version to FlowState

fowlerrr:
Adding this will allow flows to react appropriately when the platform is upgraded across a suspend.


driessamyn (dismissed):
This change looks good, but wait merge till the runtime-os is ready to be merged too.

driessamyn (commented):
LGTM (but you'll need to get in the API lock queue)

driessamyn (approved):
LGTM

PR #1010, opened 2023-03-28
CORE-10424 : Changes for managing session for interop processor

SaurabhR3:
1. Changes for managing session for interop processor
2. Added Interop State

### DIFF
--- a/data/avro-schema/src/main/resources/avro/net/corda/data/interop/InteropState.avsc
+++ b/data/avro-schema/src/main/resources/avro/net/corda/data/interop/InteropState.avsc
@@ -32,6 +32,11 @@
       "name": "aliasHoldingIdentity",
       "type": "string",
       "doc": "The alias holding id associated with ongoing sessions."
+    },
+    {
+      "name": "groupId",
+      "type": "string",
+      "doc": "The group id for alias holding id associated with ongoing sessions."
     }
   ]
 }
--- a/data/avro-schema/src/main/resources/avro/net/corda/data/interop/InteropState.avsc
+++ b/data/avro-schema/src/main/resources/avro/net/corda/data/interop/InteropState.avsc
@@ -1,0 +1,37 @@
+{
+  "type": "record",
+  "name": "InteropState",
+  "namespace": "net.corda.data.interop",
+  "fields": [
+    {
+      "name": "stateId",
+      "type": "string",
+      "doc": "The key on which inbound events should be forwarded to on the interop event topic."
+    },
+    {
+      "name": "expiryTime",
+      "type": [
+        "null",
+        "long"
+      ],
+      "doc": "The time at which this state will be set to null. This is only set for closed sessions."
+    },
+    {
+      "name": "status",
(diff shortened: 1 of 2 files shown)


yarivus (approved):
LGTM

ac101m (commented):
Best bump the api revision to 3 for this!
Bear in mind that the api and runtime respositories are currently out of sync, if you have pending changes in runtime-os as well, you'll need to wait for CORE-10420 to be merged before merging them (blame szymon!).

jmacmahonr3 (approved):
LGTM

yarivus (approved):
LGTM

data/avro-schema/src/main/resources/avro/net/corda/data/interop/InteropState.avsc
```diff
@@ -0,0 +1,37 @@
...
+        "type": "enum",
+        "name": "InteropStateType",
+        "symbols": [
+          "VALID",
+          "INVALID"
+        ]
+      }
+    },
+    {
+      "name": "aliasHoldingIdentity",
```
szymonsztuka, line 32: I think the best would be to store groupId as well, in a as separate field. StateProcessor would be completely independent from a message coming back from `FlowProcessor` in terms of deriving alias groupId (as it would store it previously) and no need to get groupId from the destination field.

SaurabhR3: I think we don't need that right now. We can easily derive it from destination alias in case of outbound message. In future if we need this we can add it.

ac101m (changes requested):
LGTM, but as my PR is still in progress, you should use version 2 rather than 3 as originally discussed.

data/avro-schema/src/main/resources/avro/net/corda/data/interop/InteropState.avsc
szymonsztuka: I would not expect that the destination alias will have a correct alias group name (of the "returning" messaged from initated flow) , we haven't really finished implementation inside FlowProcessor. Also saving alias group ID here, allows us to store the full context , not a half of the context. It's better for Session management to be fully independent, and we can also detect any incompatibilities (e.g. bug in FlowProcessor) when a message comes back (crucial in early development stage). Also from admin point of view, imagine at later stage investigating a queue to troubleshooting etc - we can se any Identity names but we have no clue what groups they belong for.

szymonsztuka (changes requested):

data/avro-schema/src/main/resources/avro/net/corda/data/interop/InteropState.avsc
SaurabhR3: Fixed in 755153e

ac101m (approved):
LGTM

szymonsztuka (approved):

yarivus (approved):
LGTM

yarivus (approved):
LGTM

yarivus (approved):
LGTM

PR #1013, opened 2023-03-29
CORE-11976 - Use session context properties for interop session identification

ac101m:
# Changes
- Add `isInteropSession` field to flow mapper state avro schema (with default value of false).
- Add `isInteropSession` field to session state avro schema (with default value of false).

# Notes
- Corresponding corda-runtime-os PR: https://github.com/corda/corda-runtime-os/pull/3497


yarivus (approved):
LGTM

gradle.properties
```diff
@@ -10,7 +10,7 @@ cordaProductVersion = 5.0.0-INTEROP
 ## IMPORTANT: 
 ## The interop feature branches track api revisions separately to the mainline branch.
 ## API version of last merge from corda mainline: 731
-cordaApiRevision = 2
+cordaApiRevision = 3
```
SaurabhR3: Need to bump up the version to 5, I just increased to to 4 for synching up with release branch.

SaurabhR3 (changes requested):

gradle.properties
ac101m: The version will be incremented when I resolve conflicts before merging and once the other changes in the PR are approved, as per the process detailed here: https://r3-cev.atlassian.net/wiki/spaces/CB/pages/4281270286/TL+DR+Rules+for+merging+breaking+API+changes+in+Corda+5
I'll just post in ledger ops rather than the main version locking thread.

jmacmahonr3 (approved):
Looks grand, but the sooner we document any potential redesigns the better

ac101m:
> Looks grand, but the sooner we document any potential redesigns the better

I think the plan for now is to stick with the session based mechanism for the time being as it's mostly complete and a redesign would still involve a few unknowns. If (as I suspect) this turns out to be necessary, szymon and I will definitely start by putting together a document.

yarivus (commented):
thx @ac101m

PR #1015, opened 2023-03-29
CORE-12143 Internalize `OpaqueBytes` and `ByteSequence`

kyriathar:
- removes `OpaqueBytes`, `ByteSequence` and `OpaqueBytesSubSequence` from public API

runtime-os PR: [#3506](https://github.com/corda/corda-runtime-os/pull/3506)


driessamyn (dismissed):
LGTM

base/src/main/java/net/corda/v5/base/types/ByteArrays.java
blsemo, line 1: Is this file still useful after this change? The `requireNotNull` function is not byte array specific, and can probably be deleted if we remove the concrete classes from the API?

kyriathar: Thanks @blsemo good catch! I will remove it and inline, it looks ugly. Also another thing is that now `ByteArrays` only (public) APIs are `toHexString` and `parseAsHex` which suggests file/ class could be renamed to something more specific to those and also it should be moved from `net.corda.v5.base.types` to `net.corda.v5.base.util` I think.

mnesbit: I think leave it. Purpose of file is clear

kyriathar: Removed `requireNotNull`.

PR #1026, opened 2023-03-30
CORE-3819: Configure default Gradle behaviour explicitly.

chrisr3:
Later versions of Gradle require org.gradle.java.installations.auto-download to be set explicitly.


ronanbrowne (approved):
LGTM

PR #1032, opened 2023-04-03
CORE-12255 Crypto API refinements

kyriathar:
- moves `ByteArrays` under utils
- changes functions in `KeyUtils` and `CompositeKey.isFulfilledBy` to accept `Set<PublicKey>` instead of `Iterable<PublicKey>` to improve their complexity

runtime-os PR: [#3541](https://github.com/corda/corda-runtime-os/pull/3541)

### DIFF
--- a/crypto/src/main/java/net/corda/v5/crypto/KeyUtils.java
+++ b/crypto/src/main/java/net/corda/v5/crypto/KeyUtils.java
@@ -42,7 +42,7 @@ public final class KeyUtils {
      * @param otherKeys An {@link Iterable} sequence of {@link PublicKey}.
      * @return True if <code>key</code> is in otherKeys.
      */
-    public static boolean isKeyInSet(@NotNull PublicKey key, @NotNull Iterable<PublicKey> otherKeys) {
+    public static boolean isKeyInKeys(@NotNull PublicKey key, @NotNull Iterable<PublicKey> otherKeys) {
         if (key instanceof CompositeKey) {
             CompositeKey compositeKey = (CompositeKey) key;
             Set<PublicKey> leafKeys = compositeKey.getLeafKeys();
@@ -78,7 +78,7 @@ public final class KeyUtils {
      * @param firstKey  The key with the requirements.
      * @param otherKeys The key to check whether requirements are fulfilled.
      */
-    public static boolean isKeyFulfilledBy(@NotNull PublicKey firstKey, @NotNull Iterable<PublicKey> otherKeys) {
+    public static boolean isKeyFulfilledByKeys(@NotNull PublicKey firstKey, @NotNull Iterable<PublicKey> otherKeys) {
         if (firstKey instanceof CompositeKey) {
             CompositeKey firstKeyComposite = (CompositeKey) firstKey;
             return firstKeyComposite.isFulfilledBy(otherKeys);
@@ -100,8 +100,8 @@ public final class KeyUtils {
      * @param firstKey The key with the requirements.
      * @param otherKey The key to check whether requirements are fulfilled.
      */
-    public static boolean isKeyFulfilledBy(@NotNull PublicKey firstKey, @NotNull PublicKey otherKey) {
-        return isKeyFulfilledBy(firstKey,
+    public static boolean isKeyFulfilledByKeys(@NotNull PublicKey firstKey, @NotNull PublicKey otherKey) {
+        return isKeyFulfilledByKeys(firstKey,
                 Collections.singleton(otherKey));
     }
 }
--- a/crypto/src/main/java/net/corda/v5/crypto/KeyUtils.java
+++ b/crypto/src/main/java/net/corda/v5/crypto/KeyUtils.java
@@ -42,7 +42,7 @@ public final class KeyUtils {
      * @param otherKeys An {@link Iterable} sequence of {@link PublicKey}.
      * @return True if <code>key</code> is in otherKeys.
      */
-    public static boolean isKeyInKeys(@NotNull PublicKey key, @NotNull Set<PublicKey> otherKeys) {
+    public static boolean isKeyInSet(@NotNull PublicKey key, @NotNull Set<PublicKey> otherKeys) {
         if (key instanceof CompositeKey) {
             CompositeKey compositeKey = (CompositeKey) key;
             Set<PublicKey> leafKeys = compositeKey.getLeafKeys();
@@ -77,7 +77,7 @@ public final class KeyUtils {
      * @param key  The key with the requirements.
      * @param otherKeys The key to check whether requirements are fulfilled.
      */
-    public static boolean isKeyFulfilledByKeys(@NotNull PublicKey key, @NotNull Set<PublicKey> otherKeys) {
+    public static boolean isKeyFulfilledBy(@NotNull PublicKey key, @NotNull Set<PublicKey> otherKeys) {
         if (key instanceof CompositeKey) {
             CompositeKey firstKeyComposite = (CompositeKey) key;
             return firstKeyComposite.isFulfilledBy(otherKeys);
@@ -96,8 +96,8 @@ public final class KeyUtils {
      * @param key The key with the requirements.
      * @param otherKey The key to check whether requirements are fulfilled.
      */
-    public static boolean isKeyFulfilledByKeys(@NotNull PublicKey key, @NotNull PublicKey otherKey) {
-        return isKeyFulfilledByKeys(key,
+    public static boolean isKeyFulfilledBy(@NotNull PublicKey key, @NotNull PublicKey otherKey) {
+        return isKeyFulfilledBy(key,
                 Collections.singleton(otherKey));
     }
 }
--- a/crypto/src/main/java/net/corda/v5/crypto/KeyUtils.java
+++ b/crypto/src/main/java/net/corda/v5/crypto/KeyUtils.java
@@ -46,14 +46,11 @@ public final class KeyUtils {
         if (key instanceof CompositeKey) {
             CompositeKey compositeKey = (CompositeKey) key;
             Set<PublicKey> leafKeys = compositeKey.getLeafKeys();
-            // TODO Here we could be checking which is set has less items and iterate through that
-            for (PublicKey otherKey : otherKeys) {
-                if (leafKeys.contains(otherKey)) return true;
-            }
+            leafKeys.retainAll(otherKeys);
+            return !leafKeys.isEmpty();
         } else {
             return otherKeys.contains(key);
         }
-        return false;
     }
 
     /**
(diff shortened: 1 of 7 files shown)


crypto/src/main/java/net/corda/v5/crypto/KeyUtils.java
```diff
@@ -42,7 +42,7 @@ private KeyUtils() {
      * @param otherKeys An {@link Iterable} sequence of {@link PublicKey}.
      * @return True if <code>key</code> is in otherKeys.
      */
-    public static boolean isKeyInSet(@NotNull PublicKey key, @NotNull Iterable<PublicKey> otherKeys) {
+    public static boolean isKeyInKeys(@NotNull PublicKey key, @NotNull Iterable<PublicKey> otherKeys) {
```
lankydan: Would `containsKey` be a better name?

crypto/src/main/java/net/corda/v5/crypto/KeyUtils.java
```diff
@@ -78,7 +78,7 @@ public static boolean isKeyInSet(@NotNull PublicKey key, @NotNull Iterable<Publi
      * @param firstKey  The key with the requirements.
      * @param otherKeys The key to check whether requirements are fulfilled.
      */
-    public static boolean isKeyFulfilledBy(@NotNull PublicKey firstKey, @NotNull Iterable<PublicKey> otherKeys) {
+    public static boolean isKeyFulfilledByKeys(@NotNull PublicKey firstKey, @NotNull Iterable<PublicKey> otherKeys) {
```
lankydan: I think the existing names of these methods are fine imo

crypto/src/main/java/net/corda/v5/crypto/SecureHash.java
```diff
@@ -38,5 +38,5 @@ public interface SecureHash {
      * Example outcome of toString(): SHA-256:98AF8725385586B41FEFF205B4E05A000823F78B5F8F5C02439CE8F67A781D90
      */
     @NotNull
-    String toString();
+    String getAlgorithmAndHexString();
```
mnesbit: This and its inverse parse are our standard naming/logging convention so don't change this.

mnesbit:
I am not convinced we need most of this. Maybe the move, but not sure the renames make much difference to clarity. Also note the higher priority for some of these is likely to be the scaling issues with taking Iterable rather then Set, though that can be fixed by adding an extra overload.

crypto/src/main/java/net/corda/v5/crypto/KeyUtils.java
```diff
@@ -42,17 +42,16 @@ private KeyUtils() {
      * @param otherKeys An {@link Iterable} sequence of {@link PublicKey}.
      * @return True if <code>key</code> is in otherKeys.
      */
-    public static boolean isKeyInSet(@NotNull PublicKey key, @NotNull Iterable<PublicKey> otherKeys) {
+    public static boolean isKeyInKeys(@NotNull PublicKey key, @NotNull Set<PublicKey> otherKeys) {
```
blsemo: I don't like this name at all, it sounds confusing to me.

crypto/src/main/java/net/corda/v5/crypto/KeyUtils.java
```diff
@@ -75,18 +74,15 @@ public static boolean isKeyInSet(@NotNull PublicKey key, @NotNull Iterable<Publi
...
      * polymorphic).
      * 
-     * @param firstKey  The key with the requirements.
+     * @param key  The key with the requirements.
      * @param otherKeys The key to check whether requirements are fulfilled.
      */
-    public static boolean isKeyFulfilledBy(@NotNull PublicKey firstKey, @NotNull Iterable<PublicKey> otherKeys) {
-        if (firstKey instanceof CompositeKey) {
-            CompositeKey firstKeyComposite = (CompositeKey) firstKey;
+    public static boolean isKeyFulfilledByKeys(@NotNull PublicKey key, @NotNull Set<PublicKey> otherKeys) {
```
blsemo: Similar, the second "Keys" doesn't add anything to the function name.

crypto/src/main/java/net/corda/v5/crypto/KeyUtils.java
```diff
@@ -97,11 +93,11 @@ public static boolean isKeyFulfilledBy(@NotNull PublicKey firstKey, @NotNull Ite
      * Since we do not define composite keys as acceptable on the second argument of this function, this relation
      * is not reflexive, not symmetric and not transitive. 
      *
-     * @param firstKey The key with the requirements.
+     * @param key The key with the requirements.
      * @param otherKey The key to check whether requirements are fulfilled.
      */
-    public static boolean isKeyFulfilledBy(@NotNull PublicKey firstKey, @NotNull PublicKey otherKey) {
-        return isKeyFulfilledBy(firstKey,
+    public static boolean isKeyFulfilledByKeys(@NotNull PublicKey key, @NotNull PublicKey otherKey) {
```
blsemo: This name is now wrong - because we only pass in one key to check fulfillment.

blsemo (commented):
I am ok with the moving and the implementations changes, but I don't think the renaming of the functions added much here, I'd rather not do that bit.
The real question I was asking myself when reading this code review is whether we actually need all these functions on the public API? Are they used in API classes? Or are they really utils that should either be shipped in a separate utils lib, or be part of runtime-os if CorDapp writers don't need them?

crypto/src/main/java/net/corda/v5/crypto/KeyUtils.java
driessamyn: I agree. I think the original name was actually better. I'd be tempted to call the second argument `set` then `isKeyInSet(key, set)` makes sense.

kyriathar: I believe "Set" does not reflect the business side of the second parameter. It's too generic.

vlajos:
> I am ok with the moving and the implementations changes, but I don't think the renaming of the functions added much here, I'd rather not do that bit. The real question I was asking myself when reading this code review is whether we actually need all these functions on the public API? Are they used in API classes? Or are they really utils that should either be shipped in a separate utils lib, or be part of runtime-os if CorDapp writers don't need them?

AFAIK atomic swap contracts are usually implemented with composite keys.
So I suspect the interop team will need at least the one with the threshold checks.
@jmacmahonr3: do I remember right?

crypto/src/main/java/net/corda/v5/crypto/KeyUtils.java
driessamyn: But you can see from the signature that is is a set of `PublicKey`.

But thinking about it now. Does the function provide anything over `set.Contains(key)` where set is a list of `PublicKey`?

jmacmahonr3:
> > I am ok with the moving and the implementations changes, but I don't think the renaming of the functions added much here, I'd rather not do that bit. The real question I was asking myself when reading this code review is whether we actually need all these functions on the public API? Are they used in API classes? Or are they really utils that should either be shipped in a separate utils lib, or be part of runtime-os if CorDapp writers don't need them?
> 
> AFAIK atomic swap contracts are usually implemented with composite keys. So I suspect the interop team will need at least the one with the threshold checks. @jmacmahonr3: do I remember right?

Hey @vlajos 

Thanks for the spot. Support for composite keys are pretty important for atomic swaps to the point we've a dependency on Ledger to add even more ease of use around collecting signatures for states down the line. It's very important that at a contract level we can assert that the asset is either held under a composite key or being released to either parties making up a composite key.

crypto/src/main/java/net/corda/v5/crypto/KeyUtils.java
kyriathar: It also takes care the case where `key` is `CompositeKey`, i.e. checks its leaves then

crypto/src/main/java/net/corda/v5/crypto/KeyUtils.java
kyriathar: Reverted name change

crypto/src/main/java/net/corda/v5/crypto/KeyUtils.java
kyriathar: Reverted name change

crypto/src/main/java/net/corda/v5/crypto/KeyUtils.java
kyriathar: Reverted name change

crypto/src/main/java/net/corda/v5/crypto/KeyUtils.java
kyriathar: I think if it was to be an extension function on `Collection<PublicKey>` then yes.

driessamyn (dismissed):
LGTM

crypto/src/main/java/net/corda/v5/crypto/KeyUtils.java
```diff
@@ -42,17 +42,16 @@ private KeyUtils() {
      * @param otherKeys An {@link Iterable} sequence of {@link PublicKey}.
      * @return True if <code>key</code> is in otherKeys.
      */
-    public static boolean isKeyInSet(@NotNull PublicKey key, @NotNull Iterable<PublicKey> otherKeys) {
+    public static boolean isKeyInSet(@NotNull PublicKey key, @NotNull Set<PublicKey> otherKeys) {
         if (key instanceof CompositeKey) {
             CompositeKey compositeKey = (CompositeKey) key;
             Set<PublicKey> leafKeys = compositeKey.getLeafKeys();
+            // TODO Here we could be checking which is set has less items and iterate through that
             for (PublicKey otherKey : otherKeys) {
```
mnesbit: Should perhaps use https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/Set.html#retainAll(java.util.Collection) to do set intersection and if non-empty return true

crypto/src/main/java/net/corda/v5/crypto/KeyUtils.java
```diff
@@ -75,18 +74,15 @@ public static boolean isKeyInSet(@NotNull PublicKey key, @NotNull Iterable<Publi
      * check fulfilment against a set of keys, without having to handle simple and composite keys separately (that is, this is
      * polymorphic).
      * 
-     * @param firstKey  The key with the requirements.
+     * @param key  The key with the requirements.
      * @param otherKeys The key to check whether requirements are fulfilled.
      */
-    public static boolean isKeyFulfilledBy(@NotNull PublicKey firstKey, @NotNull Iterable<PublicKey> otherKeys) {
```
mnesbit, line 81: I would expect to see https://github.com/corda/corda-api/blob/kyriakos/api-refinements/crypto/src/main/java/net/corda/v5/crypto/CompositeKey.java#L49 signature changed as well

kyriathar: Good catch, will apply

crypto/src/main/java/net/corda/v5/crypto/KeyUtils.java
kyriathar: Done

crypto/src/main/java/net/corda/v5/crypto/KeyUtils.java
kyriathar: Done

mnesbit (approved):
OK

PR #1033, opened 2023-04-03
CORE-11066 Create Json Factory interface and CorDapp property

nkovacsx:
### Overview
Create the main interface for contract state JSON factories. This interface will be implemented by CorDapps and one default internal factory for `ContractState`.

Added a new CorDapp property called `Corda-Ledger-Vault-Json-Factory-Classes` that will pick up the implementations from CorDapps.


cordapp-configuration/src/main/resources/net/corda/cordapp/cordapp-configuration.properties
```diff
@@ -14,6 +14,7 @@ Corda-InitiatedFlow-Classes=IMPLEMENTS;net.corda.v5.application.flows.ResponderF
 Corda-Subflow-Classes=IMPLEMENTS;net.corda.v5.application.flows.Subflow
 Corda-Token-Observer-Classes=IMPLEMENTS;net.corda.v5.ledger.utxo.observer.UtxoLedgerTokenStateObserver
 Corda-Ledger-Named-Query-Classes=IMPLEMENTS;net.corda.v5.ledger.utxo.query.VaultNamedQueryFactory
+Corda-Ledger-Vault-Json-Factory-Classes=IMPLEMENTS;net.corda.v5.ledger.utxo.query.json.ContractStateVaultJsonFactory
```
lankydan, line 17: I'm not 100% sure that `ContractStateVaultJsonFactory` is the right name, but I also can't figure out a better name atm, lets get this merged and just give it a think and if anything better comes up we can always change it in the new few weeks.

ledger/ledger-utxo/src/main/java/net/corda/v5/ledger/utxo/query/json/ContractStateVaultJsonFactory.java
```diff
@@ -0,0 +1,67 @@
...
+ * class TestUtxoStateVaultJsonFactory: ContractStateVaultJsonFactory<TestUtxoState> {
+ *
+ *     override val stateType: Class<TestUtxoState> = TestUtxoState::class.java
+ *
+ *     override fun append(state: TestUtxoState, jsonMarshallingService: JsonMarshallingService): String {
+ *         return jsonMarshallingService.format(TestUtxoStatePojo(state.testField))
+ *     }
+ * }
+ *
+ * data class TestUtxoStatePojo(val testField: String)
```
lankydan: Can you make the `data class` the same as the java example, e.g. a private class in the same position in the file? Just so it makes the examples more equivalent even though they both fine as standalone examples.

ledger/ledger-utxo/src/main/java/net/corda/v5/ledger/utxo/query/json/ContractStateVaultJsonFactory.java
```diff
@@ -0,0 +1,67 @@
+package net.corda.v5.ledger.utxo.query.json;
+
+import net.corda.v5.application.marshalling.JsonMarshallingService;
+import net.corda.v5.ledger.utxo.ContractState;
+import org.jetbrains.annotations.NotNull;
+
+/**
+ * The interface that needs to be implemented in order to create a JSON representation of a state type. Implementing
```
lankydan: Should mention that the JSON stored here is to be used by the vault named queries. We have to make sure that the link between this factory and the named queries is 100% clear.

lankydan: Maybe go with something like:

```
Implement a {@link ContractStateVaultJsonFactory} to create a JSON representation of a state to store alongside the state in the vault. This JSON representation can be used to query states from flows using {@link UtxoLedgerService#query}.
```

Maybe add a `@see` for the query factory and the ledger service query method.

ledger/ledger-utxo/src/main/java/net/corda/v5/ledger/utxo/query/json/ContractStateVaultJsonFactory.java
```diff
@@ -0,0 +1,67 @@
...
+    /**
+     * The function that defines how the given state can be represented as a JSON string.
+     *
+     * @param state The state object.
+     * @param jsonMarshallingService An instance of a {@link JsonMarshallingService} that can be used when creating a
+     * JSON representation.
+     *
+     * @return The JSON representation as a String.
+     */
+    @NotNull String append(@NotNull T state, @NotNull JsonMarshallingService jsonMarshallingService);
```
lankydan: I originally went with `create` for this naming, and I still think I prefer it. What do you think?

ledger/ledger-utxo/src/main/java/net/corda/v5/ledger/utxo/query/json/ContractStateVaultJsonFactory.java
```diff
@@ -0,0 +1,67 @@
...
+ */
+public interface ContractStateVaultJsonFactory<T extends ContractState> {
+
+    /**
+     * @return The type of the state this factory belongs to.
+     */
+    @NotNull Class<T> getStateType();
+
+    /**
+     * The function that defines how the given state can be represented as a JSON string.
```
lankydan, line 74: We should make it clear here or in the interface documentation (maybe do it in the interface to keep the longer explanations together), that we execute these factories against every state type in a state's type hierarchy and that users don't need to pull properties from super types (unless they want to) as they'll be stored in the JSON created by that super types factory (if it exists).

We should also specify that the json output will be keyed by the state type specified by the factory.

Another also, tell them that they can't register multiple factories for a single type.

ledger/ledger-utxo/src/main/java/net/corda/v5/ledger/utxo/query/json/ContractStateVaultJsonFactory.java
nkovacsx: Yeah that sounds better to me. Append might be a bit confusing.

gradle.properties
```diff
@@ -10,7 +10,7 @@ cordaProductVersion = 5.0.0
 # NOTE: update this each time this module contains a breaking change
 ## NOTE: currently this is a top level revision, so all API versions will line up, but this could be moved to
 ##   a per module property in which case module versions can change independently.
-cordaApiRevision = 742
+cordaApiRevision = 743
```
lankydan: I don't think this should need a version bump.

lankydan (dismissed):
Other than the question around the version bump, this looks good.

ledger/ledger-utxo/src/main/java/net/corda/v5/ledger/utxo/query/json/ContractStateVaultJsonFactory.java
```diff
@@ -0,0 +1,80 @@
...
+/**
+ * Implement a {@link ContractStateVaultJsonFactory} to create a JSON representation of a state to store alongside
+ * the state in the vault. This JSON representation can be used to query states from flows using
+ * {@link UtxoLedgerService#query}.
+ * <p>
+ * Classes that implement {@link ContractStateVaultJsonFactory} will be executed in a hierarchical way. That means
+ * every factory that belongs to a given state or that given state's ancestors will be executed and the end result
+ * will be a combined JSON. The combined JSON will be keyed by the state type specified by the factory.
+ * <p>
+ * Please note that only one factory can be registered for a state type.
```
lankydan, line 20: After the part where we talk about the keys, can we also state, that if a "query by state" type is needed for a particular state type, then there MUST be a factory registered, even if it returns an empty string/json object.

lankydan: Try word it nicer than I did

lankydan:
```
In order to perform a "query by state" query using `UtxoLedgerService.query` for a particular state type, a factory for that given type must be present, even if it's just returning an empty JSON string.
```

Also add a new line before this sentence.

lankydan (dismissed):

lankydan (approved):

PR #1036, opened 2023-04-04
NOTICK : Synch with release

SaurabhR3:


szymonsztuka:
The ticket CORE-10426 is not about merging from main.

gradle.properties
```diff
@@ -2,12 +2,15 @@
...
 # Versioning
 cordaProductVersion = 5.0.0-INTEROP
 # NOTE: update this each time this module contains a breaking change
 ## NOTE: currently this is a top level revision, so all API versions will line up, but this could be moved to
 ##   a per module property in which case module versions can change independently.
-## IMPORTANT: 
+## IMPORTANT:
 ## The interop feature branches track api revisions separately to the mainline branch.
 ## API version of last merge from corda mainline: 731
 cordaApiRevision = 2
```
ac101m: Bump our API version ~~to 3~~ by 1, also update the comment to contain the mainline API version that is being merged.

ac101m (changes requested):
Looks good to me aside from the version number.

SaurabhR3:
> The ticket CORE-10426 is not about merging from main.

Yeah I just added it to keep the PR title check happy. We need to add NOTICK as PR title check.
I believe for now it does not affect much.

gradle.properties
SaurabhR3: Do we really need it? we can use the 2 right?

ac101m: Nope. The purpose of the version number is to save us from unwittingly trying to build against incompatible dependencies. If there are breaking changes, it's got to be incremented!

P.S. Just remember to bump the version in runtime-os to match when you do that merge later.

szymonsztuka: Following Alex comment, as in the meantime new commit was merge (https://github.com/corda/corda-api/commit/5c106353dcaac7c0c5969130a8611e655ea0efe2 updating `cordaApiRevision = 3`), so now it should be change to `cordaApiRevision = 4`

szymonsztuka (changes requested):
Please change the ticket number, CORE-10426 is not about mere from main

gradle.properties
SaurabhR3: Fixed

gradle.properties
```diff
@@ -2,15 +2,18 @@
...
+
 # Versioning
 cordaProductVersion = 5.0.0-INTEROP
 # NOTE: update this each time this module contains a breaking change
 ## NOTE: currently this is a top level revision, so all API versions will line up, but this could be moved to
 ##   a per module property in which case module versions can change independently.
-## IMPORTANT: 
+## IMPORTANT:
 ## The interop feature branches track api revisions separately to the mainline branch.
 ## API version of last merge from corda mainline: 731
```
ac101m: This comment is here so that when merging mainline, we can tell at a glance whether there have been breaking changes in since our last merge. Please update it so that it matches the mainline version that is being merged in for this PR! (assuming it's not still 731)

ac101m (changes requested):

gradle.properties
ac101m: This doesn't appear to be resolved. Did you push your changes? Based on the commit history, this should be 742 now.

ac101m (approved):
Noice, LGTM!

szymonsztuka (approved):

gradle.properties
SaurabhR3: I have pushed it now, but I feel like another manual overhead here. Synching is ongoing process right now with interop branch, so I think we really don't need to maintain it. While merging I don't need to check/validate what was the last version of mainline. Also main branch is also keep updating.  @ac101m is there any specific requirement for this?

ac101m: It's just a convenience thing. Right now, it's unlikely we'll do a merge from mainline where there isn't a breaking change. But later this may not be the case, and in those cases our api version won't need to be incremented. In that case, I wanted to be able to see what the last merged version was without having to go check the commit history. This way, it's right there above the merge markers and can be done during the merge. It's not a big deal or anything and I'm perfectly happy for it to be removed, but if the comment is present it at very least shouldn't be incorrect.

Also yes, this is a very manual process. Blame the platform team :upside_down_face:

PR #1037, opened 2023-04-04
CORE-10583 Block receipt of messages in the Link Manager from suspended members

charlieR3:
This PR extends is existing filters to allow for suspended member info to be used to create sessions between peers. The MGM must be able to contact suspended members and vice-versa. 

Added to support https://github.com/corda/corda-runtime-os/pull/3555


williamvigorr3 (approved):
LGTM

PR #1039, opened 2023-04-04
CORE-10424 : Config changes to fix the interop processor

SaurabhR3:
Added required config for state and event topic

### DIFF
--- a/data/topic-schema/src/main/java/net/corda/schema/Schemas.java
+++ b/data/topic-schema/src/main/java/net/corda/schema/Schemas.java
@@ -94,6 +94,8 @@ public final class Schemas {
         public static final String FLOW_MAPPER_EVENT_STATE_TOPIC = getStateAndEventStateTopic(FLOW_MAPPER_EVENT_TOPIC);
         public static final String FLOW_MAPPER_EVENT_DLQ_TOPIC = getStateAndEventDLQTopic(FLOW_MAPPER_EVENT_TOPIC);
         public static final String FLOW_INTEROP_EVENT_TOPIC = "flow.interop.event";
+        public static final String FLOW_INTEROP_EVENT_STATE_TOPIC = getStateAndEventStateTopic(FLOW_INTEROP_EVENT_TOPIC);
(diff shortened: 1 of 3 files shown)


jmacmahonr3 (approved):
Looks fine to me

szymonsztuka:
need to bump up API version

SaurabhR3:
> need to bump up API version

Why do  we need to bump the version for this? I believe it should generate the 2.beta again once merged ?

yarivus (approved):
LGTM

szymonsztuka:
> > need to bump up API version
> 
> Why do we need to bump the version for this? I believe it should generate the 2.beta again once merged ?

We exactly don't want to rebuild artifact under the same version, what if Hiberate would republish hiberante-core.5.2.2.jar twice?

szymonsztuka (approved):

SaurabhR3:
> > > need to bump up API version
> > 
> > 
> > Why do we need to bump the version for this? I believe it should generate the 2.beta again once merged ?
> 
> We exactly don't want to rebuild artifact under the same version, what if Hiberate would republish hiberante-core.5.2.2.jar twice?

I believe it adds up the random_number based on some logic, and we are you using beta+ in corda-runtime-os, so it will pick the latest as there is no breaking changes. In this case it's better to avoid the version increment. I have increased it now, but this is what my opinion.

szymonsztuka:
> > > > need to bump up API version
> > > 
> > > 
> > > Why do we need to bump the version for this? I believe it should generate the 2.beta again once merged ?
> > 
> > 
> > We exactly don't want to rebuild artifact under the same version, what if Hiberate would republish hiberante-core.5.2.2.jar twice?
> 
> I believe it adds up the random_number based on some logic, and we are you using beta+ in corda-runtime-os, so it will pick the latest as there is no breaking changes. In this case it's better to avoid the version increment. I have increased it now, but this is what my opinion.

There's no random number for client project using this dependency, this broken JAR already polluted local caches of anyone using it. Hence you need to inform anyone that there is new remote version of the artifact by bumping up this number.

PR #1041, opened 2023-04-05
CORE-10576 Add group parameters to suspension/activation response

williamvigorr3:
If the member being suspended/activated is a notary then we need to update the group parameters.


charlieR3 (dismissed):

charlieR3 (dismissed):

nikinagy (dismissed):
LGTM

dimosr (dismissed):

dimosr (approved):

PR #1042, opened 2023-04-06
CORE-7269 Storage of MGM group parameters in UTXO ledger

vlajos:
CORE-7269 Support persisting MGM group parameters in vault.

Runtime: https://github.com/corda/corda-runtime-os/pull/3540

### DIFF
--- a/data/avro-schema/src/main/resources/avro/net/corda/data/ledger/persistence/FindSignedGroupParametersResponse.avsc
+++ b/data/avro-schema/src/main/resources/avro/net/corda/data/ledger/persistence/FindSignedGroupParametersResponse.avsc
@@ -1,0 +1,16 @@
+{
+  "type": "record",
+  "name": "FindSignedGroupParametersResponse",
+  "doc": "Response format for FindSignedGroupParametersResponse",
+  "namespace": "net.corda.data.ledger.persistence",
+  "fields": [
+    {
+      "name": "results",
+      "doc": "0 or 1 SignedGroupParameters depending on the success of the query.",
+      "type": {
+        "type": "array",
+        "items": "net.corda.data.membership.SignedGroupParameters"
+      }
+    }
+  ]
+}
--- a/data/db-schema/src/main/resources/net/corda/db/schema/vnode-vault/migration/ledger-utxo-creation-v1.0.xml
+++ b/data/db-schema/src/main/resources/net/corda/db/schema/vnode-vault/migration/ledger-utxo-creation-v1.0.xml
@@ -212,6 +212,9 @@
             <column name="signature_spec" type="NVARCHAR(255)">
                 <constraints nullable="false"/>
             </column>
+            <column name="created" type="TIMESTAMP">
+                <constraints nullable="false"/>
+            </column>
         </createTable>
     </changeSet>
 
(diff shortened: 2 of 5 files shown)


data/db-schema/src/main/resources/net/corda/db/schema/vnode-vault/migration/ledger-utxo-creation-v1.0.xml
```diff
@@ -190,6 +190,29 @@
         <addPrimaryKey columnNames="transaction_id, file_checksum"
                        constraintName="utxo_transaction_cpk_pkey"
                        tableName="utxo_transaction_cpk"/>
+
+        <comment>
+            Keep utxo_group_parameters's relevant fields in sync with vnode-vault-creation-v1.0.xml's vnode_group_parameters.
+        </comment>
+
+        <createTable tableName="utxo_group_parameters">
```
lankydan, line 198: Do we need a timestamp/created column like we have in most of our other tables?

vlajos: For consistency we could add. But the date itself can reflect only when we saved the record.
If there  are any dates inside the parameters, we won't be consistent with them.
Also the records coming from back chain resolution will have different dates in the different databases.  (Or we need to spend some time to stick the first date to the record when we send it forth and back.)

`vnode_group_parameters` which is roughly the same have an epoch field.

data/avro-schema/src/main/resources/avro/net/corda/data/ledger/persistence/FindSignedGroupParametersResponse.avsc
```diff
@@ -0,0 +1,16 @@
...
+  "doc": "Response format for FindSignedGroupParametersResponse",
+  "namespace": "net.corda.data.ledger.persistence",
+  "fields": [
+    {
+      "name": "results",
+      "doc": "0 or 1 SignedGroupParameters depending on the success of the query.",
+      "type": {
+        "type": "array",
+        "items": "net.corda.data.membership.SignedGroupParameters"
+      }
```
blsemo, line 13: Does AVRO not do a nullable type, so could just return `SignedGroupParameters` or `null`?

data/db-schema/src/main/resources/net/corda/db/schema/vnode-vault/migration/ledger-utxo-creation-v1.0.xml
blsemo: I think a timestamp column would be good, and if it's just for as-of queries. On the other hand maybe not essential as we don't expect this to change a lot, and I can't foresee paged queries of group paramters.

blsemo (dismissed):
It's ok, but I have a couple of question/comments.

data/avro-schema/src/main/resources/avro/net/corda/data/ledger/persistence/FindSignedGroupParametersResponse.avsc
vlajos: I did not check that. Our EntityResponse worked the same way.
Also we'll may need a way to query multiple ones in the same go.
So we can just reuse this container format.

data/db-schema/src/main/resources/net/corda/db/schema/vnode-vault/migration/ledger-utxo-creation-v1.0.xml
vlajos: OK. I have just added it.

blsemo (approved):
LGTM

PR #1044, opened 2023-04-07
CORE-11477 Revisit crypto configuration

kyriathar:
- removes unused `cryptoConnectionFactory` configuration
- removes unused `hsmService` configuration
- removes duplicate `wrappingKeys` configuration
- replaces `hsmMap` configuration which was used to hold multiple HSM configurations with `hsm` (as we only support soft hsm)
- generalizes caching configurations
- generalizes retrying configurations

runtime-os PR: [#3578](https://github.com/corda/corda-runtime-os/pull/3578)

PR #1047, opened 2023-04-11
CORE-11477 Revisit crypto hsm configuration

kyriathar:
- removes the following unused hsm configuration:
  - `hsm.categories`
  - `hsm.capacity`
  - `hsm.masterKeyAlias`
  - `hsm.supportedSchemes`
  - `hsm.masterKeyPolicy`
- flattens `hsm.config` properties into `hsm`

This PR addresses comments from [#1044](https://github.com/corda/corda-api/pull/1044) 

runtime-os PR: [#3593](https://github.com/corda/corda-runtime-os/pull/3593)

### DIFF
--- a/data/config-schema/src/main/resources/net/corda/schema/configuration/crypto/1.0/corda.crypto.json
+++ b/data/config-schema/src/main/resources/net/corda/schema/configuration/crypto/1.0/corda.crypto.json
@@ -4,6 +4,7 @@
   "title": "Corda Crypto Library Configuration Schema",
   "description": "Configuration schema for the crypto library subsection.",
   "type": "object",
+  "default": {},
   "properties": {
     "caching": {
       "description": "Caching settings",
--- a/data/config-schema/src/main/resources/net/corda/schema/configuration/crypto/1.0/corda.crypto.json
+++ b/data/config-schema/src/main/resources/net/corda/schema/configuration/crypto/1.0/corda.crypto.json
@@ -163,7 +163,6 @@
         "masterKeyPolicy",
         "capacity",
         "supportedSchemes",
-        "categories",
         "cfg"
       ],
       "additionalProperties": false
--- a/data/config-schema/src/main/resources/net/corda/schema/configuration/crypto/1.0/corda.crypto.json
+++ b/data/config-schema/src/main/resources/net/corda/schema/configuration/crypto/1.0/corda.crypto.json
@@ -92,14 +92,6 @@
           },
           "additionalProperties": false
         },
-        "wrappingKeyPolicy": {
-          "description": "Wrapping key policy, NONE - no wrapping key is required, SHARED - wrapping key shared by all tenants, UNIQUE - wrapping key unique for each tenant",
-          "enum": [
-            "NONE",
-            "SHARED",
-            "UNIQUE"
-          ]
-        },
         "wrappingKeys": {
           "description" : "Key derivation parameters for wrapping keys supplied in config",
           "type": "array",
@@ -138,7 +130,6 @@
         }
       },
       "required": [
-        "wrappingKeyPolicy",
         "wrappingKeys",
         "defaultWrappingKey"
       ],
--- a/data/config-schema/src/main/resources/net/corda/schema/configuration/crypto/1.0/corda.crypto.json
+++ b/data/config-schema/src/main/resources/net/corda/schema/configuration/crypto/1.0/corda.crypto.json
@@ -99,10 +99,6 @@
             "UNIQUE"
           ]
         },
-        "masterKeyAlias": {
-          "description": "If the masterKeyPolicy is SHARED then this specifies the alias of the shared key and as such is required",
-          "type": "string"
-        },
         "supportedSchemes": {
           "description": "Key schemes that are supported by the HSM",
           "type": "array",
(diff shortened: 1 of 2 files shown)


data/config-schema/src/main/resources/net/corda/schema/configuration/crypto/1.0/corda.crypto.json
```diff
@@ -91,107 +92,55 @@
           },
           "additionalProperties": false
         },
-        "categories": {
-          "description": "Categories which the HSM can be used for",
+        "wrappingKeyPolicy": {
```
driessamyn: This needs to go.

data/config-schema/src/main/resources/net/corda/schema/configuration/crypto/1.0/corda.crypto.json
```diff
@@ -73,7 +74,7 @@
       "type": "object",
       "default": {},
       "properties": {
-        "retry": {
+        "retrying": {
```
driessamyn, line 77: Why?

driessamyn (changes requested):

data/config-schema/src/main/resources/net/corda/schema/configuration/crypto/1.0/corda.crypto.json
kyriathar: To be more consistent with above `retrying` property. I could merge the two but this one seems specific to hsm I think.

data/config-schema/src/main/resources/net/corda/schema/configuration/crypto/1.0/corda.crypto.json
kyriathar: Ok, I will remove and hardcode it to be our current default which is `MasterKeyPolicy.UNIQUE`

kyriathar: Removed

driessamyn (dismissed):

data/config-schema/src/main/resources/net/corda/schema/configuration/crypto/1.0/corda.crypto.json
```diff
@@ -91,107 +92,46 @@
           },
           "additionalProperties": false
         },
-        "categories": {
-          "description": "Categories which the HSM can be used for",
+        "wrappingKeys": {
+          "description" : "Key derivation parameters for wrapping keys supplied in config",
```
LWogan, line 96: missing "default": {}, under wrappingKeys

data/config-schema/src/main/resources/net/corda/schema/configuration/crypto/1.0/corda.crypto.json
```diff
@@ -91,107 +92,46 @@
...
-          "enum": [
-            "NONE",
-            "SHARED",
-            "UNIQUE"
-          ]
-        },
-        "masterKeyAlias": {
-          "description": "If the masterKeyPolicy is SHARED then this specifies the alias of the shared key and as such is required",
+        "defaultWrappingKey": {
+          "description": "The default wrapping key, which must be in the wrappingKeys array.",
```
LWogan, line 128: This is marked as type string and not nullable. I think the integration test will fail if you dont define a default value such as the one below. As the integration test passes an empty config to the schema and asks for it to generate defaults. but as the field is "required" then it will fail validation as theres no default value or real value set.

"default": "",

LWogan (commented):
added comments about why your integration test is failing. not 100% if default empty values is what makes sense as the solution.

data/config-schema/src/main/resources/net/corda/schema/configuration/crypto/1.0/corda.crypto.json
dickon: Why? At least one wrapping key must be specified

data/config-schema/src/main/resources/net/corda/schema/configuration/crypto/1.0/corda.crypto.json
dickon: Again that was intentional, there must not be defaults for the wrapping keys.

dickon (approved):
LGTM

dickon (changes requested):
Hang on, not sure why we deleting supported schemes.

data/config-schema/src/main/resources/net/corda/schema/configuration/crypto/1.0/corda.crypto.json
LWogan: after some discussion with dickon and kyriakos, i think the path forward to to not set defaults for the required fields and rely on the config passed to the defaulter to have these fields set already

kyriathar:
>Hang on, not sure why we deleting supported schemes.

I believe it was not actually used anywhere.

dickon (dismissed):
@kyriathar points out that the supported schemes config setting is not pickde up and has no effect. That's more dangerous than not having it. So we will need to figure out how to do this properly and add it back in.

driessamyn (approved):
LGTM

mnesbit (approved):
OK

PR #1048, opened 2023-04-13
CORE-11960 Capture the originator of a notarization request

relyafi:
### Summary

This PR makes the necessary DB schema and API changes to allow the originator of a notarization request to be captured within the uniqueness DB for audit purposes.

Note that this is technically a "breaking" change due to the addition of an extra parameter to `LedgerUniquenessCheckerClientService.requestUniquenessCheck`. However, in practice this is currently only used by our non-validating notary server plugin, which has been updated as part of the corresponding `corda-runtime-os` PR.

### Testing

Tested as part of the corresponding `corda-runtime-os` PR, see https://github.com/corda/corda-runtime-os/pull/3607.


blsemo (dismissed):
The change itself looks good to me.
The breaking change should be ok as it's on an interface that is currently only implemented in runtime os.

blsemo (approved):
Re-approved

PR #1051, opened 2023-04-14
CORE-6123 Adjust some default values to ensure they relate to each other correctly.

LWogan:
Add additional text to describe the relationship between certain configs


mnesbit (approved):
OK

PR #1052, opened 2023-04-14
CORE-11753 Change JSON column type based on db type

nkovacsx:
Tested using smoke tests for Postgres.

### DIFF
--- a/data/db-schema/src/main/resources/net/corda/db/schema/vnode-vault/migration/ledger-utxo-visible-transaction-state-creation-v1.0.xml
+++ b/data/db-schema/src/main/resources/net/corda/db/schema/vnode-vault/migration/ledger-utxo-visible-transaction-state-creation-v1.0.xml
@@ -4,7 +4,6 @@
                    xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-4.3.xsd">
 
     <property name="json.column.type" value="JSONB" dbms="postgresql"/>
-    <property name="json.column.type" value="VARCHAR(1024)" dbms="hsqldb"/>
 
     <changeSet author="R3.Corda" id="ledger-utxo-visible-transaction-state-creation-postgres-v1.0">
 
--- a/data/db-schema/src/main/resources/net/corda/db/schema/vnode-vault/migration/ledger-utxo-visible-transaction-state-creation-v1.0.xml
+++ b/data/db-schema/src/main/resources/net/corda/db/schema/vnode-vault/migration/ledger-utxo-visible-transaction-state-creation-v1.0.xml
@@ -4,6 +4,8 @@
                    xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-4.3.xsd">
 
     <property name="json.column.type" value="JSONB" dbms="postgresql"/>
+
+    <!-- Please note that HSQLDB schema is supported for integration test purposes only. -->
     <property name="json.column.type" value="CLOB" dbms="hsqldb"/>
 
     <changeSet author="R3.Corda" id="ledger-utxo-visible-transaction-state-creation-v1.0">
--- a/data/db-schema/src/main/resources/net/corda/db/schema/vnode-vault/migration/ledger-utxo-visible-transaction-state-creation-v1.0.xml
+++ b/data/db-schema/src/main/resources/net/corda/db/schema/vnode-vault/migration/ledger-utxo-visible-transaction-state-creation-v1.0.xml
@@ -4,8 +4,9 @@
                    xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-4.3.xsd">
 
     <property name="json.column.type" value="JSONB" dbms="postgresql"/>
+    <property name="json.column.type" value="CLOB" dbms="hsqldb"/>
 
-    <changeSet author="R3.Corda" id="ledger-utxo-visible-transaction-state-creation-postgres-v1.0">
+    <changeSet author="R3.Corda" id="ledger-utxo-visible-transaction-state-creation-v1.0">
 
         <createTable tableName="utxo_visible_transaction_state">
             <column name="transaction_id" type="VARCHAR(160)">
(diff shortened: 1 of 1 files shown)


data/db-schema/src/main/resources/net/corda/db/schema/vnode-vault/migration/ledger-utxo-visible-transaction-state-creation-v1.0.xml
```diff
@@ -3,7 +3,10 @@
                    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
                    xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-4.3.xsd">
 
-    <changeSet author="R3.Corda" id="ledger-utxo-visible-transaction-state-creation-postgres-v1.0" dbms="postgresql">
+    <property name="json.column.type" value="JSONB" dbms="postgresql"/>
+    <property name="json.column.type" value="VARCHAR(1024)" dbms="hsqldb"/>
```
nkovacsx: @mnesbit @blsemo There was an argument about using `CLOB` instead of `VARCHAR(1024)`. Any strong opinions on this?

mnesbit: Yes I have a strong opinion that our public API shouldn't suggest we support a DB type we haven't in fact made work

data/db-schema/src/main/resources/net/corda/db/schema/vnode-vault/migration/ledger-utxo-visible-transaction-state-creation-v1.0.xml
```diff
@@ -3,7 +3,9 @@
                    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
                    xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-4.3.xsd">
 
-    <changeSet author="R3.Corda" id="ledger-utxo-visible-transaction-state-creation-postgres-v1.0" dbms="postgresql">
+    <property name="json.column.type" value="JSONB" dbms="postgresql"/>
+
+    <changeSet author="R3.Corda" id="ledger-utxo-visible-transaction-state-creation-postgres-v1.0">
```
chrisr3: You can now remove "postgres" from the changeset id. Happy to approve with this change.

data/db-schema/src/main/resources/net/corda/db/schema/vnode-vault/migration/ledger-utxo-visible-transaction-state-creation-v1.0.xml
```diff
@@ -3,7 +3,9 @@
                    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
                    xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-4.3.xsd">
 
-    <changeSet author="R3.Corda" id="ledger-utxo-visible-transaction-state-creation-postgres-v1.0" dbms="postgresql">
+    <property name="json.column.type" value="JSONB" dbms="postgresql"/>
```
chrisr3, line 6: Maybe also add:
```xml
<property name="json.column.type" value="CLOB" dbms="hsqldb"/>
```

chrisr3 (commented):
Just a few thoughts.

data/db-schema/src/main/resources/net/corda/db/schema/vnode-vault/migration/ledger-utxo-visible-transaction-state-creation-v1.0.xml
chrisr3: Personally, I think we _should_ **make** HSQLDB work, just so that we can prove that we _can_ support multiple database types. `CLOB` sounds about right for HSQLDB.

But _not_ providing a `json.column.type` value for HSQLDB means that Liquibase will fail to load the vault schema for anyone running the integration tests locally. Which in turn will **_massively_** discourage people from writing any integration tests for this functionality.

---
FWIW [this PR](https://github.com/corda/corda-runtime-os/pull/2609) also happens to fix _all_ of the (existing) integration tests to work _locally_ with HSQLDB.

chrisr3: > Yes I have a strong opinion that our public API shouldn't suggest we support a DB type we haven't in fact made work.

Fortunately, we can still add this value for HSQLDB _programmatically_ in `LiquibaseSchemaMigratorImpl`:
```kotlin
 lb.changeLogParameters.set("json.column.type", "CLOB", null, null, "hsqldb")
```
So this should make everyone happy.

nkovacsx: I'm happy to go either way, just wanted to make sure we are on the same page here. @mnesbit what do you think?

mnesbit: I have chatted to Chris and I am ok with it too

data/db-schema/src/main/resources/net/corda/db/schema/vnode-vault/migration/ledger-utxo-visible-transaction-state-creation-v1.0.xml
```diff
@@ -3,7 +3,10 @@
...
+    <property name="json.column.type" value="CLOB" dbms="hsqldb"/>
```
chrisr3, line 9: I'm confused - I thought we'd agreed _not_ to define `json.column.type` for HSQLDB here, but to set this programmatically in `LiquibaseSchemaMigratorImpl` instead?

---

Nothing has changed - I'm _still_ confused!

data/db-schema/src/main/resources/net/corda/db/schema/vnode-vault/migration/ledger-utxo-visible-transaction-state-creation-v1.0.xml
nkovacsx: I've made the changes, @mnesbit @chrisr3 let me know what you think.

nkovacsx:
After discussing this with @mnesbit and @chrisr3 we came to the conclusion that we can add the HSQLDB/Postgres property switch into the API liquibase XML. 

We also decided to add a comment just to make sure users are aware of the fact that HSQLDB is only used for testing purposes as of now.

mnesbit (approved):
OK

chrisr3 (approved):
LGTM - _and relax!_ :relieved:

PR #1054, opened 2023-04-15
CORE-3819: Replace deprecated Gradle APIs.

chrisr3:
Replace deprecated APIs which have been removed in Gradle 8+. Also only configure any `JavaCompile` and `Test` tasks if the module applies the `java` plugin. (This will also become important for Gradle 8+.)


ronanbrowne (approved):
lgtm

PR #1062, opened 2023-04-20
CORE-10330: Remove Kotlin from transitive dependencies.

chrisr3:
We no longer need a runtime dependency on `kotlin-osgi-bundle` now that everything is written in Java.

Also upgrade our tests to use Kotlin 1.8.20, for consistency.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -10,10 +10,10 @@ cordaProductVersion = 5.0.0
 # NOTE: update this each time this module contains a breaking change
 ## NOTE: currently this is a top level revision, so all API versions will line up, but this could be moved to
 ##   a per module property in which case module versions can change independently.
-cordaApiRevision = 754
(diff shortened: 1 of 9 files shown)


gradle.properties
```diff
@@ -13,7 +13,7 @@ cordaProductVersion = 5.0.0
 cordaApiRevision = 754
```
vkolomeyko: Arguably, it is a breaking change, hence version needs to be bumped.

chrisr3: It _shouldn't_ have been a breaking change, because that dependency no longer has any business _existing_ in `corda-api`.

It appears the "Kotlin -> Java" migration was "incomplete".

wzur-r3 (approved):
LGTM

yarivus (approved):
LGTM

PR #1063, opened 2023-04-20
CORE-10426 : Sync release with interop

SaurabhR3:
Sync release 5.0 with interop


szymonsztuka (approved):

ac101m (commented):
At a glance, not seeing anything that should cause issues with the flow mapper. Should note though that the interop smoke test is currently disabled, so any failures _won't_ show up on Jenkins. I will have to check the branch locally. Approval to follow.

P.S. Typo in title, should be "sync" not "synch"!

ac101m (approved):
LGTM, as my issues appear to be local.

Should fix the spelling mistake in the title though, as that will end up as part of the commit history.

PR #1064, opened 2023-04-21
CORE-12860 Create keys for uniqueness tables as part of table creation

relyafi:
### Summary

Previously, the Liquibase changelog for the uniqueness DB first created a table (and its columns), before adding a key in a subsequent changeset. This works fine for Postgres, however, with Cockroach DB (not yet officially supported), this results in Cockroach DB creating a default primary key using just the row id of a record. When the changeset to then add the key is run, it detects the key already exists, and skips the changeset. Even if this was not done, it would fail to add the key due to it already existing.

This change creates the primary key as part of the same changeset as the table creation, which results in the correct behavior for Cockroach DB. Whilst not officially supported yet, making this change now avoids potential migration / backwards compatibility headaches down the line, since it will be difficult to make changes (as opposed to additions) to the liquibase changelogs post-GA.

NOTE: This change does not require an API version bump, as there are no implications on the `corda-runtime-os` repo and the end result (on Postgres) is the same.

### Testing

Created a Cockroach DB cluster, and ran the `backingStoreBenchmark` command against this DB. Verified that uniqueness DB tables were created with the correct keys.

Also ran the uniqueness checker DB integration tests against HSQLDB and Postgres, and standard smoke tests against Postgres, and confirmed no regressions. Also verified the keys remain correct on a Postgres by manual inspection.

### DIFF
--- a/data/db-schema/src/main/resources/net/corda/db/schema/vnode-uniqueness/migration/vnode-uniqueness-creation-v1.0.xml
+++ b/data/db-schema/src/main/resources/net/corda/db/schema/vnode-uniqueness/migration/vnode-uniqueness-creation-v1.0.xml
@@ -30,18 +30,12 @@
                 <constraints nullable="true"/>
             </column>
         </createTable>
-    </changeSet>
-
-    <changeSet author="R3.Corda" id="uniqueness-creation-v1.0-2" runInTransaction="false">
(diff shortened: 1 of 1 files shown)


blsemo (approved):
LGTM

PR #1072, opened 2023-04-26
CORE-12912 `ResultSet.next` API changes

lankydan:
Usage of the API would look like

```kotlin
val resultSet = query.execute()
results += resultSet.results
while (resultSet.hasNext()) {
  results += resultSet.next()
}
return results
```

### DIFF
--- a/application/src/main/java/net/corda/v5/application/persistence/PagedQuery.java
+++ b/application/src/main/java/net/corda/v5/application/persistence/PagedQuery.java
@@ -63,7 +63,8 @@ public interface PagedQuery<R> {
          * Extracts the results of a {@link ResultSet} from a previously executed query.
          * <p>
          * This method does not execute a query itself, call {@link PagedQuery#execute()} to execute a query and
-         * generate a {@link ResultSet}.
+         * generate a {@link ResultSet} or {@link ResultSet#next()} with an existing {@link ResultSet} to re-execute the
+         * query with the offset increased to the next page's offset.
          *
          * @return The results contained in the result set.
          */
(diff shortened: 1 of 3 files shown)


lankydan:
Runtime - https://github.com/corda/corda-runtime-os/pull/3729

application/src/main/java/net/corda/v5/application/persistence/PagedQuery.java
```diff
@@ -68,5 +69,33 @@
...
+         * behaviour when calling {@link ResultSet#next()}.
+         *
+         * @return {@code true} if the {@link ResultSet} has more results to retrieve.
+         *
+         * @return
+         */
+        boolean hasNext();
+
+        /**
+         * Executes the query linked to the {@link ResultSet} to retrieve the next page of results.
```
driessamyn: I'm ok with this API change, but it could be confusing because we don't have any guarantees that the resultset doesn't change between result pages. This is one of the reasons I originally preferred the explicit offset/limit API which reflects what a DB would do.
This API could give the impression that the resultset is fixed when `execute()` is called and you page through those results, which is not the case.
So if we implement this API, I think we need to be very specific about the fact that these guarantees don't exist in the general docs but also i these JavaDocs I think.

lankydan:
Was discussing with @owenstanford. I could change `next` to shift the page along without returning anything and only accessing results using `getResults`, which would look like:

```kotlin
val resultSet = query.execute()
while (resultSet.hasNext()) {
  results += resultSet.results()
  resultSet.next()
}
return results
```

This would move it away from looking like an `Iterator` but might be conceptually easier to understand? This code can actually still be achieved with my current implementation if you decide to never use the results returned by `next`.

driessamyn (dismissed):
LGTM

blsemo (commented):
LGTM

blsemo (dismissed):

blsemo (approved):
Re-approved

PR #1076, opened 2023-04-27
CORE-12804: Add error kind

yift-r3:
This will add an error kind to the membership persistence failure response. It will allow us to propagate the type of error from the DB worker to the persistent client.

Currently, it has only one kind (that is needed for CORE-12804); we should add more kinds in the future (for example, for CORE-11847)

The runtime pull request is in https://github.com/corda/corda-runtime-os/pull/3746


williamvigorr3 (dismissed):
LGTM

williamvigorr3 (dismissed):
LGTM

williamvigorr3 (approved):
LGTM

dimosr (approved):

PR #1079, opened 2023-04-28
CORE-11162: Remove source and destination from inbound unauthenticated messages

yift-r3:
This pull request removes the source and destination from the inbound unauthenticated message header. It also adds the MGM to the registration request header (to be able to decrypt the message).

The runtime pull request is in https://github.com/corda/corda-runtime-os/pull/3757.


dimosr (approved):
LGTM

PR #1080, opened 2023-04-28
CORE-12539 - Update jetbrains annotations version

driessamyn:


chrisr3:
We don't need to do this. We are using 13.0 **because this is the version that Kotlin uses**, _despite JetBrains also being the company that creates Kotlin!_

I am assuming that JetBrains has a Good Reason for keeping Kotlin using v13.0 of their own annotations rather than upgrading it to use 24.0.1.

The only reason we are including these annotations in the first place is because the Kotlin compiler can detect `@Nullable` and `@NotNull` on Java signatures to provide better Java /Kotlin interoperability. And since Kotlin depends on `annotations:13.0`, the Kotlin compiler must therefore ignore any extra annotations added in later versions.

I _**strongly**_ recommend leaving these annotations at 13.0 until Kotlin itself decides to upgrade.

There is no risk of CVEs from a bundle containing only annotations.

gradle.properties
```diff
@@ -17,7 +17,7 @@ kotlinVersion = 1.8.21
 kotlin.stdlib.default.dependency = false
 
 # These are the same annotations that Kotlin uses.
-jetbrainsAnnotationsVersion = 13.0
+jetbrainsAnnotationsVersion = 24.0.1
```
chrisr3, line 20: We don't need to do this. We are using 13.0 **because this is the version that Kotlin uses**, _despite JetBrains also being the company that creates Kotlin!_

I am assuming that JetBrains has a Good Reason for keeping Kotlin using v13.0 of their own annotations rather than upgrading it to use 24.0.1 :no_mouth:.

The only reason we are including these annotations in the first place is because the Kotlin compiler can detect `@Nullable` and `@NotNull` on Java signatures to provide better Java /Kotlin interoperability. And since Kotlin depends on `annotations:13.0`, the Kotlin compiler must therefore ignore any extra annotations added in later versions.

I _**strongly**_ recommend leaving these annotations at 13.0 until Kotlin itself decides to upgrade.

There is no risk of CVEs from a bundle containing only annotations.

PR #1083, opened 2023-05-02
CORE-8714 Move the protocol mode from the Initiator(Responder)HelloMesssage to the HandshakeMessages

williamvigorr3:
Send the protocol mode as part of the handshake messages instead of the hello messages.


dimosr (approved):
LGTM

PR #1084, opened 2023-05-02
CORE-10329 Create new context map for non-member context properties

charlieR3:
This change introduces an additional context map to the internal network registration request. This is because the original context map is signed over and the signature becomes invalid if the map is modified so whatever is submitted in that context map became part of the member's member context map. It's preferred to not include the pre-auth token in the member info distributed across the network. To support this and other properties in the future which should not be distributed, an additional context map has been added. 

This new context map is also signed. To avoid duplication, data and a signature over that data was pulled out in to it's own schema and membership schemas were updated where the same structure is already in use. 

Runtime OS PR: https://github.com/corda/corda-runtime-os/pull/3781

### DIFF
--- a/data/avro-schema/src/main/resources/avro/net/corda/data/membership/common/RegistrationRequestDetails.avsc
+++ b/data/avro-schema/src/main/resources/avro/net/corda/data/membership/common/RegistrationRequestDetails.avsc
@@ -56,12 +56,12 @@
       "type": "net.corda.data.KeyValuePairList"
     },
     {
-      "name": "registrationSignature",
+      "name": "registrationContextSignature",
       "doc": "Signature of the member over the registrationContext.",
       "type": "net.corda.data.crypto.wire.CryptoSignatureWithKey"
     },
     {
-      "name": "registrationSignatureSpec",
+      "name": "registrationContextSignatureSpec",
       "doc": "Signature spec of member signature over the registrationContext.",
       "type": "net.corda.data.crypto.wire.CryptoSignatureSpec"
     },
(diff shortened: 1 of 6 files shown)


data/avro-schema/src/main/resources/avro/net/corda/data/membership/common/RegistrationRequestDetails.avsc
```diff
@@ -50,6 +50,21 @@
       "doc": "Signature spec of member signature.",
       "type": "net.corda.data.crypto.wire.CryptoSignatureSpec"
     },
+    {
+      "name": "registrationContext",
+      "doc": "Registration context submitted for registration.",
+      "type": "net.corda.data.KeyValuePairList"
+    },
+    {
+      "name": "registrationSignature",
```
dimosr: would `registrationContextSignature` be better?

data/avro-schema/src/main/resources/avro/net/corda/data/membership/common/RegistrationRequestDetails.avsc
```diff
@@ -50,6 +50,21 @@
...
+      "doc": "Signature of the member over the registrationContext.",
+      "type": "net.corda.data.crypto.wire.CryptoSignatureWithKey"
+    },
+    {
+      "name": "registrationSignatureSpec",
```
dimosr: similarly `registrationContextSignatureSpec`?

dimosr (commented):
LGTM - just a couple of minor comments/question on naming (that sadly will become part of the wire format, otherwise I wouldn't care)

dimosr (approved):

dimosr:
LGTM (pending conflicts resolution and bumping to the latest api version)

PR #1086, opened 2023-05-02
CORE-9980 update send transaction builder api doc

vlajos:
Since Commands and Output states do not get deduplicated.


blsemo (approved):
LGTM

PR #1087, opened 2023-05-03
CORE-11656 Align `DigestServiceImpl.parseSecureHash` kDoc to added validations

kyriathar:
- aligns `DigestServiceImpl.parseSecureHash` kDoc to added validations on secure hash parsing; that 1. digest algorithm is supported and 2. the hex value of the hash is of the expected length


driessamyn (approved):
LGTM

PR #1089, opened 2023-05-03
CORE-13257: Don't allow other sandbox types in configuration

fowlerrr:
We have three different types of sandbox.  Adding a field in the schema to disallow configuration of other types (that don't actually exist in code) will keep users from accidentally setting the wrong thing.  (eg. `db` instead of `persistence`)

### DIFF
--- a/data/config-schema/src/main/resources/net/corda/schema/configuration/sandbox/1.0/corda.sandbox.json
+++ b/data/config-schema/src/main/resources/net/corda/schema/configuration/sandbox/1.0/corda.sandbox.json
@@ -4,8 +4,8 @@
   "title": "Corda Sandbox Configuration Schema",
   "description": "Configuration schema for the sandbox section.",
   "type": "object",
+  "additionalProperties": false,
   "properties": {
-    "additionalProperties": false,
     "flow": {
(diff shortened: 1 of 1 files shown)


data/config-schema/src/main/resources/net/corda/schema/configuration/sandbox/1.0/corda.sandbox.json
```diff
@@ -4,6 +4,7 @@
   "title": "Corda Sandbox Configuration Schema",
   "description": "Configuration schema for the sandbox section.",
   "type": "object",
+  "additionalProperties": false,
```
LWogan, line 7: you have to add this at every "level" in this file. Anywhere that has a "properties" block

fowlerrr:
Testing this locally I was able to produce a 400:Bad Request error with 
```
{
  "title": "Configuration \"db {cache {size = -1}}\" could not be validated. Valid JSON or HOCON expected. Cause: Configuration failed to validate for key corda.sandbox at schema version null: $.db: is not defined in the schema and the schema does not allow additional properties",
  "status": 400,
  "details": {}
}
```

LWogan (approved):
LGTM

PR #1091, opened 2023-05-03
CORE-12555 Interop Facades as Corda service

szymonsztuka:
A Corda service (working) prototype to integrate Facades into flows (both client proxies and server dispatchers).


ac101m (approved):
LGTM

yarivus (approved):
LGTM

PR #1093, opened 2023-05-04
CORE-9481 remove filename from chunk, it is to be stored in properties instead

ciaran238:
Removed filename from chunk avro schema and made the related DB change, filename will be stored in properties instead


LWogan (approved):
LGTM. We've resolved the issues related to the previous merge attempt and theres green build in runtime-os with this latest api version

thiagoviana (approved):

PR #1112, opened 2023-05-10
ES-371/set correct license

seanbrereton:
As part of the C5 build process a license file is injected into the pom.xml of a produced Jar this license file is specified as part of the gradle.properties of each repo the current license properties are not correct / outdated and should be updated pre GA

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -20,9 +20,8 @@ kotlin.stdlib.default.dependency = false
 jetbrainsAnnotationsVersion = 13.0
 
 # License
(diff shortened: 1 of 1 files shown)


ronanbrowne:
Discussed with @sachingokhale and @marcellofarina75 OKd from the product for entering post freeze - needed to ensure correct POM generation

ronanbrowne (approved):

yarivus (commented):
thx @seanbrereton

PR #1113, opened 2023-05-10
ES-383: use classic JavaDoc style when generating api docs 

ronanbrowne:
With the removal of Kotlin code from this repo, we should be producing classic-styled JavaDoc rather than Dokka-style Kotlin docs.

This change modified the existing `generateMultiModuleDocs` task to hook into the `dokkaJavadocCollector` task rather than `dokkaHtmlMultiModule`.

Also, since Gradle 8.1.1 Gradle has become more opinionated about [implicit dependencies ](https://docs.gradle.org/8.1.1/userguide/validation_problems.html#implicit_dependency)between tasks that are not explicitly declared - this is why we need to change the configuration of `dokkaJavadocCollector` to add a dependency on `generateAvro `and `generateOSGiPackageInfo` without these the build will fail with the following 

`    Reason: Task ':dokkaJavadocCollector' uses this output of task ':data:avro-schema:generateOSGiPackageInfo' without declaring an explicit or implicit dependency. This can lead to incorrect results being produced, depending on what order the tasks are executed.`

The resulting output here is an archive containing JavaDoc in the classic style, as can be seen in the screen grab below taken from files generated in the local execution of this task:

image

Previous output with Kotlin styling:
image

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -348,6 +348,10 @@ tasks.register('generateMultiModuleDocs', Zip) {
 
 }
 
+tasks.named('dokkaJavadocCollector') {
+    dependsOn ':data:avro-schema:generateAvro', 'data:avro-schema:generateOSGiPackageInfo'
+}
+
 if (project.hasProperty('publishMultiModulekDocs')) {
     publishing {
         publications {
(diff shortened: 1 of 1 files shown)


ronanbrowne:
~~Issue on publish stage that needs to be resolved, putting back to draft while fixing~~

yarivus (approved):
LGTM

PR #1114, opened 2023-05-11
CORE-13100 - Make expiration timeframe and frequency configurable in ExpirationProcessor

nikinagy:
Operators should be able to configure expiration timeframe and frequency for registration request expiration.


data/config-schema/src/main/resources/net/corda/schema/configuration/membership/1.0/corda.membership.json
```diff
@@ -13,6 +13,20 @@
...
+      "description": "The maximum duration in minutes between polling for expired registration requests.",
+      "type": "integer",
+      "minimum": 10,
+      "maximum": 2880,
+      "default": 300
+    },
+    "expirationTimeFrame": {
+      "description": "The time frame in minutes after which a registration request is considered as expired and gets force declined.",
+      "type": "integer",
+      "minimum": 30,
```
charlieR3: Any reason the minimums are `10` and `30` and we don't allow any lower?

nikinagy: Mainly, I just didn't want to give a too low value.
For expirationTimeFrame I added a slightly higher value, because I am not sure what would be realistic value for the response time from member side and if I can see it right the default TTL for verifyMemberRequest is 20 mins, so I wanted something higher than that.
But we can lower the minimums to eg 3 mins and 20 mins if that sounds better?

charlieR3: > if I can see it right the default TTL for verifyMemberRequest is 20 mins

It's the default though and not the minimum so it could be set lower if someone wanted. Unless there's a technical reason it can't be low I'd prefer not to limit this too much. Maybe even set the minimum to 0 or 1 on both. I think we just need to set a reasonable default.

nikinagy: addressed in [8c3bdb1](https://github.com/corda/corda-api/pull/1114/commits/8c3bdb17b3f1728e1ed4712e232d4d2da11c058a)

charlieR3 (dismissed):

ronanbrowne (dismissed):
LGTM - (Jenkins change )

charlieR3 (approved):

PR #1117, opened 2023-05-11
CORE-10453 : Interface for remote alias lookup service

SaurabhR3:
Info for this change : 

Interop client flow will inject or use RemoteAliasLookUpService to fetch the required interop information.
RemoteAliasLookUpService will be loaded at the time of instantiation for interop worker/ flow worker.
Internally RemoteAliasLookUpService will use the cache of AliasIdentityInfo. 
Cache will contain the data of alias group info with holding identity short hash as a key.
RemoteAliasLookUpService will only fetch the alias group info for specific short hash and it will have cache entry for that short hash.
Interop client flow can query its own cached entry.

https://r3-cev.atlassian.net/wiki/spaces/CB/pages/4448321990/WIP+Alias+Identities+at+Flow+Level

Changes : 
Created interfaces for RemoteAliasLookUpService
Created new interop module


application/src/main/java/net/corda/v5/application/interop/RemoteAliasLookUpService.java
```diff
@@ -0,0 +1,16 @@
...
+ * <p>This interface represents the injectable service for client corDapp to look up the remote alias identity.
+ *
+ */
+@DoNotImplement
+public interface RemoteAliasLookUpService {
+
+    /**
+     * @return The {@link AliasMemberInfo} representing the alias member info for given alias member id.
+     */
+    AliasMemberInfo get(String identifier);
```
yarivus: maybe more fitting:
```suggestion
    AliasMemberInfo lookup(String identifier);
```

interop/build.gradle
```diff
@@ -0,0 +1,18 @@
+plugins {
```
ronanbrowne, line 1: FYI you have not declared this new module in the sessings.gradle in corda-api - so as far as gradle is concerned it will know nothing about this module.

you need to specify this in the included submodules here

https://github.com/corda/corda-api/blob/release/os/5.1/settings.gradle#L87-L104

your runtime-os equivalent branch should also update it's composite build logic to include this, like so 

https://github.com/corda/corda-runtime-os/blob/release/os/5.1/settings.gradle#L458-L476

ronanbrowne (changes requested):

application/src/main/java/net/corda/v5/application/interop/AliasMemberInfo.java
```diff
@@ -0,0 +1,35 @@
...
+import net.corda.v5.base.annotations.CordaSerializable;
+import org.jetbrains.annotations.NotNull;
+import java.util.List;
+
+/**
+ * <p>This interface represents the alias member details like x500name, facadeIds etc.
+ *
+ */
+@CordaSerializable
+public interface AliasMemberInfo {
```
yarivus: why is this an interface and not simply a data class ?

yarivus (commented):
see my comments

interop/build.gradle
SaurabhR3: Thanks @ronanbrowne

application/src/main/java/net/corda/v5/application/interop/AliasMemberInfo.java
SaurabhR3: Initially I thought about it, but when I looked at other corda-api modules all modules have interface for such data. I took the reference of membership and kept it consistent.

yarivus: OK thanks for clarifying

yarivus (approved):
LGTM

yarivus (approved):
LGTM

ronanbrowne (approved):
LGTM - build

yarivus (approved):
LGTM

yarivus (approved):
LGTM

szymonsztuka (approved):

szymonsztuka (approved):

PR #1120, opened 2023-05-12
CORE-13353 Added support for storing flow metric state in the checkpoint

owenstanford:
Adding support for storing flow metric state is the checkpoint. This is used for metrics that span flow suspensions.


owenstanford:
No plans to merge this to 5.0 just using this PR to generate an alpha build for some testing

PR #1122, opened 2023-05-15
CORE-11274 Add Java version for code examples in crypto API Javadocs

kyriathar:


dickon (changes requested):
@driessamyn  pointed out in other places we have both Java and Kotlin versions, so we should probably follow that pattern for consistency.

kyriathar:
>@driessamyn pointed out in other places we have both Java and Kotlin versions, so we should probably follow that pattern for consistency.

Example added in Kotlin too, alongside Java. Both Java and Kotlin version tested they compile and `digest` methods tested too.

dickon (approved):
LGTM

driessamyn (approved):
Approved (doc changes only, not code changes)

PR #1128, opened 2023-05-16
CORE-13740: Add minimal db connection pool size

yift-r3:
The runtime pull request is in https://github.com/corda/corda-runtime-os/pull/3909

This will add a minimal size of the connection pool size.

Additional properties will be added in CORE-11035. The minimal pool size is needed to cache a pool for each virtual node without having too many concurrent open connections.

### DIFF
--- a/data/config-schema/src/main/resources/net/corda/schema/configuration/db/1.0/corda.db.json
+++ b/data/config-schema/src/main/resources/net/corda/schema/configuration/db/1.0/corda.db.json
@@ -53,7 +53,7 @@
               "default": 10
             },
             "min_size": {
-              "description": "The minimal database pool size.",
+              "description": "The minimum database pool size.",
               "default": null,
               "anyOf": [
                 {
--- a/data/config-schema/src/main/resources/net/corda/schema/configuration/db/1.0/corda.db.json
+++ b/data/config-schema/src/main/resources/net/corda/schema/configuration/db/1.0/corda.db.json
@@ -53,7 +53,7 @@
               "default": 10
             },
             "min_size": {
-              "description": "The minimum database pool size.",
+              "description": "The minimum database pool size. If left null will default to the `max_size` value.",
               "default": null,
               "anyOf": [
                 {
--- a/data/config-schema/src/main/resources/net/corda/schema/configuration/db/1.0/corda.db.json
+++ b/data/config-schema/src/main/resources/net/corda/schema/configuration/db/1.0/corda.db.json
@@ -51,6 +51,19 @@
               "type": "integer",
               "minimum": 1,
               "default": 10
+            },
+            "min_size": {
+              "description": "The minimal database pool size.",
+              "default": null,
+              "anyOf": [
+                {
+                  "type": "integer",
+                  "minimum": 0
+                },
+                {
+                  "type": "null"
+                }
+              ]
             }
           },
           "additionalProperties": false
--- a/data/config-schema/src/main/java/net/corda/schema/configuration/DatabaseConfig.java
+++ b/data/config-schema/src/main/java/net/corda/schema/configuration/DatabaseConfig.java
@@ -10,4 +10,5 @@ public final class DatabaseConfig {
     public static final String JDBC_DRIVER_DIRECTORY = "database.jdbc.directory";
     public static final String JDBC_URL = "database.jdbc.url";
     public static final String DB_POOL_MAX_SIZE = "database.pool.max_size";
+    public static final String DB_POOL_MIN_SIZE = "database.pool.min_size";
 }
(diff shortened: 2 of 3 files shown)


data/config-schema/src/main/resources/net/corda/schema/configuration/db/1.0/corda.db.json
```diff
@@ -51,6 +51,19 @@
               "type": "integer",
               "minimum": 1,
               "default": 10
+            },
+            "min_size": {
+              "description": "The minimal database pool size.",
```
kyriathar: "minimal" -> "minimum"

data/config-schema/src/main/resources/net/corda/schema/configuration/db/1.0/corda.db.json
```diff
@@ -51,6 +51,19 @@
...
+              "default": null,
+              "anyOf": [
+                {
+                  "type": "integer",
+                  "minimum": 0
+                },
+                {
+                  "type": "null"
```
kyriathar, line 64: Why do we want this as optional parameter? Why not have it as not nullable with a defaulted value?

yift-r3: The default should be the maximum, but if I set it to 10, and someone puts the maximum as five and forget to set the min, it will be confusing...

kyriathar: Not really sure why it should be the maximum though? Maybe we should be adding a check to assert that minumValue <= maximumValue but that should be it?

yift-r3: That's what Hikari [recommends:](https://github.com/brettwooldridge/HikariCP) "... for maximum performance and responsiveness to spike demands, we recommend not setting this value and instead allowing HikariCP to act as a fixed size connection pool. Default: same as maximumPoolSize". I think they know what they are talking about.

data/config-schema/src/main/resources/net/corda/schema/configuration/db/1.0/corda.db.json
```diff
@@ -51,6 +51,19 @@
               "type": "integer",
               "minimum": 1,
               "default": 10
+            },
+            "min_size": {
+              "description": "The minimum database pool size.",
```
kyriathar: Maybe worth mentioning here that if left null then means minimumPoolSize = maximumPoolSize.

data/config-schema/src/main/resources/net/corda/schema/configuration/db/1.0/corda.db.json
dimosr: Our use case is slightly different to the typical use case of a web application that operates under a predictable workload and wants a fully initialised and always warm pool. I think actually a default of min pool size = 0 might be a more sensible default.

If we set the default of max pool size to 10 for example (looking at the config this seems to already be the case), then in a cluster that has 10 vnodes deployed, we'll be reserving up 100 connections by default even though we might need to use a much smaller number of them. This can be an issue in clusters that want to run in a cost-efficient way and with relatively small footprint at the database level.

yift-r3: This configuration is for the virtual node and the cluster databases...

For the node pools, the min size will be zero, so we will not reserve any connection (and we will evacuate any unused connection after 2 minutes).

dimosr: [Discussed a bit offline with Yiftach]

We currently have a single schema (and as a result a single default) for vnode and cluster DBs. That's a bit confusing and maybe something we want to change at some point in the future (not something for this PR), since each case has possibly different requirements. For example, for the cluster DB having min pool size = max pool size might make sense as a default, but for vnode DBs it probably doesn't as explained above.  

Now from a practical point of view, the associated runtime-os [PR](https://github.com/corda/corda-runtime-os/pull/3909) overrides the defaults for vnode DBs to set min pool size to 0. And it sets min pool size = max pool size for cluster DB connections, which is the current behaviour (and the default in the config). So, I think that's ok for now.

kyriathar (approved):

dimosr (approved):

data/config-schema/src/main/resources/net/corda/schema/configuration/db/1.0/corda.db.json
driessamyn: This is where this overlaps with some of the work that was in-flight.

We create the VNode connections, and they should be configured differently as they have very different usage patterns. Cluster DB connections may be ok with 5 connections at start-up, but the vnode ones only need 1 generally.

The proposal in the work that @kyriathar and @simon-johnson-r3 were going to pick up was for there to be 2 different sets of defaults, one for the general cluster connections, another a default for the DBAdmin module to use when _creating_ DB connections. 

So I don't think this change is actually correct.

PR #1131, opened 2023-05-19
CORE-11990 - Modify signature handling

nikinagy:
Review how we handle signed data for member info and registration request. We should store and retrieve the data in its original serialised form instead of serialising and deserialising back and forth.

As part of this I modified PersistentMemberInfo to hold the original byte arrays.

Added signedMembers to MemberInfoQueryResponse.avsc to retrieve the signature information along with member information and remove query member signature function call.


data/avro-schema/src/main/resources/avro/net/corda/data/membership/SignedContexts.avsc
```diff
@@ -0,0 +1,18 @@
+{
+  "type": "record",
+  "name": "SignedContexts",
```
charlieR3, line 3: This is named `SignedContexts` but it looks like just just two fields for serialised data and there is no signature information so I'm not sure that this is actually signed data.

data/avro-schema/src/main/resources/avro/net/corda/data/membership/PersistentMemberInfo.avsc
```diff
@@ -11,13 +11,19 @@
...
-      "doc": "MGM provided data in MemberInfo.",
+      "doc": "Deprecated field, signedData should be used instead. MGM provided data in MemberInfo.",
       "type": "net.corda.data.KeyValuePairList"
+    },
+    {
+      "name": "signedData",
+      "doc": "The original byte data part which was signed.",
+      "type": ["null", "net.corda.data.membership.SignedContexts"],
+      "default": null
     }
```
charlieR3, line 27: We have a SignedData schema so that might be better here. i.e. `net.corda.data.membership.SignedData`

e.g.
```suggestion
    {
      "name": "signedMemberContext",
      "doc": "Signed member provided data in MemberInfo.",
      "type": ["null", "net.corda.data.membership.SignedData"],
      "default": null
    },
    {
      "name": "signedMgmContext",
      "doc": "Signed MGM provided data in MemberInfo.",
      "type": ["null", "net.corda.data.membership.SignedData"],
      "default": null
    }
```

data/avro-schema/src/main/resources/avro/net/corda/data/membership/db/response/query/MemberInfoQueryResponse.avsc
```diff
@@ -11,6 +11,18 @@
...
       }
+    },
+    {
+      "name": "signedMembers",
+      "doc": "Member Infos with signature representing members found for a given query.",
+      "type": [
+        "null",
+        {
+          "type": "array",
+          "items": "net.corda.data.membership.PersistentSignedMemberInfo"
```
charlieR3, line 22: Can we merge the PersistenceMemberInfo and PersistenceSignedMemberInfo now that the key value pair lists are in their serialised forms?

data/avro-schema/src/main/resources/avro/net/corda/data/membership/SignedContexts.avsc
```diff
@@ -0,0 +1,18 @@
...
+  "doc": "Avro representation of the original byte data part which was signed in the MemberInfo.",
+  "fields": [
+    {
+      "name": "memberContext",
+      "doc": "Member provided data in MemberInfo serialised as byte array by using KeyValuePairList.",
+      "type": "bytes"
+    },
+    {
+      "name": "mgmContext",
+      "doc": "MGM provided data in MemberInfo serialised as byte array by using KeyValuePairList.",
```
charlieR3, line 14: Where is the signature for this context?

charlieR3 (commented):
I understand we need to jump through hoops a bit for backwards compatibility so we might be stuck but the schema structure is quite confusing. 

`MemberInfoQueryResponse` [has `PersistentSignedMemberInfo`]
  -> `PersistentSignedMemberInfo` [has `PersistentMemberInfo` and the member context signature]
    -> `PersistentMemberInfo` [has `SignedContexts`]
      -> `SignedContexts` [has unsigned member context and unsigned mgm context]

Ideally the MGM context signature would be present somewhere if possible and the signatures would be at the same level as the data which was signed over. 

If we could merge `PersistenceMemberInfo` and `PersistenceSignedMemberInfo` and use `SignedData` for the signed contexts I think it would make more sense and align better with the structure of `SignedGroupParameters` for example. With backwards compatibility support I'm not sure how possible that is.

data/avro-schema/src/main/resources/avro/net/corda/data/membership/PersistentMemberInfo.avsc
```diff
@@ -11,13 +11,19 @@
...
+      "doc": "Deprecated field, signedData should be used instead. Member provided data in MemberInfo.",
       "type": "net.corda.data.KeyValuePairList"
     },
     {
       "name": "mgmContext",
-      "doc": "MGM provided data in MemberInfo.",
+      "doc": "Deprecated field, signedData should be used instead. MGM provided data in MemberInfo.",
       "type": "net.corda.data.KeyValuePairList"
+    },
+    {
```
driessamyn, line 22: NOTE: AVRO schemas cannot be changed between versions!!

data/avro-schema/src/main/resources/avro/net/corda/data/membership/PersistentMemberInfo.avsc
nikinagy: While I understand that we have mgm signatures as well, MGM signature information is not stored in the database. So I don't really see how could I return the mgm context with the signature from the database when we are doing querying? PersistentMemberInfo resembles the db entry, but we don't have anything regarding mgm signature only the context.

data/avro-schema/src/main/resources/avro/net/corda/data/membership/db/response/query/MemberInfoQueryResponse.avsc
nikinagy: This was made as per the avro evolution guidelines: https://docs.confluent.io/platform/current/schema-registry/fundamentals/schema-evolution.html
I think we concluded to keep the old fields for a while and just deprecate them. So here PersistentMemberInfo is the old deprecated field, I just forgot to modify the docs and say it's been deprecated.

Seeing that the schemas can't be changed between versions, I think this evolution (and PR) will change completely. If totally new schemas are needed, then yes I am planning to just keep PersistentSignedMemberInfo.

data/avro-schema/src/main/resources/avro/net/corda/data/membership/SignedContexts.avsc
nikinagy: Yes, probably bad naming. I wanted to keep the original serialised data which has been used for signing.

data/avro-schema/src/main/resources/avro/net/corda/data/membership/SignedContexts.avsc
nikinagy: See my other comment for: https://github.com/corda/corda-api/pull/1131/files/4d64c47d4f14d69b292a8ea8b7705a0e9c76dac7#r1222728784

data/avro-schema/src/main/resources/avro/net/corda/data/membership/PersistentMemberInfo.avsc
simon-johnson-r3: When you change an avro schema a new fingerprint is generated for it (a fingerprint is unique to the schema definition in other words). An old version of Corda will already have published all avro schemas to a Kafka topic, keyed by their fingerprint. So if a new version of Corda is picking up an old message, it will be able to grab the old schema by fingerprint from that topic, and the new version of the schema from the compiled JAR, and that's all it needs to maintain backwards compatibility.

So you can change Avro schemas as long as they adhere to the standard Avro rules. I think Avro provide a class to test this: https://avro.apache.org/docs/current/api/java/org/apache/avro/SchemaCompatibility.html
In this case you would be reading messages with the new schema written with the old one.

Obviously your code needs to ensure it can cope with messages all of "versions" whatever that means for any particular message.

PR #1136, opened 2023-05-29
CORE-10453: Replace RemoteAliasLookUp with InteropIdentityLookup

jmacmahonr3:
An alternative more simplistic implementation of the InterOp equivalent of the MemberLookup. 

What's noteworthy about this new API is that the Application name is the sole identifier an the flow level.


interop/src/main/java/net/corda/v5/interop/InterOpIdentityInfo.java
```diff
@@ -0,0 +1,31 @@
...
+@CordaSerializable
+public interface InterOpIdentityInfo {
+    /**
+     * @return The {@link String} identifier for the InterOp Identity.
+     */
+    @NotNull
+    String applicationName();
+
+    /**
+     * @return The {@link String} of the Identities X500.
```
yarivus:
```suggestion
     * @return The {@link String} of the Identity's X500.
```

yarivus (changes requested):

yarivus (approved):
LGTM

interop/src/main/java/net/corda/v5/interop/InterOpIdentityInfo.java
```diff
@@ -0,0 +1,31 @@
...
+    @NotNull
+    String applicationName();
+
+    /**
+     * @return The {@link String} of the Identity's X500.
+     */
+    @NotNull
+    String getX500Name();
+
+
```
szymonsztuka: Remove a new line

interop/src/main/java/net/corda/v5/interop/InterOpIdentityInfo.java
```diff
@@ -0,0 +1,31 @@
+package net.corda.v5.interop;
+
+import net.corda.v5.base.annotations.CordaSerializable;
+import org.jetbrains.annotations.NotNull;
+import java.util.List;
+
+/**
+ * <p>This interface represents the alias member details like Application Name, facadeIds etc.
```
szymonsztuka: Very minor - I think we can remove "< p >" and extra empty line in that Javadoc

application/src/main/java/net/corda/v5/application/interop/InteropIdentityLookUp.java
```diff
@@ -0,0 +1,32 @@
...
+import net.corda.v5.application.interop.facade.FacadeId;
+import net.corda.v5.base.annotations.DoNotImplement;
+import net.corda.v5.base.annotations.Suspendable;
+import net.corda.v5.interop.InterOpIdentityInfo;
+import org.jetbrains.annotations.NotNull;
+
+import java.util.List;
+
+/**
+ * <p>This interface represents the injectable service for client corDapp to look up the remote alias identity.
```
szymonsztuka: We would need to clearly separate this service from any services returning normal member group, i would use explicit words like _Interoperability group_ etc. Also javadoc for other services doesn't mention `injectable`, all Corda API services are injectable.
Maybe something like that, which explicitly shows the different with existing `memberLookup`:
``The service allows a flow to retrieve alter-ego members from other networks, which are register in the holding identity's interoperability group.``

interop/src/main/java/net/corda/v5/interop/InterOpIdentityInfo.java
szymonsztuka: Maybe a bit later not for this PR/ so just suggestion, but the description doesn't match the interface name or doesn't refer: "Identity" or "InteroOp",  how "alias member" relates to "InterOpIdentity"?

application/src/main/java/net/corda/v5/application/interop/InteropIdentityLookUp.java
```diff
@@ -0,0 +1,32 @@
...
+    /**
+     * @return The {@link InterOpIdentityInfo} representing the alias member info for given alias x500name and host network.
+     */
+    @Suspendable
+    @NotNull
+    InterOpIdentityInfo lookup(String applicationName);
+
+    /**
+     * @param facadeId The facadeId of the facade.
+     * @return The {@link List< InterOpIdentityInfo >} representing the alias member info for given facadeId.
```
szymonsztuka: Remove spaces after `<` and before `>`

application/src/main/java/net/corda/v5/application/interop/InteropIdentityLookUp.java
```diff
@@ -0,0 +1,32 @@
...
+
+/**
+ * <p>This interface represents the injectable service for client corDapp to look up the remote alias identity.
+ *
+ */
+@DoNotImplement
+public interface InteropIdentityLookUp {
+
+    /**
+     * @return The {@link InterOpIdentityInfo} representing the alias member info for given alias x500name and host network.
```
szymonsztuka, line 24: This is a crucial parameter and it's not described, is this intentional?  
What is `applicationName`, how does it relates to  "x500name" and "host network".?

szymonsztuka (commented):
Needs to increment API version in gradle.properties.

application/src/main/java/net/corda/v5/application/interop/InteropIdentityLookUp.java
jmacmahonr3: Please check updated docs

yarivus (approved):
LGTM

szymonsztuka (approved):
Approving, the `cordaApiRevision` probably should be `7` and not `8`, but doesn't really matter.

PR #1143, opened 2023-06-07
CORE-13567 Add extra datasource properties to Db config schema

kyriathar:
- adds configuration for the following datasource properties:
  - `idleTimeout`
  - `maxLifetime`
  - `keepaliveTime`
  - `validationTimeout`

### DIFF
--- a/data/config-schema/src/main/resources/net/corda/schema/configuration/db/1.0/corda.db.json
+++ b/data/config-schema/src/main/resources/net/corda/schema/configuration/db/1.0/corda.db.json
@@ -66,19 +66,19 @@
               ]
             },
             "idleTimeout": {
-              "description": "The maximum time (in minutes) a connection can stay idle in the pool. A value of 0 means that idle connections are never removed from the pool.",
+              "description": "The maximum time (in seconds) a connection can stay idle in the pool. A value of 0 means that idle connections are never removed from the pool.",
               "type": "integer",
               "minimum": 0,
-              "default": 2
+              "default": 120
             },
             "maxLifetime": {
-              "description": "The maximum time (in minutes) a connection can stay in the pool, regardless if it is idle or not.",
+              "description": "The maximum time (in seconds) a connection can stay in the pool, regardless if it has been idle or has been recently used. If a connection is in-use and has reached \"maxLifetime\" timeout, it will be removed only when it becomes idle.",
               "type": "integer",
               "minimum": 1,
-              "default": 30
+              "default": 1800
             },
             "keepaliveTime": {
-              "description": "The interval time (in seconds) in which connections will be tested for aliveness, keeping them alive by the act of checking. A value of 0 means this is disabled.",
+              "description": "The interval time (in seconds) in which connections will be tested for aliveness. Connections which are no longer alive are removed from the pool. A value of 0 means this check is disabled.",
               "type": "integer",
               "minimum": 0,
               "default": 0
--- a/data/config-schema/src/main/resources/net/corda/schema/configuration/db/1.0/corda.db.json
+++ b/data/config-schema/src/main/resources/net/corda/schema/configuration/db/1.0/corda.db.json
@@ -66,23 +66,27 @@
               ]
             },
             "idleTimeout": {
-              "description": "The TODO time in minutes.",
+              "description": "The maximum time (in minutes) a connection can stay idle in the pool. A value of 0 means that idle connections are never removed from the pool.",
               "type": "integer",
+              "minimum": 0,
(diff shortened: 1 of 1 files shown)


data/config-schema/src/main/resources/net/corda/schema/configuration/db/1.0/corda.db.json
```diff
@@ -64,6 +64,30 @@
                   "type": "null"
                 }
               ]
+            },
+            "idleTimeout": {
+              "description": "The maximum time (in minutes) a connection can stay idle in the pool. A value of 0 means that idle connections are never removed from the pool.",
```
kyriathar: @simon-johnson-r3 I believe we don't want to keep the "_A value of 0 means that idle connections are never removed from the pool_" option in which case we should prevent it by setting `"minimum": 1`?

simon-johnson-r3: An interesting question, I'm not an expert on db connection tuning although I doubt we'd ever use it. That said I don't see keeping 0 as an option a problem myself.

data/config-schema/src/main/resources/net/corda/schema/configuration/db/1.0/corda.db.json
```diff
@@ -64,6 +64,30 @@
...
+              "type": "integer",
+              "minimum": 0,
```
simon-johnson-r3, line 71: We have a mixture of minutes and seconds here which strikes me as a bad idea. I expect this comes from the Hikari api? But in any case I think we should standardise on seconds.

data/config-schema/src/main/resources/net/corda/schema/configuration/db/1.0/corda.db.json
```diff
@@ -64,6 +64,30 @@
...
+              "default": 2
+            },
+            "maxLifetime": {
+              "description": "The maximum time (in minutes) a connection can stay in the pool, regardless if it is idle or not.",
```
simon-johnson-r3: I don't think this description is accurate, or at least it contradicts the Hikari data source factory which states
`An in-use connection will never be retired, only when it is closed will it then be removed.`
The description here reads like you would kick a non-idle connection out of the pool, is that your intention?

data/config-schema/src/main/resources/net/corda/schema/configuration/db/1.0/corda.db.json
```diff
@@ -64,6 +64,30 @@
...
+              "type": "integer",
+              "minimum": 1,
+              "default": 30
+            },
+            "keepaliveTime": {
+              "description": "The interval time (in seconds) in which connections will be tested for aliveness, keeping them alive by the act of checking. A value of 0 means this is disabled.",
```
simon-johnson-r3: instead of `keeping them alive by the act of checking` perhaps consider a new sentence along the lines of: `Connections which are no longer alive are removed from the pool`?

simon-johnson-r3 (changes requested):

data/config-schema/src/main/resources/net/corda/schema/configuration/db/1.0/corda.db.json
kyriathar: Yeah all these defaults and time scales were taken from `DataSourceFactory.create`. I do agree they should be standardized.

data/config-schema/src/main/resources/net/corda/schema/configuration/db/1.0/corda.db.json
kyriathar: Applied

data/config-schema/src/main/resources/net/corda/schema/configuration/db/1.0/corda.db.json
kyriathar: Yes you are right, good catch

kyriathar: Changed to say an in-use connection that has reached the timeout will get removed from pool when it gets **idle** instead of **closed** which is what the Hikari docs mention

data/config-schema/src/main/resources/net/corda/schema/configuration/db/1.0/corda.db.json
kyriathar: Changed them to be all seconds

simon-johnson-r3 (approved):

PR #1144, opened 2023-06-09
ES-89: removed obsolete parameter of shared pipeline

wzur-r3:
* removed obsolete `nexusAppId` parameter
* we no longer run checks with Nexus and the functionality is gone from shared pipelines
* on top of this `cordaPipelineKubernetesAgent` actively checks for non-existing parameters and it throws an exception when it finds one

This is just a cosmetic change and it does not modify any behaviour.


yarivus (approved):
LGTM

PR #1145, opened 2023-06-09
CORE-11037 Add constants for datasource properties config keys

kyriathar:
- adds constants for config keys for the following datasource  properties:
  - `idleTimeout`
  - `maxLifetime`
  - `keepaliveTime`
  - `validationTimeout`

### DIFF
--- a/data/config-schema/src/main/java/net/corda/schema/configuration/DatabaseConfig.java
+++ b/data/config-schema/src/main/java/net/corda/schema/configuration/DatabaseConfig.java
@@ -11,4 +11,8 @@ public final class DatabaseConfig {
     public static final String JDBC_URL = "database.jdbc.url";
     public static final String DB_POOL_MAX_SIZE = "database.pool.max_size";
     public static final String DB_POOL_MIN_SIZE = "database.pool.min_size";
+    public static final String DB_POOL_IDLE_TIMEOUT = "database.pool.idleTimeout";
+    public static final String DB_POOL_MAX_LIFETIME = "database.pool.maxLifetime";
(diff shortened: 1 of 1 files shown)


data/config-schema/src/main/java/net/corda/schema/configuration/DatabaseConfig.java
```diff
@@ -11,4 +11,8 @@ private DatabaseConfig() {
     public static final String JDBC_URL = "database.jdbc.url";
     public static final String DB_POOL_MAX_SIZE = "database.pool.max_size";
     public static final String DB_POOL_MIN_SIZE = "database.pool.min_size";
+    public static final String DB_POOL_IDLE_TIMEOUT = "database.pool.idleTimeout";
+    public static final String DB_POOL_MAX_LIFETIME = "database.pool.maxLifetime";
+    public static final String DB_POOL_KEEP_ALIVE_TIME = "database.pool.keepaliveTime";
```
simon-johnson-r3: as you've not capitalised the 'a' in keepalive, this should be`DB_POOL_KEEPALIVE_TIME`

simon-johnson-r3 (changes requested):

data/config-schema/src/main/java/net/corda/schema/configuration/DatabaseConfig.java
kyriathar: Good catch, addressed.

simon-johnson-r3 (approved):

driessamyn (approved):
LGTM

PR #1149, opened 2023-06-15
CORE-14564 - Review backward compatibility for RegistrationStatus

nikinagy:
Adding new values for enums in avro is not considered as a backward compatible change.
We’ve added a new value STARTED_PROCESSING_BY_MGM to flag in-progress registrations. This change unfortunately broke the registration when doing cross-version testing. This PR contains the fix for this issue.


dimosr (approved):
LGTM - some comments on the runtime-os only

PR #1153, opened 2023-06-20
CORE-14685: Add Corda-Api tag to all API artifacts.

chrisr3:
Add `Corda-Api` tag to every API jar that we build. The proposed value of this tag is
```
${cordaProductVersion}.${cordaApiRevision}
```
although my intended use-case is just to allow me to identify Corda's API jars for use in my "test driver".


yarivus (approved):
LGTM

PR #1154, opened 2023-06-20
CORE-14685: Use corda-api.common-library to build Avro bundle.

chrisr3:
Now that all Corda's APIs are written in Java, we can use the `corda-api.common-library` Gradle extension plugin to build the `avro-schema` bundle.

This allows us to remove some duplication from the `avro-schema` bundle's build script.


knguyen-r3 (approved):
LGTM

yarivus (commented):
thx @chrisr3

PR #1162, opened 2023-06-27
CORE-13098 - Sync with release branch

nikinagy:
Sync the re-registration feature branch with 5.1 release branch.

Resolved conflicts:
gradle.properties
Jenkinsfile
data/topic-schema/src/main/java/Schemas.java


gradle.properties
```diff
@@ -3,14 +3,13 @@
...
 org.gradle.java.installations.auto-download=false
-org.gradle.jvmargs=-Dfile.encoding=UTF-8
 
 # Versioning
 cordaProductVersion = 5.1.0
 # NOTE: update this each time this module contains a breaking change
 ## NOTE: currently this is a top level revision, so all API versions will line up, but this could be moved to
 ##   a per module property in which case module versions can change independently.
-cordaApiRevision = 747-MGM
+cordaApiRevision = 5-MGM
```
dimosr, line 12: that looks a bit strange to me - how come we're decreasing the version?

Is it the case it was reset down to 1 (due to the bump of the major version to 5.1) at some point and we hadn't synced that yet? Or something else?

dimosr (commented):
LGTM - just a minor question.

Also, just for the future - providing the patch files that contain the diffs of the files that were left for conflict resolution after trying the merge can make reviewing this even easier (especially if you are using 3-way merge) because the reviewer can have open the PR diff in one window and the patch file in the other and see very quickly how conflicts were resolved and what was the state in each branch.

gradle.properties
nikinagy: Yes, it was reset : https://github.com/corda/corda-api/commit/6028bb677d518f3cfbcc4f8dda7a917fe0a85bda
I guess we haven't had that yet or kept using 74+-MGM when we did the last sync.

dimosr (approved):

PR #1163, opened 2023-06-27
CORE-14355 Interop / Rename Facade Service method

szymonsztuka:
FacadeService is returning not the Facade but the proxy binding it to TokensFacade. Rename the method to "getProxy".

 Matching PR in corda-runtime-os : https://github.com/corda/corda-runtime-os/pull/4168


julia-filipczak (approved):

yarivus (commented):
thx @szymonsztuka

PR #1170, opened 2023-07-05
CORE-14484 Implement interop group management

ac101m:
# Changes
- Introduce `holdingIdentityShortHash` into interop identity avro schema


data/avro-schema/src/main/resources/avro/net/corda/data/interop/InteropIdentity.avsc
```diff
@@ -15,6 +15,11 @@
       "doc": "Interop identity x500 name.",
       "name": "x500Name",
       "type": "string"
+    },
+    {
+      "doc": "Short hash of holding identity to which this interop identity corresponds.",
+      "name": "holdingIdentityShortHash",
```
szymonsztuka: If this is a value to distinguish whenever it is a) my alias versus b) an alias to someone I see, than this should be just a boolean flag. For aliases which are not "mine" ( b) ) the field is null, which is (n-1)*100/n % of times, where n is number of entries in the interop group.

ac101m: I strongly disagree. If this object is received outside the context of a holding identity or virtual node, then an "is my holding identity" flag loses all meaning. If we need to get "our" interop identity within a group, then the bookkeeping for that should be added to the new class I introduced into the cache.

szymonsztuka: but the record on Kafka topic is always under the key, this Avro type is only "Transport Object", when you read off the topic then you can create more proper business domain object.

ac101m: Oh I see, just for the avro type. I was imagining this being applied throughout, yeah I suppose that makes sense. Will update this.

yarivus (approved):
LGTM

jmacmahonr3 (approved):
LGTM

PR #1180, opened 2023-07-17
CORE-10114 Schema update for custom metadata

YashNabar:
Update static registration context and group policy schemas to allow custom properties to be added to the MemberInfo and group parameters.
https://github.com/corda/corda-runtime-os/pull/4330

### DIFF
--- a/data/membership-schema/src/main/resources/net/corda/schema/membership/group/policy/1.0/corda.group.policy.json
+++ b/data/membership-schema/src/main/resources/net/corda/schema/membership/group/policy/1.0/corda.group.policy.json
@@ -263,8 +263,18 @@
             "groupParameters": {
               "description": "Custom key-value properties to be included in the group parameters of a static network.",
               "type": "object",
-              "additionalProperties": {
-                "type": "string"
+              "patternProperties": {
+                "^corda.minimum.platform.version$": {
+                  "description": "Optional. The minimum platform version of the static network.",
+                  "type": "integer",
+                  "minimum": 50000,
+                  "maximum": 99999
+                },
+                "^ext\\.[a-zA-Z0-9.]+$": {
+                  "description": "Optional. User-specified additional properties, which must be prefixed with `ext.`.",
+                  "type": "string",
+                  "maxLength": 256
+                }
               }
             }
           }
--- a/data/membership-schema/src/main/resources/net/corda/schema/membership/group/policy/1.0/corda.group.policy.json
+++ b/data/membership-schema/src/main/resources/net/corda/schema/membership/group/policy/1.0/corda.group.policy.json
@@ -259,6 +259,13 @@
                 "description": "A member template as a map of values for a member to be added to the static network member list.",
                 "type": "object"
               }
+            },
+            "groupParameters": {
+              "description": "Custom key-value properties to be included in the group parameters of a static network.",
+              "type": "object",
+              "additionalProperties": {
+                "type": "string"
+              }
             }
           }
         }
--- a/data/membership-schema/src/main/resources/net/corda/schema/membership/group/policy/1.0/corda.group.policy.json
+++ b/data/membership-schema/src/main/resources/net/corda/schema/membership/group/policy/1.0/corda.group.policy.json
@@ -264,16 +264,18 @@
               "description": "Custom key-value properties to be included in the group parameters of a static network.",
               "type": "object",
               "patternProperties": {
-                "^corda.minimum.platform.version$": {
-                  "description": "Optional. The minimum platform version of the static network.",
-                  "type": "integer",
-                  "minimum": 50000,
-                  "maximum": 99999
-                },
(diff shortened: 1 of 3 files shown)


data/membership-schema/src/main/resources/net/corda/schema/membership/member/static/registration/1.0/corda.member.static.registration.json
```diff
@@ -17,6 +17,11 @@
       "description": "Version supported for the flow protocol used by the notary service. Valid only when one of the roles is notary.",
       "type": "integer",
       "minimum": 1
+    },
+    "^ext\\.[a-zA-Z0-9.]+$": {
+      "description": "Optional. A user can specify additional properties, they must be prefixed with `ext.`.",
```
charlieR3: Maybe either
`Optional. A user can specify additional properties, which must be prefixed with `ext.`.`
or
`Optional. A user can specify additional properties. They must be prefixed with `ext.`.`

data/membership-schema/src/main/resources/net/corda/schema/membership/group/policy/1.0/corda.group.policy.json
```diff
@@ -259,6 +259,13 @@
                 "description": "A member template as a map of values for a member to be added to the static network member list.",
                 "type": "object"
               }
+            },
+            "groupParameters": {
```
charlieR3, line 263: Should `corda.minimum.platform.version` be explicitly added here or is it intentionally left out?

data/membership-schema/src/main/resources/net/corda/schema/membership/member/static/registration/1.0/corda.member.static.registration.json
```diff
@@ -17,6 +17,11 @@
...
+      "type": "string",
+      "maxLength": 256
```
dimosr, line 24: I think one requiremenent we have for this is users of this feature are going to be using this for building permissioning schemes based off cryptographic keys, right? Are they going to include PEM-formatted public keys in this attributes (either in member info or in the group policy) and do we need to relax these limitations a bit if that's the case?

I think this might already be in place for member info custom metadata, so we could have a separate ticket to adjust this globally. But, worth confirming now if it's needed or not.

data/membership-schema/src/main/resources/net/corda/schema/membership/group/policy/1.0/corda.group.policy.json
YashNabar: We were relying on the verification done during parsing, but would be good to have it in the schema. Added in [592198a](https://github.com/corda/corda-api/pull/1180/commits/592198a818280b06d34d277d8cbc25ff56cf03f4)

data/membership-schema/src/main/resources/net/corda/schema/membership/member/static/registration/1.0/corda.member.static.registration.json
YashNabar: Yes would be good to have a standard limit for all fields where we expect PEM input. I noticed we may either have limits on the lower side (e.g. custom metadata) or have no limits set (e.g. MGM registration schema)

dimosr: In that case, can you raise a ticket under that epic to come back and bump all the defaults to account for this?

To get to a reasonable number, some calculations/research is needed to find out roughly what's a size of a PEM-formatted key for the different key types we support (e.g. RSA/ECDSA etc.) and add some extra headroom on top of that. 

P.S.: The limits for member side were put in place mostly for DoS protection AFAIK. No harm adding some limits for the group parameters custom metadata on the MGM side if none are present, although possibly not that critical given it's the network operator so effectively trusted by everyone.

YashNabar: I've raised a ticket [here](https://r3-cev.atlassian.net/browse/CORE-15641)

data/membership-schema/src/main/resources/net/corda/schema/membership/group/policy/1.0/corda.group.policy.json
```diff
@@ -259,6 +259,23 @@
...
+              "description": "Custom key-value properties to be included in the group parameters of a static network.",
+              "type": "object",
+              "patternProperties": {
+                "^corda.minimum.platform.version$": {
+                  "description": "Optional. The minimum platform version of the static network.",
+                  "type": "integer",
+                  "minimum": 50000,
+                  "maximum": 99999
+                },
```
charlieR3: I think this doesn't have to be defined as a regex in `patternProperties`

charlieR3 (approved):

PR #1183, opened 2023-07-20
CORE-11805 - Update json schema spec

dimosr:
Upgrade json schema spec to 2019-09. This is pre-requisite in order to be aligned with the latest spec version specified when building `JsonSchemaFactory` as used by https://github.com/corda/corda-runtime-os/pull/4341


vkolomeyko (approved):
LGTM

PR #1188, opened 2023-07-25
CORE-12783: Add topics and AVRO messages for EVM operations 

fowlerrr:
Added new topics and Avro messages for communication between the flow and an EVM processor


application/src/main/java/net/corda/v5/application/interop/evm/TransactionReceipt.java
```diff
@@ -0,0 +1,191 @@
...
+    private String to;
+    private String from;
+    private String contractAddress;
+    private String cumulativeGasUsed;
+    private String gasUsed;
+    private String effectiveGasPrice;
+    private String root;
+    private String status;
+    private String revertReason;
+    private String type;
```
yarivus: comments to each field may help

yarivus (commented):
some comments

application/src/main/java/net/corda/v5/application/interop/evm/EvmService.java
```diff
@@ -0,0 +1,201 @@
...
+    @Suspendable
+    @NotNull
+    Transaction getTransactionByHash(
+            @NotNull String hash
+    );
+
+    /**
+     * Sends a <a href="https://www.quicknode.com/docs/ethereum/eth_getTransactionCount">eth_getTransactionCount</a> query to the EVM.
+     *
+     * @param address the address to check for balance
```
eierina: not balance but tx count

application/src/main/java/net/corda/v5/application/interop/evm/EvmService.java
```diff
@@ -0,0 +1,201 @@
...
+    @Suspendable
+    int getBalance(
+            @NotNull String address,
+            @NotNull String blockNumber
+    );
+
+    /**
+     * Sends a <a href="https://www.quicknode.com/docs/ethereum/eth_getCode">eth_getCode</a> query to the EVM.
+     *
+     * @param address the address to check for balance
```
eierina: not balance but contract to get code for

application/src/main/java/net/corda/v5/application/interop/evm/EvmService.java
```diff
@@ -0,0 +1,201 @@
...
+     * Sends a <a href="https://www.quicknode.com/docs/ethereum/eth_call">eth_call</a> query to the EVM.
+     *
+     * @param functionName    the function name on the EVM smart contract to call
+     * @param contractAddress the address of the smart contract
+     * @param parameters      any parameters (in order listed by the call) for the contract
+     * @return the returned value of the contract query
+     */
+    @Suspendable
+    @NotNull
+    Object call(@NotNull String contract,
```
eierina: @fowlerrr  what is `contract` here? The @param description is missing.

application/src/main/java/net/corda/v5/application/interop/evm/EvmService.java
```diff
@@ -0,0 +1,201 @@
...
+@DoNotImplement
+public interface EvmService {
+
+    /**
+     * Sends a <a href="https://www.quicknode.com/docs/ethereum/eth_call">eth_call</a> query to the EVM.
+     *
+     * @param functionName    the function name on the EVM smart contract to call
+     * @param contractAddress the address of the smart contract
+     * @param parameters      any parameters (in order listed by the call) for the contract
+     * @return the returned value of the contract query
```
eierina: I think response or values may be better than value?

application/src/main/java/net/corda/v5/application/interop/evm/EvmService.java
fowlerrr: Ah, I didn't put that in the docs.  It should maybe be `contractName`.  It's basically how we know what the function we're going to encode is.  Ideally it would be something like

```
myContract.itsFunction()
```

but since we don't have autogeneration (yet) it's more like

```
evmService.call("myContract", "itsFunction)
```

I'll change the parameter to `contractName` and add the javadoc

application/src/main/java/net/corda/v5/application/interop/evm/EvmService.java
```diff
@@ -0,0 +1,201 @@
...
+
+    /**
+     * Sends a <a href="https://www.quicknode.com/docs/ethereum/eth_getTransactionCount">eth_getTransactionCount</a> query to the EVM.
+     *
+     * @param address the address to check for balance
+     * @param blockNumber the block number in hexadecimal format or the string "latest", "earliest", "pending", "safe" or "finalized"
+     * @return the number of transactions sent from the address
+     */
+    @Suspendable
+    int getTransactionCount(
```
eierina: I suppose this call hides the logic to things such nextNonce = currentNonce + pendingTransactions.length;

application/src/main/java/net/corda/v5/application/interop/evm/EvmService.java
eierina: @fowlerrr I see, but is contractName a required param at all? A contract is identified by its deploymnt address, and the call's "entry point" by the hash of the function name and parameter types only.

application/src/main/java/net/corda/v5/application/interop/evm/EvmService.java
eierina: @fowlerrr would be best to think for a moment at the case where, for some reasons, we end up with a nonce gap in the mempool. Imagine for example in the mempool there are 3 pending transactions with nonces 101, 102, 104. Since 103 is missing, the tx with nonce 104 will never be picked up, unless a new tx with nonce 103 is sent to the mempool. And once 101, 102 will be sent, the next nonce will be 103, but since there is a pending transction count of 1 nextNonce = 103 + 1 = 104 which will keep failing when trying to send to the mempool.

A gap can happen (both because of this single application failing, or because some other application/wallet competing with transactions signed by the same private key) and we need to have a way to recover the situation.

application/src/main/java/net/corda/v5/application/interop/evm/EvmService.java
fowlerrr: It's required by _us_ to know what function we're encoding.  Otherwise we can't properly encode its parameters.  Basically we'll use those first two parameters to find the matching `functionName` in `contractName`s ABI.  Then use that information from the ABI to encode the parameters correctly for sending to the external worker.

application/src/main/java/net/corda/v5/application/interop/evm/EvmService.java
fowlerrr: Definitely worth us being aware of.  Though I think this will just be a direct call to the node.  So it'd really just be "currentNonce".  We'd leave the rest to the flow developer, right @ILoomans ?

eierina: @fowlerrr Not sure what you mean, the scenario that I describe happens in the context of "direct call to the node". At the flow level you cannot control this as it only can be controlled just before the transaction is signed and sent (the nonce is part of the non-signed transaction), so one of the problems we need to be able to solve is to be able to queue unsigned transactions and just set the nonce right before signing it and right before sending it.

And when operating with EVM networks we need to build for an adversary environment. If it can happen, it will happen.

application/src/main/java/net/corda/v5/application/interop/evm/EvmService.java
eierina: @fowlerrr I see, so contrat name is the lookup key for contracts ABI and the function name is the lookup key for the function in the matching contract. 

if you have a contract like the following

```solidity
contract MyContract {

    function myFunction(uint256 x) external {
        ...
    } 

    function myFunction(bytes calldata b) external {
        ...
    } 
```

By using only the function name you will have 2 values for the same keys (myFunction).
To call the first myFunction you really only need the contract deployment address and the entry point which is calculated as entryPoint = keccak256("myFunction(uint256)");

And one more doubt. It seems like the interface/EvmService works on top of the contracts, given that you can lookup the ABI by contract name. That would be a bit strange because all the ABI should be available/know above this interface (not under) so that you can produce the data to pass to this interface.

application/src/main/java/net/corda/v5/application/interop/evm/EvmService.java
fowlerrr: Ah, I see.  But that logic is handled in the `sendRawTransaction` code on the processor.  Though you raise a point that this is now confusing.  We'll handle getting the nonce correct via `sendRawTransaction` in order to limit the number of times the flow suspends.

It does raise the point of `getTransactionCount`, though, and if we need it.  Is there a use case for it outside of setting the nonce?

application/src/main/java/net/corda/v5/application/interop/evm/EvmService.java
fowlerrr: But then you would have to rewrite all of the contracts that might exist.  The point here is to allow for a more generic approach.  If someone wanted to add a contract they'd just provide the ABI and that would be enough.

It's not ideal but it reduces the burden on them from having to write a Java version (especially since we can't use Web3j to generate the contract for us).

Eventually we'd prefer to have our own autogeneration, but that'll take more time than we have right now.

application/src/main/java/net/corda/v5/application/interop/evm/EvmService.java
eierina: @fowlerrr 
I would leave the getTransactionCount, but I think I either don't understand something in the planned design or I did not explain well my concern.

some of the immutable assumptions are:
- sendRawTransaction  takes the signed transaction as input, therefore at that point the nonce was extracted, used to build the transaction, signed within the transaction.
- mempool fail/refuses to queue a transaction if a signed transaction is sent to the mempool with a nonce that is already in the mempool for that signer
- miners won't mine a signed transaction from the memory pool if there is a nonce gap before it (e.g. nonce 101, nonce 102, nonce 104 => the tx with nonce 104 will never be mined if you don't submit a "patch" transaction to fill the 103 gap, but the patch transaction will also execute before 104 even if it was prepared days or months after the 104 was submitted to the pool.
- you cannot delete transactions form the mempool

So all this is to say that _*flows cannot sign transactions*_. Probably for more than one reason, but mainly because of the possibility of race conditions between flows which are not guaranteed to execute in order of arrival and can be suspended any time and in essence A can start before B but B can end before A.

The only way you can ensure flows can sign transactions in a safe way is that flows need to only submit transaction data to a queue. The queue itself signs the transactions in a serialized fashion, and:
- only set the transaction nonce just before signing and submitting to the pool
- only process the next transaction once the previous has been accepted by the pool
- fill gaps between transactions if a gap is found. But since this causes transaction reordering should be applied conditionally where one of the conditions is `the transaction is coming from different flows`, but there may be cases where neither "different flows" may be acceptable if that causes transactions reordering.

Adding @r3domfox just as an extra eye in case my explanation is not clear or does not make sense.

fowlerrr: I think the problem is in our API.  It seems like you're expecting this to be _exactly_ as if you were calling out via Web3j.  This is actually just a proxy call to the `EvmProcessor` which will handle all of this work for you.  So I think, in trying to make it look more like you'd use as Web3j, we've created the expectation that, well, _you'd use it like Web3j_!

Which, I now think, is clearly wrong.  Because you should *not* be expecting the flow itself to sign the transaction.  Nor to determine the nonce.  It wouldn't make any sense for the flow to do that when it can't directly access the EVM in the first place.  We mean for all of that work to be handled outside of the flow and you should just want to do a simple `call` on a command (ie. something like `IERC20.balanceOf`) or a transaction (ie like `IERC20.transfer`).

But the simple fact that you think you'll be calculating the nonce in the flow (or adding it to the transaction yourself) means this isn't right.  Just to be clear, this isn't me pointing a finger at you, but at myself for making the API confusing.  But thank you because that's exactly what I needed to know from this review.

I'll need to think more about the API and how to make it simpler.  And remove your expectations for doing "too much".  I suspect we've given too many options and maybe it should be as simple as just doing the `call()` with some contract and everything else is too confusing.  But I'll go over it with @ILoomans once he's free.

r3domfox: Just to make sure I understand: the sequence of nonces is associated with a given account, not global across the whole ledger? So if I am signerA, I must submit my transactions with ascending nonce sequence numbers for the mempool to accept them, and therefore all transactions submitted across various flows have to be placed in some serial order prior to transmission? This also presumably means that the flow shouldn't be trying to decide the nonce number, as it can't know what other flows are simultaneously trying to submit transactions?

eierina: @r3domfox that is a good summary of the problem, yes.

PR #1199, opened 2023-08-03
CORE-15778: add meaningful exception messages in StateRef.parse()

jennyang-r3:
### Overview
The PR 
- added an exception in malformed value in StateRef where delimiter, ":", isn't in value.
- optimise the method by utilising an existing index exception.

### Testing
I added a unit test for StateRef.parse() method, testing if:
- valid StateRef is parsed correctly
- value with zero delimiter throws a correct exception
- value with missing index throws a correct exception
- any other malformed value throws transaction Id exception

### DIFF
--- a/ledger/ledger-utxo/src/main/java/net/corda/v5/ledger/utxo/StateRef.java
+++ b/ledger/ledger-utxo/src/main/java/net/corda/v5/ledger/utxo/StateRef.java
@@ -68,20 +68,20 @@ public final class StateRef {
      * @return Returns a {@link StateRef} parsed from the specified {@link String} value.
      * @throws IllegalArgumentException if the specified value cannot be parsed.
      */
-    public static StateRef parse(@NotNull final String value, DigestService digestService) {
-        try {
-            if (value.replaceAll("[^" + DELIMITER + "]", "").length() != 2) {
-                throw new IllegalArgumentException(
-                        MessageFormat.format("Failed to parse a StateRef from the specified value. The number of delimiter ({0}) should be two in value: {1}.", DELIMITER, value)
-                );
-            }
 
-            final int lastIndexOfDelimiter = value.lastIndexOf(DELIMITER);
+    public static StateRef parse(@NotNull final String value, DigestService digestService) {
+        final int lastIndexOfDelimiter = value.lastIndexOf(DELIMITER);
+        if (lastIndexOfDelimiter == -1) {
+            throw new IllegalArgumentException(
+                    MessageFormat.format("Failed to parse a StateRef from the specified value. At least one delimiter ({0}) is expected in value: {1}.", DELIMITER, value)
+            );
+        }
+
+        try {
             final String subStringBeforeDelimiter = value.substring(0, lastIndexOfDelimiter);
             final String subStringAfterDelimiter = value.substring(lastIndexOfDelimiter + 1);
             final int index = Integer.parseUnsignedInt(subStringAfterDelimiter);
             final SecureHash transactionId = digestService.parseSecureHash(subStringBeforeDelimiter);
-
             return new StateRef(transactionId, index);
         } catch (NumberFormatException numberFormatException) {
             throw new IllegalArgumentException(
@@ -89,6 +89,7 @@ public final class StateRef {
                     numberFormatException
             );
         } catch (IllegalArgumentException illegalArgumentException) {
+            System.out.println("another transaction id invalid exception");
             throw new IllegalArgumentException(
                     MessageFormat.format("Failed to parse a StateRef from the specified value. The transaction ID is malformed: {0}.", value),
                     illegalArgumentException
--- a/ledger/ledger-utxo/src/main/java/net/corda/v5/ledger/utxo/StateRef.java
+++ b/ledger/ledger-utxo/src/main/java/net/corda/v5/ledger/utxo/StateRef.java
@@ -89,7 +89,6 @@ public final class StateRef {
                     numberFormatException
             );
         } catch (IllegalArgumentException illegalArgumentException) {
-            System.out.println("another transaction id invalid exception");
             throw new IllegalArgumentException(
                     MessageFormat.format("Failed to parse a StateRef from the specified value. The transaction ID is malformed: {0}.", value),
                     illegalArgumentException
--- a/ledger/ledger-utxo/src/main/java/net/corda/v5/ledger/utxo/StateRef.java
+++ b/ledger/ledger-utxo/src/main/java/net/corda/v5/ledger/utxo/StateRef.java
@@ -68,7 +68,6 @@ public final class StateRef {
      * @return Returns a {@link StateRef} parsed from the specified {@link String} value.
      * @throws IllegalArgumentException if the specified value cannot be parsed.
      */
-
     public static StateRef parse(@NotNull final String value, DigestService digestService) {
         final int lastIndexOfDelimiter = value.lastIndexOf(DELIMITER);
         if (lastIndexOfDelimiter == -1) {
--- a/ledger/ledger-utxo/src/main/java/net/corda/v5/ledger/utxo/StateRef.java
+++ b/ledger/ledger-utxo/src/main/java/net/corda/v5/ledger/utxo/StateRef.java
@@ -19,10 +19,6 @@ public final class StateRef {
      * Specifies the delimiter that separates transaction ID and output index.
      */
     private static final String DELIMITER = ":";
-    /**
-     * The minimum SHA-256D algorithm hash length including SHA-256D: prefix
-     */
-    private static final Integer MINIMUM_HASH_LENGTH = 73;
 
     /**
      * The ID of the transaction in which the referenced state was created.
@@ -74,21 +70,17 @@ public final class StateRef {
      */
     public static StateRef parse(@NotNull final String value, DigestService digestService) {
         try {
-            if (!value.contains(DELIMITER)) {
+            final int lastIndexOfDelimiter = value.lastIndexOf(DELIMITER);
+            if (lastIndexOfDelimiter == -1) {
                 throw new IllegalArgumentException(
-                        MessageFormat.format("Failed to parse a StateRef from the specified value. The value: {0} doesn't contain expected delimiter (:)", value)
-                        );
-            }
-            if (value.length() < MINIMUM_HASH_LENGTH) {
-                throw new IllegalArgumentException(
-                        MessageFormat.format("Failed to parse a StateRef from the specified value. The length of the value: {0} doesn't meet the required SHA-256D algorithm length: 73", value.length())
+                        MessageFormat.format("Failed to parse a StateRef from the specified value. The delimiter (:) is missing: {0}.", value)
                 );
             }
-            final int lastIndexOfDelimiter = value.lastIndexOf(DELIMITER);
             final String subStringBeforeDelimiter = value.substring(0, lastIndexOfDelimiter);
             final String subStringAfterDelimiter = value.substring(lastIndexOfDelimiter + 1);
+            final int index = Integer.parseUnsignedInt(subStringAfterDelimiter);
             final SecureHash transactionId = digestService.parseSecureHash(subStringBeforeDelimiter);
-            final int index = Integer.parseInt(subStringAfterDelimiter);
+
             return new StateRef(transactionId, index);
         } catch (NumberFormatException numberFormatException) {
             throw new IllegalArgumentException(
--- a/ledger/ledger-utxo/src/main/java/net/corda/v5/ledger/utxo/StateRef.java
+++ b/ledger/ledger-utxo/src/main/java/net/corda/v5/ledger/utxo/StateRef.java
@@ -70,12 +70,13 @@ public final class StateRef {
      */
     public static StateRef parse(@NotNull final String value, DigestService digestService) {
         try {
-            final int lastIndexOfDelimiter = value.lastIndexOf(DELIMITER);
-            if (lastIndexOfDelimiter == -1) {
+            if (value.replaceAll("[^" + DELIMITER + "]", "").length() != 2) {
                 throw new IllegalArgumentException(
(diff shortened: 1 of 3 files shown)


ledger/ledger-utxo/src/main/java/net/corda/v5/ledger/utxo/StateRef.java
```diff
@@ -70,6 +74,16 @@ public int getIndex() {
      */
     public static StateRef parse(@NotNull final String value, DigestService digestService) {
         try {
+            if (!value.contains(DELIMITER)) {
+                throw new IllegalArgumentException(
+                        MessageFormat.format("Failed to parse a StateRef from the specified value. The value: {0} doesn't contain expected delimiter (:)", value)
+                        );
```
relyafi: Think indentation is wrong?

ledger/ledger-utxo/src/main/java/net/corda/v5/ledger/utxo/StateRef.java
```diff
@@ -19,6 +19,10 @@ public final class StateRef {
      * Specifies the delimiter that separates transaction ID and output index.
      */
     private static final String DELIMITER = ":";
+    /**
+     * The minimum SHA-256D algorithm hash length including SHA-256D: prefix
+     */
+    private static final Integer MINIMUM_HASH_LENGTH = 73;
```
relyafi: This seems problematic, because it assumes we're dealing with a SHA-256D hash? True now, but Corda 5 will support multiple hash algorithms at some point.

ledger/ledger-utxo/src/main/java/net/corda/v5/ledger/utxo/StateRef.java
```diff
@@ -70,6 +74,16 @@ public int getIndex() {
      */
     public static StateRef parse(@NotNull final String value, DigestService digestService) {
         try {
+            if (!value.contains(DELIMITER)) {
+                throw new IllegalArgumentException(
+                        MessageFormat.format("Failed to parse a StateRef from the specified value. The value: {0} doesn't contain expected delimiter (:)", value)
+                        );
+            }
+            if (value.length() < MINIMUM_HASH_LENGTH) {
```
relyafi: Hmm, this is really doing validation of the `SecureHash` rather than the `StateRef`. I wonder if we should actually be improving the validation within `SecureHash.parse` instead?

That said, based on my earlier comment, it's not clear if we have a stable minimum hash length we can rely on (regardless of algorithm), so do we need this check at all?

ledger/ledger-utxo/src/main/java/net/corda/v5/ledger/utxo/StateRef.java
```diff
@@ -70,6 +74,16 @@ public int getIndex() {
      */
     public static StateRef parse(@NotNull final String value, DigestService digestService) {
         try {
+            if (!value.contains(DELIMITER)) {
```
relyafi: This check doesn't seem quite right; we are already doing a delimiter check within `SecureHash.parse`. A State ref takes the form <Algorithm>:<Hash>:<State Index>. It's really the last bit that is `StateRef` specific - that the end of the string is the delimiter followed by a >= 0 integer.

relyafi (commented):
I've posted some specific comments, but having looked at the change as a whole, I wonder if there's a simpler solution. We're already parsing the hash, and verifying the index, so could the issue be solved by parsing the `index` before calling `parseSecureHash`? The index check could also probably be made a bit stronger by replacing the (now deprecated?) `parseInt` with `toUInt`?

I also think it's worth elaborating on the description, and adding test notes - @nkovacsx PR's are often good examples of this, e.g. https://github.com/corda/corda-runtime-os/pull/4230

Finally, it feels like we're currently missing a unit test case if we were able to make this change without breaking any tests - I think it's worth adding a case to check that we get the expected error if a malformed state ref is passed in.

ledger/ledger-utxo/src/main/java/net/corda/v5/ledger/utxo/StateRef.java
```diff
@@ -74,21 +70,17 @@ public int getIndex() {
...
     public static StateRef parse(@NotNull final String value, DigestService digestService) {
         try {
-            if (!value.contains(DELIMITER)) {
-                throw new IllegalArgumentException(
-                        MessageFormat.format("Failed to parse a StateRef from the specified value. The value: {0} doesn't contain expected delimiter (:)", value)
-                        );
-            }
-            if (value.length() < MINIMUM_HASH_LENGTH) {
+            final int lastIndexOfDelimiter = value.lastIndexOf(DELIMITER);
+            if (lastIndexOfDelimiter == -1) {
```
relyafi: I don't think this does quite the right thing - a `SecureHash` contains a delimiter, but would not be a valid `StateRef`.

Technically if we did not have this check, then I think it would fail the index parsing anyway (since it would try parsing a hash, not an int). But if you want a slightly better error message, then we should probably check for exactly 2 instances of the delimiter - the first would be part of the `SecureHash`, the second would separate this from the index.

ledger/ledger-utxo/src/main/java/net/corda/v5/ledger/utxo/StateRef.java
```diff
@@ -74,21 +70,17 @@ public int getIndex() {
      */
     public static StateRef parse(@NotNull final String value, DigestService digestService) {
         try {
-            if (!value.contains(DELIMITER)) {
-                throw new IllegalArgumentException(
-                        MessageFormat.format("Failed to parse a StateRef from the specified value. The value: {0} doesn't contain expected delimiter (:)", value)
-                        );
-            }
-            if (value.length() < MINIMUM_HASH_LENGTH) {
+            final int lastIndexOfDelimiter = value.lastIndexOf(DELIMITER);
```
relyafi: Unnecessary variable as it's only used once.

ledger/ledger-utxo/src/main/java/net/corda/v5/ledger/utxo/StateRef.java
```diff
@@ -74,21 +70,17 @@ public int getIndex() {
...
+            if (lastIndexOfDelimiter == -1) {
                 throw new IllegalArgumentException(
-                        MessageFormat.format("Failed to parse a StateRef from the specified value. The length of the value: {0} doesn't meet the required SHA-256D algorithm length: 73", value.length())
+                        MessageFormat.format("Failed to parse a StateRef from the specified value. The delimiter (:) is missing: {0}.", value)
```
relyafi: I would probably use the `DELIMITER` constant in this message, as it guards against changing the character in future.
```suggestion
                        MessageFormat.format("Failed to parse a StateRef from the specified value. The delimiter ($DELIMITER) is missing: {0}.", value)
```

relyafi: (Just realised the suggestion is Kotlin, not Java, but you get the idea :))

ledger/ledger-utxo/src/main/java/net/corda/v5/ledger/utxo/StateRef.java
jennyang-r3: Isn't this variable used in three places? 
``` kotlin
final int lastIndexOfDelimiter = value.lastIndexOf(DELIMITER);
            if (lastIndexOfDelimiter == -1) {
                throw new IllegalArgumentException(
                        MessageFormat.format("Failed to parse a StateRef from the specified value. The delimiter (:) is missing: {0}.", value)
                );
            }
            final String subStringBeforeDelimiter = value.substring(0, lastIndexOfDelimiter);
            final String subStringAfterDelimiter = value.substring(lastIndexOfDelimiter + 1);
```

relyafi: Yes, apparently I'm blind!

ledger/ledger-utxo/src/main/java/net/corda/v5/ledger/utxo/StateRef.java
```diff
@@ -70,11 +70,18 @@ public int getIndex() {
      */
     public static StateRef parse(@NotNull final String value, DigestService digestService) {
         try {
+            if (value.replaceAll("[^" + DELIMITER + "]", "").length() != 2) {
```
relyafi: I'm not sure this is safe - `replaceAll` performs [in-line replacement of the string](https://docs.oracle.com/javase/8/docs/api/java/lang/String.html#replaceAll-java.lang.String-java.lang.String-), so this will have a side effect on `value`, which might cause issues if it is used again in the calling code.

[This](https://www.baeldung.com/java-count-chars#4-using-java-8-features) may be a safer and cleaner approach. Unfortunately there doesn't seem to be an easy way with the standard Java libraries to do this.

Alternatively, I question whether it makes sense to have all this implementation in `corda-api`, or if we should create a `StateRefImpl` in `corda-runtime-os`, much as we do with `SecureHash` and `SecureHashImpl`. That means we'd have access to slightly nicer Kotlin libs. But that would be a much bigger change and we'd need to be careful not to break backwards compatibility.

@vlajos @nkovacsx any thoughts on the above? My feeling is maybe we'd have the separate impl if we were starting again, but it's a bit of a risk to do it now?

vlajos: After reading the ticket I expected something like:
```
if (lastIndexOfDelimiter == -1) {
   throw ...
}
```
And `parseSecureHash` could handle if the other delimiter is missing. Its exception looks understandable enough for me: `"Provided string: $algoNameAndHexString should be of format algorithm:hexadecimal"`

vlajos: > I'm not sure this is safe - `replaceAll` performs [in-line replacement of the string](https://docs.oracle.com/javase/8/docs/api/java/lang/String.html#replaceAll-java.lang.String-java.lang.String-), so this will have a side effect on `value`, which might cause issues if it is used again in the calling code.
> 
> [This](https://www.baeldung.com/java-count-chars#4-using-java-8-features) may be a safer and cleaner approach. Unfortunately there doesn't seem to be an easy way with the standard Java libraries to do this.
> 
> Alternatively, I question whether it makes sense to have all this implementation in `corda-api`, or if we should create a `StateRefImpl` in `corda-runtime-os`, much as we do with `SecureHash` and `SecureHashImpl`. That means we'd have access to slightly nicer Kotlin libs. But that would be a much bigger change and we'd need to be careful not to break backwards compatibility.
> 
> @vlajos @nkovacsx any thoughts on the above? My feeling is maybe we'd have the separate impl if we were starting again, but it's a bit of a risk to do it now?

I think it is a bit late now.
Also, I think it is good to have this parse accessible without more services.

ledger/ledger-utxo/src/main/java/net/corda/v5/ledger/utxo/StateRef.java
vlajos: Checking the `SecureHash` delimiter in this class mixes the two classes slightly. StateRefs should not need to know the internal structure of `SecureHash`es.
Theoretically, we should only check that we have a `:` and a number after it.
The `SecureHash` class should check whether the first part is a proper SecureHash or not.
But the theory part would be only deal-breaker if we had any chances to have SecureHashes with more than one or zero `:`s.
On the other hand, I think the `lastIndexOf()` solution is simpler and cleaner slightly than checking two delimiters.

ledger/ledger-utxo/src/test/java/net/corda/v5/ledger/utxo/uniqueness/client/StateRefParseTest.java
```diff
@@ -0,0 +1,75 @@
+package net.corda.v5.ledger.utxo.uniqueness.client;
```
vlajos: Why is this test in the `...uniqueness.client` package?

jennyang-r3: should this be inside `...uniqueness` rather than `...uniqueness.client`?
I was just trying to match with `StateRef.java` package name

vlajos: Yeah. The same package should be fine. (`net.corda.v5.ledger.utxo`)

jennyang-r3: Yeah, I just realised that is in `net.corda.v5.ledger.utxo` not in uniqueness

ledger/ledger-utxo/src/test/java/net/corda/v5/ledger/utxo/StateRefParseTest.java
```diff
@@ -0,0 +1,70 @@
...
+    void parseValidValue() {
+        final String value = "SHA-256D:ED87C7285E1E34BF5E46302086F76317ACE9B17AEF7BD086EE09A5ACBD17CEA4:0";
+
+        final int lastIndexOfDelimiter = value.lastIndexOf(DELIMITER);
+        final String subStringBeforeDelimiter = value.substring(0, lastIndexOfDelimiter);
+        final String subStringAfterDelimiter = value.substring(lastIndexOfDelimiter + 1);
+        final int index = Integer.parseUnsignedInt(subStringAfterDelimiter);
+        final SecureHash secureHash = mock(SecureHash.class);
+        final String digestName = "SHA-256D";
+        final String hexString = subStringBeforeDelimiter.substring(digestName.length() + 1);
```
vlajos: Do we need all of this logic?
I would probably just hardcode the needed bits.

vlajos (approved):
LGTM

relyafi (approved):
LGTM

driessamyn (approved):
LGTM

PR #1201, opened 2023-08-07
CORE-11990 - Modify signature handling

nikinagy:
Review how we handle signed data for member info and registration request. We should store and retrieve the data in its original serialised form instead of serialising and deserialising back and forth.
Remove queryMembersSignatures and get the signature information for data distribution by using queryMemberInfo instead.


data/avro-schema/src/main/resources/avro/net/corda/data/membership/common/RegistrationRequestDetails.avsc
```diff
@@ -42,33 +42,23 @@
...
+      "type": "net.corda.data.membership.SignedData"
     },
     {
-      "name": "registrationContextSignatureSpec",
-      "doc": "Signature spec of member signature over the registrationContext.",
-      "type": "net.corda.data.crypto.wire.CryptoSignatureSpec"
+      "name": "deserializedRegistrationContext",
+      "doc": "Additional registration context submitted for registration.",
+      "type": "net.corda.data.KeyValuePairList"
     },
```
charlieR3: Do we need both the serialised and the deserialised data? I think it would be best to keep just the serialised/signed if possible since there's a risk the serialised and deseriliased could differ if pass along separately.

nikinagy: addressed in [4683e6f](https://github.com/corda/corda-api/pull/1201/commits/4683e6fc25e3aacce324f6c655714c664975eda6)

charlieR3 (dismissed):

charlieR3 (approved):

PR #1202, opened 2023-08-09
CORE-15909: make CurrentGroupParametersLookup public api

jennyang-r3:
This PR introduces a new public api for `CurrentGroupParametersService` that will allow users access to `GroupParameters`

### DIFF
--- a/ledger/ledger-utxo/src/main/java/net/corda/v5/ledger/utxo/CurrentGroupParametersService.java
+++ b/ledger/ledger-utxo/src/main/java/net/corda/v5/ledger/utxo/CurrentGroupParametersService.java
@@ -4,10 +4,10 @@ import net.corda.v5.membership.GroupParameters;
 import org.jetbrains.annotations.NotNull;
 
 /**
- * {@link GroupParametersService} allows flows to access {@link GroupParameters} for further verification,
+ * {@link CurrentGroupParametersService} allows flows to access {@link GroupParameters} for further verification,
  * for example, accessing to check if issued money is from authoritative party in flows.
  */
-public interface GroupParametersService {
+public interface CurrentGroupParametersService {
 
     /**
      * Returns the point in time {@link GroupParameters}
--- a/ledger/ledger-utxo/src/main/java/net/corda/v5/ledger/utxo/CurrentGroupParametersService.java
+++ b/ledger/ledger-utxo/src/main/java/net/corda/v5/ledger/utxo/CurrentGroupParametersService.java
@@ -3,9 +3,6 @@ package net.corda.v5.ledger.utxo;
 import net.corda.v5.membership.GroupParameters;
 import org.jetbrains.annotations.NotNull;
 
-import java.security.PublicKey;
-import java.util.List;
-
 /**
  * {@link CurrentGroupParametersService} allows flows to access {@link GroupParameters} for further verification,
  * for example, accessing to check if issued money is from authoritative party in flows.
@@ -17,10 +14,4 @@ public interface CurrentGroupParametersService {
      */
     @NotNull
     GroupParameters getCurrentGroupParameters();
-
-    /**
-     * Returns the list of {@link PublicKey} of MGM.
-     */
-    @NotNull
-    List<PublicKey> getMgmKeys();
 }
--- a/ledger/ledger-utxo/src/main/java/net/corda/v5/ledger/utxo/CurrentGroupParametersService.java
+++ b/ledger/ledger-utxo/src/main/java/net/corda/v5/ledger/utxo/CurrentGroupParametersService.java
@@ -1,0 +1,26 @@
+package net.corda.v5.ledger.utxo;
+
+import net.corda.v5.membership.GroupParameters;
+import org.jetbrains.annotations.NotNull;
+
+import java.security.PublicKey;
+import java.util.List;
+
+/**
+ * {@link CurrentGroupParametersService} allows flows to access {@link GroupParameters} for further verification,
+ * for example, accessing to check if issued money is from authoritative party in flows.
+ */
+public interface CurrentGroupParametersService {
+
+    /**
+     * Returns the point in time {@link GroupParameters}
+     */
+    @NotNull
+    GroupParameters getCurrentGroupParameters();
+
+    /**
+     * Returns the list of {@link PublicKey} of MGM.
+     */
+    @NotNull
+    List<PublicKey> getMgmKeys();
+}
--- a/ledger/ledger-utxo/src/main/java/net/corda/v5/ledger/utxo/GroupParametersLookup.java
+++ b/ledger/ledger-utxo/src/main/java/net/corda/v5/ledger/utxo/GroupParametersLookup.java
@@ -4,10 +4,10 @@ import net.corda.v5.membership.GroupParameters;
 import org.jetbrains.annotations.NotNull;
 
 /**
- * {@link GroupParametersService} allows flows to access {@link GroupParameters} for further verification,
+ * {@link GroupParametersLookup} allows flows to access {@link GroupParameters} for further verification,
  * for example, accessing to check if issued money is from authoritative party in flows.
  */
-public interface GroupParametersService {
+public interface GroupParametersLookup {
 
     /**
      * Returns the point in time {@link GroupParameters}
--- a/ledger/ledger-utxo/src/main/java/net/corda/v5/ledger/utxo/GroupParametersService.java
+++ b/ledger/ledger-utxo/src/main/java/net/corda/v5/ledger/utxo/GroupParametersService.java
@@ -13,5 +13,5 @@ public interface GroupParametersService {
      * Returns the point in time {@link GroupParameters}
      */
     @NotNull
-    GroupParameters getGroupParameters();
+    GroupParameters getCurrentGroupParameters();
 }
--- a/ledger/ledger-utxo/src/main/java/net/corda/v5/ledger/utxo/GroupParametersService.java
+++ b/ledger/ledger-utxo/src/main/java/net/corda/v5/ledger/utxo/GroupParametersService.java
@@ -4,10 +4,10 @@ import net.corda.v5.membership.GroupParameters;
 import org.jetbrains.annotations.NotNull;
 
 /**
- * {@link CurrentGroupParametersService} allows flows to access {@link GroupParameters} for further verification,
+ * {@link GroupParametersService} allows flows to access {@link GroupParameters} for further verification,
  * for example, accessing to check if issued money is from authoritative party in flows.
  */
-public interface CurrentGroupParametersService {
+public interface GroupParametersService {
 
     /**
      * Returns the point in time {@link GroupParameters}
--- a/ledger/ledger-utxo/src/main/java/net/corda/v5/ledger/utxo/GroupParametersService.java
+++ b/ledger/ledger-utxo/src/main/java/net/corda/v5/ledger/utxo/GroupParametersService.java
@@ -4,14 +4,14 @@ import net.corda.v5.membership.GroupParameters;
 import org.jetbrains.annotations.NotNull;
 
 /**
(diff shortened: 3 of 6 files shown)


ledger/ledger-utxo/src/main/java/net/corda/v5/ledger/utxo/CurrentGroupParametersService.java
```diff
@@ -0,0 +1,26 @@
...
+     * Returns the point in time {@link GroupParameters}
+     */
+    @NotNull
+    GroupParameters getCurrentGroupParameters();
+
+    /**
+     * Returns the list of {@link PublicKey} of MGM.
+     */
+    @NotNull
+    List<PublicKey> getMgmKeys();
```
vlajos: I think this one does not need to be publicly accessible.
You can remove it from here.

ledger/ledger-utxo/src/main/java/net/corda/v5/ledger/utxo/CurrentGroupParametersService.java
```diff
@@ -0,0 +1,17 @@
+package net.corda.v5.ledger.utxo;
+
+import net.corda.v5.membership.GroupParameters;
+import org.jetbrains.annotations.NotNull;
+
+/**
+ * {@link CurrentGroupParametersService} allows flows to access {@link GroupParameters} for further verification,
+ * for example, accessing to check if issued money is from authoritative party in flows.
+ */
+public interface CurrentGroupParametersService {
```
lankydan: I think this should be called `GroupParametersService`, at least based on what me and @relyafi decided a few weeks ago, let me know if that decision/idea has changed while i was away.

jennyang-r3: Does it mean I should have created a new `GroupParametersService` rather than using the existing `CurrentGroupParametersService`?

blsemo: Can't you just rename this class to `GroupParametersService`?

jennyang-r3: I can. What I'm not sure of is the service we are extending this interface in runtime repo is called CurrentGroupParametersService so my question is whether I have to rename it too or just this public API.

jennyang-r3: wdyt @vlajos

lankydan: i think renaming the one in the runtime repo is fine.

ledger/ledger-utxo/src/main/java/net/corda/v5/ledger/utxo/transaction/UtxoLedgerTransaction.java
```diff
@@ -288,4 +289,7 @@ default List<ContractState> getOutputContractStates() {
     // TODO : Rename to getOutputContractStates
```
lankydan: Can you remove this `TODO`s in this class while you're here since they'll never be changed now (above `getOutputStates` and `getInputStates`).

ledger/ledger-utxo/src/main/java/net/corda/v5/ledger/utxo/CurrentGroupParametersService.java
vlajos: I agree with @lankydan that renaming both is the most consistent.

ledger/ledger-utxo/src/main/java/net/corda/v5/ledger/utxo/GroupParametersService.java
```diff
@@ -0,0 +1,17 @@
...
+ * {@link GroupParametersService} allows flows to access {@link GroupParameters} for further verification,
+ * for example, accessing to check if issued money is from authoritative party in flows.
+ */
+public interface GroupParametersService {
+
+    /**
+     * Returns the point in time {@link GroupParameters}
+     */
+    @NotNull
+    GroupParameters getGroupParameters();
```
lankydan: I would still call this `getCurrentGroupParameters`, might be worth getting @charlieR3's eyes on this just to make sure he's aware and agrees.

lankydan (dismissed):

ledger/ledger-utxo/src/main/java/net/corda/v5/ledger/utxo/transaction/UtxoLedgerTransaction.java
```diff
@@ -285,7 +286,9 @@ default List<ContractState> getOutputContractStates() {
      * @param type The type of the {@link ContractState}.
      * @return Returns all output contract states that match the specified type.
      */
-    // TODO : Rename to getOutputContractStates
     @NotNull
     <T extends ContractState> List<T> getOutputStates(@NotNull Class<T> type);
+
+    @NotNull
+    GroupParameters getGroupParameters();
```
charlieR3, line 298: Missing kdoc

ledger/ledger-utxo/src/main/java/net/corda/v5/ledger/utxo/GroupParametersService.java
```diff
@@ -0,0 +1,17 @@
+package net.corda.v5.ledger.utxo;
+
+import net.corda.v5.membership.GroupParameters;
+import org.jetbrains.annotations.NotNull;
+
+/**
+ * {@link GroupParametersService} allows flows to access {@link GroupParameters} for further verification,
+ * for example, accessing to check if issued money is from authoritative party in flows.
+ */
+public interface GroupParametersService {
```
charlieR3: LGTM, but my only thought here would be that a name like `GroupParametersLookup` would align better with `NotaryLookup` and `MemberLookup` services.

ledger/ledger-utxo/src/main/java/net/corda/v5/ledger/utxo/GroupParametersService.java
charlieR3: Yeah, I agree. `getCurrentGroupParameters` in this context makes more sense I think.

ledger/ledger-utxo/src/main/java/net/corda/v5/ledger/utxo/GroupParametersService.java
jennyang-r3: wdyt? @lankydan

lankydan: I'm fine with it being named `GroupParametersLookup` if that makes things more consistent. @vlajos what do you think?

charlieR3: I'm happy either way btw so if you prefer the Service suffix that's fine by me

vlajos: It is fine for me too.

lankydan: Lets go with `GroupParametersLookup` then. We can always change it later if we aren't happy with it.

charlieR3 (dismissed):
LGTM

lankydan (dismissed):

ledger/ledger-utxo/src/main/java/net/corda/v5/ledger/utxo/GroupParametersLookup.java
```diff
@@ -0,0 +1,17 @@
+package net.corda.v5.ledger.utxo;
```
lankydan: _Leaving this as a note to myself, so don't worry about it for now._

Should group parameters be available for the consensual ledger as well? If the answer is yes, then `GroupParametersLookup` should either live in the `membership` module or the `ledger-common` module.

There is an argument it should be in the `membership` module regardless.

lankydan: @blsemo @dimosr @charlieR3 let me know if you have any thoughts.

vlajos: I think adding anything to Consensual Ledger is a product question.
And as I heard, Consensual ledger is not one of the top priorities.
On the other hand, this module does not have anything UTXO-specific.
Also, future ledger modules may need it.

lankydan: Yeah, that's my thinking as well.

lankydan: Potentially this should actually go into `application` along side the `MemberLookup` interface.

charlieR3: Yes, I think it could fit in `application` with the `MemberLookup` interface. The `GroupParameters` interface is in the `membership` module already. 

Saying that though, the `NotaryInfo` interface is in `membership` but `NotaryLookup` is in `ledger-common`. I think `NotaryInfo` is mainly in `membership` though because `GroupParameters` depends on it and we don't want `membership` to depend on `ledger`.

I would lean towards having it in `membership` so that it is not ledger specific. We allow adding any custom metadata into to group parameters for example which doesn't necessarily need to be ledger specific.

lankydan: Maybe we should consider moving `NotaryLookup` too (we'd have to deprecate the existing one and copy it to `application`/`membership`).

I think feels like the right fit for `GroupParametersLookup` and possibly `NotaryLookup`.

Happy to go with the majority on this.

charlieR3: The original logic for `NotaryLookup` being in `ledger-common` was that it's tightly linked with the ledger layer so it seems like a better place for it. But not sure that matters much if `NotaryInfo` is in `membership` anyway. `membership` is a bit of a cross layer concept anyway so I'm happy to go with the majority also.

vlajos: I would handle NotaryLookup in a separate ticket.
But I would move `GroupParametersLookup` to the most future-proof place in this one.
(However, I am not too sure whether application or membership is the best.)

lankydan: Lets move `GroupParametersLookup` to `v5.application.membership`. We can raise a tech debt for `NotaryLookup`, ideally to be reviewed before 5.1. Then we can decide if they should all move to the `membership` module or if they all live in `v5.application.membership`.

ledger/ledger-utxo/src/main/java/net/corda/v5/ledger/utxo/GroupParametersLookup.java
```diff
@@ -0,0 +1,17 @@
...
+
+import net.corda.v5.membership.GroupParameters;
+import org.jetbrains.annotations.NotNull;
+
+/**
+ * {@link GroupParametersLookup} allows flows to access {@link GroupParameters} for further verification,
+ * for example, accessing to check if issued money is from authoritative party in flows.
+ */
+public interface GroupParametersLookup {
```
lankydan: Just realised we should add `@DoNotImplement` to the interface.

vlajos (approved):

lankydan (approved):

blsemo (approved):
LGTM

PR #1204, opened 2023-08-10
CORE-15179 - Updated the avro object TokenClaim

filipesoliveira:
The field `claimedTokenStateRefs` has been deprecated. 
The field `claimedTokens` has been added in.
The fields `claimedTokenStateRefs` and `claimedTokens` are by default empty arrays to ensure backwards compatibility.


owenstanford (approved):
LGTM

PR #1206, opened 2023-08-11
CORE-15586 Split verification from TransactionSignatureService into TransactionSignatureVerificationService

vlajos:
CORE-15586 Split verification from TransactionSignatureService into TransactionSignatureVerificationService to let it be injected separately in the verification sandbox. That way, we can keep the verification sandbox more streamlined.

### DIFF
--- a/ledger/ledger-common/src/main/java/net/corda/v5/ledger/common/transaction/TransactionSignatureVerificationService.java
+++ b/ledger/ledger-common/src/main/java/net/corda/v5/ledger/common/transaction/TransactionSignatureVerificationService.java
@@ -1,0 +1,54 @@
+package net.corda.v5.ledger.common.transaction;
+
+import net.corda.v5.application.crypto.DigitalSignatureAndMetadata;
+import net.corda.v5.application.crypto.DigitalSignatureVerificationService;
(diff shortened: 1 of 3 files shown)


ledger/ledger-common/src/main/java/net/corda/v5/ledger/common/transaction/TransactionSignatureVerificationService.java
```diff
@@ -0,0 +1,54 @@
...
+ * It can be used in both flows and contracts.
+ */
+@DoNotImplement
+public interface TransactionSignatureVerificationService {
+    /**
+     * Verifies a signature against a transaction.
+     * The underlying verification service signals the verification failures with different exceptions.
+     * {@link DigitalSignatureVerificationService}
+     *
+     * @param transaction           The original transaction.
```
vlajos, line 26: This method (verifySignature against TransactionWithMetadata) uses the metadata only to cross-match the batch settings against the digital signature's batch settings.
The verification itself works without the transaction's metadata.
We could deprecate this function if we do not need this extra verification logic.
But since that would be an API breaking change, it needs to stay.

blsemo (approved):
LGTM

PR #1210, opened 2023-08-13
CORE-14685: Corda 5.1 cannot use SLF4J 2.x yet.

chrisr3:
Force Gradle to resolve SLF4J 1.7.36, even if SLF4J 2.x has been provided, because Corda 5.1 doesn't support SLF4J 2.x yet.


yarivus (approved):
LGTM

PR #1217, opened 2023-08-15
CORE-16313: Fix JUnit installation to be compatible with Gradle 9.

chrisr3:
Replace `junit-jupiter-api` and `junit-jupiter-params` with `junit-jupiter` so that we also get the JUnit BOM.

Also upgrade JUnit 5.9.3 -> 5.10.0.


ronanbrowne (approved):
LGTM

yarivus (commented):
thx @chrisr3

PR #1218, opened 2023-08-16
CORE-16329 - Avro and DB schema for new Scheduler component.

driessamyn:
Avro and DB schema for the new Scheduler component.

### DIFF
--- a/data/db-schema/src/main/resources/net/corda/db/schema/config/migration/scheduler-creation-v5.1.xml
+++ b/data/db-schema/src/main/resources/net/corda/db/schema/config/migration/scheduler-creation-v5.1.xml
@@ -1,0 +1,22 @@
+<?xml version="1.1" encoding="UTF-8" standalone="no"?>
+<databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+                   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-4.3.xsd">
+
+    <changeSet author="R3.Corda" id="scheduler-creation-v5.1">
+
+        <createTable tableName="task_scheduler_log">
+            <column name="task_name" type="VARCHAR(255)">
+                <constraints nullable="false"/>
+            </column>
+            <column name="last_scheduled" type="DATETIME">
+                <constraints nullable="false"/>
+            </column>
+            <column name="scheduler_id" type="VARCHAR(255)">
+                <constraints nullable="false"/>
+            </column>
+        </createTable>
+
+        <addPrimaryKey columnNames="task_name" constraintName="task_name_pk" tableName="task_scheduler_log"/>
+    </changeSet>
+</databaseChangeLog>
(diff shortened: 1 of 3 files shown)


data/db-schema/src/main/resources/net/corda/db/schema/config/migration/scheduler-creation-v5.1.xml
```diff
@@ -0,0 +1,25 @@
...
+            </column>
+            <column name="last_scheduled" type="DATETIME" defaultValueDate="${now}">
+                <constraints nullable="false"/>
+            </column>
+            <column name="scheduler_id" type="VARCHAR(255)">
+                <constraints nullable="false"/>
+            </column>
+        </createTable>
+
+        <addPrimaryKey columnNames="task_name" constraintName="task_name_pk" tableName="task_scheduler_log"/>
```
mbrkic-r3: I'm not sure whether we have naming conventions for DB objects, but it might be better that PK constraints contain table names, to make sure they have uniques name (as multiple tables could have same field names). Not having unique constraint name might be a problem for some databases. Also when constraint violation references PK name, it would be clear for which table the problem happened.

data/db-schema/src/main/resources/net/corda/db/schema/config/migration/scheduler-creation-v5.1.xml
```diff
@@ -0,0 +1,25 @@
...
+    <property name="now" value="now()" dbms="hsqldb"/>
+    <property name="now" value="now()" dbms="postgresql"/>
+
+    <changeSet author="R3.Corda" id="scheduler-creation-v5.1">
+
+        <createTable tableName="task_scheduler_log">
+            <column name="task_name" type="VARCHAR(255)">
+                <constraints nullable="false"/>
+            </column>
+            <column name="last_scheduled" type="DATETIME" defaultValueDate="${now}">
```
mbrkic-r3: We should either use type that doesn't have issues with timezones or have a convention to always store times in UTC timezone.

driessamyn: What change do you suggest?

For postgres, this will [be converted to](https://dba-presents.com/liquibase/216-liquibase-3-6-x-data-types-mapping-table) a `TIMESTAMP` field.

data/db-schema/src/main/resources/net/corda/db/schema/config/migration/scheduler-creation-v5.1.xml
driessamyn: Updated.

data/db-schema/src/main/resources/net/corda/db/schema/config/migration/scheduler-creation-v5.1.xml
mbrkic-r3: Either using `timestamptz` for Postgres, i.e. using DB specific DDL and DML (Hibernate might help with reducing DB specific code) or continue using `DATETIME` but making sure that in code all time/date values sent/received are converted from/to UTC timezone. This should be consistent across the whole code base so probably requires a wider discussion.

driessamyn: The implementation currently actually only uses the DB time. I.e. we compare the last recorded time with the DB `now()`, so it should be ok.
But I agree that maybe it needs a wider review as this follows a pattern we've used elsewhere.
@simon-johnson-r3 , maybe this is something for the tech debt backlog? Unless you'd like me to add it to the general tech backlog?

mbrkic-r3: Even if all values come from the same source (e.g. database), we should take care that UTC timezone is being set, as in other timezones the same time could occur multiple times due to DST.

mbrkic-r3 (approved):
LGTM

data/db-schema/src/main/resources/net/corda/db/schema/config/migration/scheduler-creation-v5.1.xml
driessamyn: Updated to use `TIMESTAMP WITH TIME ZONE`

PR #1220, opened 2023-08-17
CORE-5533 Moving additionalProperties: false flag for the corda flow configuration schema

ciaran238:
### Context
The `corda.flow.json` schema was allowing for the creation of additional properties because of a misplaced `"additionalProperties": false` flag.
### Solution
Moving the `"additionalProperties": false` flag to outside the correct properties field


emilybowe (approved):
LGTM

PR #1222, opened 2023-08-18
CORE-13621 Idempotency id table definition

kyriathar:
Introduces table in vnode-vault schema which holds executed persistence requests (ids) from `PersistenceService`, to be used for deduplicating them.

runtime-os PR: [#4491](https://github.com/corda/corda-runtime-os/pull/4491)

### DIFF
--- a/data/db-schema/src/main/resources/net/corda/db/schema/vnode-vault/migration/vnode-vault-creation-v5.1.xml
+++ b/data/db-schema/src/main/resources/net/corda/db/schema/vnode-vault/migration/vnode-vault-creation-v5.1.xml
@@ -6,5 +6,10 @@
         <addColumn tableName="vnode_member_info">
             <column name="is_deleted" type="BOOLEAN" defaultValueBoolean="false"/>
         </addColumn>
+        <createTable tableName="vnode_persistence_request_id">
+            <column name="request_id" type="NVARCHAR(255)">
+                <constraints nullable="false" primaryKey="true"/>
+            </column>
+        </createTable>
     </changeSet>
 </databaseChangeLog>
--- a/data/db-schema/src/main/resources/net/corda/db/schema/vnode-vault/migration/vnode-vault-creation-v5.1.xml
+++ b/data/db-schema/src/main/resources/net/corda/db/schema/vnode-vault/migration/vnode-vault-creation-v5.1.xml
@@ -6,6 +6,7 @@
         <addColumn tableName="vnode_member_info">
             <column name="is_deleted" type="BOOLEAN" defaultValueBoolean="false"/>
         </addColumn>
+
         <createTable tableName="vnode_persistence_request_id">
             <column name="request_id" type="NVARCHAR(255)">
                 <constraints nullable="false" primaryKey="true"/>
--- a/data/db-schema/src/main/java/net/corda/db/schema/DbSchema.java
+++ b/data/db-schema/src/main/java/net/corda/db/schema/DbSchema.java
@@ -38,6 +38,7 @@ public final class DbSchema {
     public static final String VNODE_GROUP_PARAMETERS = "vnode_group_parameters";
     public static final String VNODE_GROUP_APPROVAL_RULES = "vnode_group_approval_rules";
     public static final String VNODE_PRE_AUTH_TOKENS = "vnode_pre_auth_tokens";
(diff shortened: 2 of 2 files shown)


simon-johnson-r3 (approved):

data/db-schema/src/main/resources/net/corda/db/schema/vnode-vault/migration/vnode-vault-creation-v5.1.xml
```diff
@@ -6,5 +6,11 @@
         <addColumn tableName="vnode_member_info">
             <column name="is_deleted" type="BOOLEAN" defaultValueBoolean="false"/>
         </addColumn>
+
+        <createTable tableName="vnode_persistence_request_id">
+            <column name="request_id" type="NVARCHAR(255)">
+                <constraints nullable="false" primaryKey="true"/>
+            </column>
+        </createTable>
```
driessamyn, line 14: What is the cleanup strategy for this table?
Would it not be useful to have a timestamp column that can be used for partitioning and/or cleanup?
Given it is insert only, I suspect it would be very cheap to have a DB auto-populated timestamp column, but we could check with @ifrankrui

kyriathar: The cleanup strategy will be done in follow-up work. This PR only introduces the part of the deduplication table (the request ids) needed to deduplicate the persistence requests (in https://github.com/corda/corda-runtime-os/pull/4491).

driessamyn: But do we have a design yet for the clean-up strategy?
I'm a bit uncomfortable with merging something in which we know will cause problems, without a plan for how we will mitigate it.
And if we know what the design will be, we may as well get the DB structure correct from the get-go.

ifrankrui: Don't you already include the housekeeping strategy [here](https://github.com/corda/platform-eng-design/pull/568)? 

Yes, you could have a timestamp column for the deduplication table and populate the timestamp as a part of the insert request. However, I would not recommend using DB auto-populated timestamp column. They are implemented differently between databases and there isn't one simple way to define them in Liquibase change log.

driessamyn: > there isn't one simple way to define them in Liquibase change log.

@ifrankrui I'm interested in this comment. Why is that?
We do this in a number of places already (e.g.: https://github.com/corda/corda-api/blob/9ea985216d72b74f946d9f967699244bce97c34b/data/db-schema/src/main/resources/net/corda/db/schema/config/migration/config-creation-v1.0.xml#L6)

ifrankrui: @driessamyn As pointed out in the above changelog, you have to define an extra property for one type of database when you support multiple database types. If you populate the timestamp column as a part of a persistence request, then you don't need these Liquibase properties.

davidcurrie (approved):
LGTM

data/db-schema/src/main/resources/net/corda/db/schema/vnode-vault/migration/vnode-vault-creation-v5.1.xml
```diff
@@ -6,5 +6,11 @@
         <addColumn tableName="vnode_member_info">
             <column name="is_deleted" type="BOOLEAN" defaultValueBoolean="false"/>
         </addColumn>
+
+        <createTable tableName="vnode_persistence_request_id">
+            <column name="request_id" type="NVARCHAR(255)">
```
sprintengine, line 11: Just passing by, but I wonder if NVARCHAR(36) would be suffice as I think a UUID fits in it? I guess it depends how we build the idempotency ID.

kyriathar: We only store the `UUID.toString()` in it. Will take a look to see if we can shorten the DB field.

PR #1223, opened 2023-08-20
CORE-16318 - API changes to support state manager

conal-smith:
The following [design](https://github.com/corda/platform-eng-design/blob/8cc7b58d8d38cf54da2030e51e5b04c7ed2c62ea/core/corda-5/corda-5.0/flow-worker/state-manager.md) describes the state manager.

This set of changes facilitates these [runtime-os](https://github.com/corda/corda-runtime-os/pull/4501) changes. Most of these changes relate to https://r3-cev.atlassian.net/browse/CORE-16318.

- State manager config will be presented to workers via boot config on the command line. 
- Defined a state-manager config schema that is polymorphic, if stateManagerType is DATABASE it holds database config. In future "REDIS" can provide its own config file.
- Defined the state-manager database schema migration files. This sets up `state` table with some column types depending on the database being postgres or hsqldb.
- A few other corda-api state manager related changes.


data/state-manager-schema/src/main/resources/net/corda/statemanager/database/migration/state-manager-migration-v1.0.xml
```diff
@@ -0,0 +1,17 @@
+<?xml version="1.1" encoding="UTF-8" standalone="no"?>
```
conalsmith-r3, line 1: todo ; will remove the separate state manager schema

data/config-schema/src/main/resources/net/corda/schema/configuration/stateManager/1.0/corda.stateManager.json
```diff
@@ -0,0 +1,41 @@
+{
+  "$schema": "https://json-schema.org/draft/2019-09/schema",
+  "$id": "https://corda.r3.com/net/corda/schema/configuration/messaging/1.0/corda.stateManager.json",
+  "title": "Corda State Manager Configuration Schema",
+  "description": "Configuration schema for the state manager section.",
+  "type": "object",
```
JamesHR3: Think this might be more like a section under messaging rather than a new section.

data/db-schema/src/main/resources/net/corda/db/schema/statemanager/migration/state-manager-migration-v1.0.xml
```diff
@@ -0,0 +1,35 @@
...
+    <property name="json.column.type" value="CLOB" dbms="hsqldb"/>
+    <property name="state.column.type" value="binary" dbms="hsqldb"/>
+
+    <changeSet author="R3.Corda" id="state-manager-migration-v1.0">
+        <createTable tableName="state">
+            <column name="key" type="varchar(255)">
+                <constraints nullable="false"/>
+            </column>
+            <column name="state" type="${state.column.type}">
+                <constraints nullable="false"/>
```
sprintengine, line 21: May need to double check this, but when we complete a flow, we null the state, we'll have to now delete the key  from state storage if we can't null the state (probably a better thing anyway)

conal-smith (commented):
Lgtm

jujoramos:
State manager configurations settings have been moved to https://github.com/corda/corda-api/pull/1240.

PR #1226, opened 2023-08-24
CORE-16244 Adding flow mapper worker to topic config schemas

ben-millar:
### Overview
A new `FlowMapperWorker` is being created, removing the flow mapper from the `FlowWorker` and moving it to its own dedicated processor and worker which can be deployed and scaled separately. As part of this, we need to add the `flow-mapper` as a producer/consumer in the requisite topic schemas.

### Related PRs:
- https://github.com/corda/corda-runtime-os/pull/4513
- https://github.com/corda/corda-shared-build-pipeline-steps/pull/1103
- https://github.com/corda/corda-e2e-tests/pull/222

### DIFF
--- a/data/topic-schema/src/main/resources/net/corda/schema/Flow.yaml
+++ b/data/topic-schema/src/main/resources/net/corda/schema/Flow.yaml
@@ -34,16 +34,18 @@ topics:
     name: flow.mapper.event
     consumers:
       - flow
+      - flow-mapper
     producers:
-      - flow
+      - flow-mapper
       - rest
     config:
   FlowMapperEventStateTopic:
     name: flow.mapper.event.state
     consumers:
       - flow
+      - flow-mapper
     producers:
-      - flow
+      - flow-mapper
     config:
       cleanup.policy: compact
       segment.ms: 600000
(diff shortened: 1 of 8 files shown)


data/topic-schema/src/main/resources/net/corda/schema/Flow.yaml
```diff
@@ -3,10 +3,12 @@ topics:
     name: flow.event
     consumers:
       - flow
+      - flowMapper
```
LWogan: Flow Mapper doesnt consume this.

data/topic-schema/src/main/resources/net/corda/schema/P2P.yaml
```diff
@@ -3,6 +3,7 @@ topics:
     name: p2p.in
     consumers:
       - flow
```
LWogan: is the FlowP2PFilterService in flow mapper worker or flow worker now? i think it should probably go into flow mapper worker and then flow is no longer a consumer of the p2p.in

data/topic-schema/src/main/resources/net/corda/schema/P2P.yaml
```diff
@@ -13,6 +14,7 @@ topics:
       - link-manager
     producers:
       - flow
```
LWogan: flow wont produce to p2p.out anymore. the mapper does that

LWogan (commented):
Couple comments

PR #1228, opened 2023-08-25
CORE-15820 - Added a new Token observer version

filipesoliveira:
This PR depends on: https://github.com/corda/corda-runtime-os/pull/4652


filipesoliveira:
The compatibility checks are failing because the existing observer has been marked as deprecated which is being marked as an error instead of an warn during the build time.

owenstanford (approved):

josephzunigadaly (approved):

PR #1229, opened 2023-08-30
CORE-13119 Add `resumePoint` Avro field to support stable query paging

relyafi:
This is a pre-requisite change to enable a `resumePoint` field to be passed back and forth between ledger custom queries initiated in a flow processor and the query execution logic in a ledger persistence processor.

See https://github.com/corda/corda-runtime-os/pull/4548 for full details of the changes.

### DIFF
--- a/data/avro-schema/src/main/resources/avro/net/corda/data/persistence/EntityResponse.avsc
+++ b/data/avro-schema/src/main/resources/avro/net/corda/data/persistence/EntityResponse.avsc
@@ -16,6 +16,12 @@
       "name": "metadata",
       "doc": "Metadata returned from the execution of the persistence operation",
       "type": "net.corda.data.KeyValuePairList"
+    },
+    {
+      "name": "resumePoint",
+      "doc": "Used by queries that support stable paging to return opaque data that indicates where the next page should resume from",
+      "type": ["null", "bytes"],
+      "default": null
     }
   ]
 }
--- a/data/avro-schema/src/main/resources/avro/net/corda/data/persistence/FindWithNamedQuery.avsc
+++ b/data/avro-schema/src/main/resources/avro/net/corda/data/persistence/FindWithNamedQuery.avsc
@@ -1,7 +1,7 @@
 {
   "type": "record",
   "name": "FindWithNamedQuery",
-  "doc": "Find entities matched by a named query. Parameters for the query may be specified. Pagination is supported, but there's no guarantee of consistency if the underlying data changes. The data will not be sorted in Corda; if the database returns an unpredictable order that must be handled by the caller. FindWithNamedQuery can be used to enforce ordering if required. Replies with {@link EntityResponse}",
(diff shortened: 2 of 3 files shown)


data/db-schema/src/main/resources/net/corda/db/schema/vnode-vault/migration/ledger-utxo-creation-v5.1.xml
```diff
@@ -0,0 +1,20 @@
+<?xml version="1.1" encoding="UTF-8" standalone="no"?>
+<databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+                   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-4.3.xsd">
+
+    <changeSet author="R3.Corda" id="ledger-utxo-creation-v5.1">
+        <dropPrimaryKey tableName="utxo_visible_transaction_state"/>
+        <addColumn tableName="utxo_visible_transaction_state">
+            <column name="sequence_num" type="bigint" autoIncrement="true">
```
vlajos: Is this for stable paging?
Some RDBMS tend to cache similar things, which may be worth checking from a monotonicity point of view.
(https://docs.oracle.com/en/database/oracle/oracle-database/19/sqlrf/CREATE-SEQUENCE.html#GUID-E9C78A8C-615A-4757-B2A8-5E6EFB130571__GUID-7E390BE1-2F6C-4E5A-9D5C-5A2567D636FB vs RAC environment, also db session-level caching could be problematic.)

Hopefully, none of these are real problems.

relyafi: Hmm, sadly I think your concerns may be valid: https://docs.liquibase.com/change-types/add-auto-increment.html

Looks like Oracle doesn't support auto-increment columns.

We may instead have to look at re-using the existing key for providing stable query paging. That should still work, but it will be more convoluted / less efficient (we will need a much bigger, multi-field where clause when doing the querying).

blsemo (approved):
LGTM

vlajos (approved):

nkovacsx (approved):

data/avro-schema/src/main/resources/avro/net/corda/data/persistence/EntityResponse.avsc
```diff
@@ -16,6 +16,12 @@
       "name": "metadata",
       "doc": "Metadata returned from the execution of the persistence operation",
       "type": "net.corda.data.KeyValuePairList"
+    },
+    {
+      "name": "resumePoint",
+      "doc": "Used by queries that support stable paging to return opaque data that indicates where the next page should resume from",
+      "type": ["null", "bytes"],
```
driessamyn, line 23: Why is this a byte array rather than a structured type?

relyafi: We don't want to be opinionated on the contents of the field, it could vary from query to query - the data is opaque from the perspective of the interface.

PR #1233, opened 2023-09-04
CORE-16626: Replace string UUIDs with UUID objects

julia-filipczak:
Refactoring to take in UUIDs as UUID objects rather than Strings wherever possible.


ac101m (approved):
LGTM

jmacmahonr3 (approved):

PR #1238, opened 2023-09-06
CORE-14025 Add new request to find existing transactions in database and their status

nkovacsx:
### Overview

Adding a new request to the UTXO ledger layer to fetch transaction alongside with their status.

This is needed when doing a back-chain resolution. We don't want to fetch transactions that are already present in our database. We just want to fetch their IDs though, no need to deserialise the whole object.


data/avro-schema/src/main/resources/avro/net/corda/data/ledger/persistence/FindExistingNotInvalidTransactionIds.avsc
```diff
@@ -0,0 +1,14 @@
+{
+  "type": "record",
+  "name": "FindExistingNotInvalidTransactionIds",
+  "doc": "A request to fetch transaction that are not invalid (i.e. Verified or Unverified).",
+  "namespace": "net.corda.data.ledger.persistence",
```
lankydan: Looking forward, do we think there would be any benefit of having the ability to find transaction in a set of statuses? @relyafi @blsemo.

If not, then we should go with what Nandor has done, otherwise we can make it more conditional.

relyafi: @lankydan I'd imagine for the most part, people would be interested in valid transactions only? But at the state level, a lot of our queries are implicitly filtered on that already by virtue of only being in the outputs tables if they are verified.

vlajos: That implicit `NotInvalid` is a bit strange. Accepting a set of statuses would make it more generic.
Also, do we need `Existing` in the name?

nkovacsx: Okay, I'm happy to modify it to be more generic.

vlajos (dismissed):

data/avro-schema/src/main/resources/avro/net/corda/data/ledger/persistence/FindTransactionIds.avsc
```diff
@@ -0,0 +1,23 @@
...
+    },
+    "doc": "A list of the transaction IDs to fetch."
+  },
+  {
+    "name": "transactionStatuses",
+    "type": {
+      "type" : "array",
+      "items" : "string"
+    },
+    "doc": "The statuses of the transactions. The transaction must to have one of these statuses in order to be fetched."
```
blsemo:
```suggestion
    "doc": "The required statuses of the transactions. The transaction must have one of these statuses in order to be fetched."
```

data/avro-schema/src/main/resources/avro/net/corda/data/ledger/persistence/FindTransactionIds.avsc
```diff
@@ -0,0 +1,23 @@
...
+  "doc": "A request to fetch transaction IDs that are in one of the given statuses.",
+  "namespace": "net.corda.data.ledger.persistence",
+  "fields": [
+  {
+    "name": "ids",
+    "type": {
+      "type" : "array",
+      "items" : "string"
+    },
+    "doc": "A list of the transaction IDs to fetch."
```
blsemo: I am not quite clear how this is used - do you give it a list of IDs and statuses, and it will check the status of the transactions and return the list filtered down to the required status?

nkovacsx: You give a list of transaction IDs and statuses and it will return a list of transaction IDs that are present in the database with the given statuses.

vlajos (commented):
It needs a sync, but otherwise, LGTM.

vlajos (dismissed):
LGTM
(PR's title is outdated a bit.)

lankydan (dismissed):

lankydan (approved):

blsemo (approved):

PR #1239, opened 2023-09-06
CORE-9975: Add support for key category parameter to signing API

malachyb:
Added support for context parameter as a `map<string, string>` to signing API.


dickon (approved):
LGTM. SInce we are adding a new overload method this should be fully compatible.

davidcurrie (approved):
LGTM

PR #1240, opened 2023-09-07
CORE-16318: State Manager Configuration

jujoramos:
Add new polymorphic section under the current messaging configuration
section to hold state manager settings.

👮🏻👮🏻👮🏻 !!!! DESCRIBE YOUR CHANGES HERE !!!! DO NOT FORGET !!!! 👮🏻👮🏻👮🏻

# PR Checklist:

- [X] Have you run the unit, integration and smoke tests as described [here](https://docs.corda.net/head/testing.html)?
- [X] If you added public APIs, did you write the JavaDocs?
- [X] If the changes are of interest to application developers, have you added them to the [changelog](https://docs.corda.net/head/changelog.html) (`/docs/source/changelog.rst`), and potentially the [release notes](https://docs.corda.net/head/release-notes.html) (`/docs/source/release-notes.rst`)?
- [X] If you are contributing for the first time, please read the [contributor agreement](https://docs.corda.net/head/contributing.html) now and add a comment to this pull request stating that your PR is in accordance with the [Developer's Certificate of Origin](https://docs.corda.net/head/contributing.html#merging-the-changes-back-into-corda).

Thanks for your code, it's appreciated! :)

PR #1242, opened 2023-09-11
CORE-17154 - boot config for state manager

conal-smith:
Introduces boot config keys for the state manager. Follows on from [1240](https://github.com/corda/corda-api/pull/1240).

Runtime-os change: https://github.com/corda/corda-runtime-os/pull/4624

### DIFF
--- a/data/config-schema/src/main/java/net/corda/schema/configuration/MessagingConfig.java
+++ b/data/config-schema/src/main/java/net/corda/schema/configuration/MessagingConfig.java
@@ -95,7 +95,7 @@ public final class MessagingConfig {
         public static final String JDBC_PASS = DB_PROPERTIES + ".pass";
 
         public static final String JDBC_URL = DB_PROPERTIES + ".jdbc.url";
-        public static final String JDBC_DRIVER = DB_PROPERTIES + "jdbc.driver";
+        public static final String JDBC_DRIVER = DB_PROPERTIES + ".jdbc.driver";
         public static final String JDBC_DRIVER_DIRECTORY = DB_PROPERTIES + ".jdbc.directory";
         public static final String JDBC_PERSISTENCE_UNIT_NAME = DB_PROPERTIES + ".jdbc.persistenceUnitName";
         public static final String JDBC_POOL_MAX_SIZE = DB_PROPERTIES + ".pool.maxSize";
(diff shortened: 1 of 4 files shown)


jujoramos (dismissed):

data/config-schema/src/main/java/net/corda/schema/configuration/MessagingConfig.java
```diff
@@ -95,14 +95,14 @@ private StateManager() {
...
-        public static final String JDBC_DRIVER = DB_PROPERTIES + "jdbc.driver";
+        public static final String JDBC_DRIVER = DB_PROPERTIES + ".jdbc.driver";
         public static final String JDBC_DRIVER_DIRECTORY = DB_PROPERTIES + ".jdbc.directory";
         public static final String JDBC_PERSISTENCE_UNIT_NAME = DB_PROPERTIES + ".jdbc.persistenceUnitName";
         public static final String JDBC_POOL_MAX_SIZE = DB_PROPERTIES + ".pool.maxSize";
         public static final String JDBC_POOL_MIN_SIZE = DB_PROPERTIES + ".pool.minSize";
         public static final String JDBC_POOL_IDLE_TIMEOUT_SECONDS = DB_PROPERTIES + ".pool.idleTimeoutSeconds";
         public static final String JDBC_POOL_MAX_LIFETIME_SECONDS = DB_PROPERTIES + ".pool.maxLifetimeSeconds";
-        public static final String JDBC_POOL_KEEP_ALIVE_TIME_SECONDS = DB_PROPERTIES + ".pool.keepAliveTimeSeconds";
+        public static final String JDBC_POOL_KEEP_ALIVE_TIME_SECONDS = DB_PROPERTIES + ".pool.keepaliveTimeSeconds";
```
sprintengine: I noticed we've introduced slightly inconsistent naming for this property, I just want to keep it consistent

jujoramos: I've noticed the inconsistency, and I changed the name on purpose actually... the correct value is `keepAliveTimeSeconds` so we should keep that instead of the old (wrong) one in my opinion 🤷‍♂️

sprintengine: I'm not sure we can actually change the others (crypto and rbac) at this stage, that's probably a breaking change? I'm assuming this values schema is considered part of our public API?

jujoramos: We probably can't without breaking backward compatibility, true, but why keep the wrong name in newly introduced API instead of using the correct one?.

sprintengine: It's unfortunately spread around the codebase quite a bit, I'd prefer we named it proper camelcase but my gut tells me to keep it consistent. I will change it back since we can't really change the others, at least stateManager will be correct.

jujoramos (approved):
LGTM

JamesHR3 (approved):

PR #1244, opened 2023-09-12
CORE-16614: index utxo_visible_transaction_state.consumed

jennyang-r3:
This PR indexes `consumed` field of `utxo_visible_transaction_state`. 
This is to improve performance of `UtxoPersistenceService.findUnconsumedVisibleStatesByType` query so that other tables are pre-sorted by `consumed` timestamp and query performance improvement.

**Index created:**
image

**the query**
image

**execution plan before indexing**
before

**database optimizer has chosen to use an index to retrieve rows from the table**
after

### DIFF
--- a/data/db-schema/src/main/resources/net/corda/db/schema/vnode-vault/migration/ledger-utxo-creation-v5.1.xml
+++ b/data/db-schema/src/main/resources/net/corda/db/schema/vnode-vault/migration/ledger-utxo-creation-v5.1.xml
@@ -7,8 +7,9 @@
        <createIndex
            indexName="idx_consumed"
            tableName="utxo_visible_transaction_state">
-           <column name="consumed"/>
-           <where>consumed IS NULL</where>
+           <column name="consumed">
+               <where>consumed IS NULL</where>
+           </column>
        </createIndex>
     </changeSet>
 
--- a/data/db-schema/src/main/resources/net/corda/db/schema/vnode-vault/migration/ledger-utxo-creation-v5.1.xml
+++ b/data/db-schema/src/main/resources/net/corda/db/schema/vnode-vault/migration/ledger-utxo-creation-v5.1.xml
@@ -7,9 +7,8 @@
        <createIndex
            indexName="idx_consumed"
            tableName="utxo_visible_transaction_state">
-           <column name="consumed">
-               <where>consumed IS NULL</where>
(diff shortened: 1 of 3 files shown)


jennyang-r3:
A partial index is an alternative to achieve this, which is faster by indexing null records only (unconsumed states)
The drawbacks of the implementation is:
- maintenance if there are other types dbms
- implementation doesn't look intuitive as below ([stackOverFlow link](https://stackoverflow.com/questions/58292778/create-partial-index-with-liquibase))
```sql
<changeSet  author="some_author" id="someId"  context="some_context" dbms="some_db_key">
    <preConditions onFail="MARK_RAN">

        <columnExists tableName="some_table_name" columnName="Col1"/>

        <columnExists tableName="some_table_name" columnName="Col2"/>

    </preConditions>

    <sql splitStatements="true" stripComments="true" endDelimiter="\nGO">
        CREATE UNIQUE INDEX idx_unique_colName1_colName2 ON some_table_name (Col1, Col2)
        WHERE Col2 = '1'
    </sql>
</changeSet>
```

lankydan (dismissed):
LGTM but can someone else check too.

data/db-schema/src/main/resources/net/corda/db/schema/vnode-vault/migration/ledger-utxo-creation-v5.1.xml
```diff
@@ -0,0 +1,14 @@
+<?xml version="1.1" encoding="UTF-8" standalone="no"?>
+<databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+                   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-4.3.xsd">
+
+    <changeSet author="R3.Corda" id="ledger-utxo-creation-v5.1">
+       <createIndex
+           indexName="idx_consumed"
```
```suggestion
           indexName="utxo_visible_transaction_state_idx_consumed"
```

Consistent with other indexes in other tables, and avoids ambiguity if we have similar indexes elsewhere in future.

relyafi (commented):
One minor comment, otherwise looks good. I like the test note detail that shows the query plan before & after.

relyafi (approved):
LGTM

PR #1251, opened 2023-09-15
CORE-16615: create index for utxo_transaction_output.type

jennyang-r3:
This PR created index for `type` of `utxo_transaction_output` in order to improve performance of `findUnconsumedStatesByType` query

**query unconsumed states by type**
image

**the query used utxo_transaction_output_pkey index before**
before

**the query is now using utxo_transaction_output_idx_type index**
after


vlajos:
Did anything else change in the plan's logic?

vlajos (approved):
LGTM, assuming the new query with the related WHERE condition will be added.

PR #1253, opened 2023-09-19
CORE-9975: Add support for key category parameter to signing API

malachyb:
Added support for context parameter as an object to signing API.

### DIFF
--- a/application/src/main/java/net/corda/v5/application/crypto/SigningService.java
+++ b/application/src/main/java/net/corda/v5/application/crypto/SigningService.java
@@ -41,6 +41,10 @@ public interface SigningService {
     @NotNull
     DigitalSignature.WithKeyId sign(@NotNull byte[] bytes, @NotNull PublicKey publicKey, @NotNull SignatureSpec signatureSpec);
 
+    @Suspendable
+    @NotNull
+    DigitalSignature.WithKeyId sign(@NotNull byte[] bytes, @NotNull PublicKey publicKey, @NotNull SignatureSpec signatureSpec, @NotNull Map<String, String> context);
+
     /**
      * Looks into a set of signing keys to find keys owned by the caller. In case of {@link CompositeKey} it looks into
      * the composite key leaves and returns the firstly found owned composite key leaf.
--- a/application/src/main/java/net/corda/v5/application/crypto/SigningService.java
+++ b/application/src/main/java/net/corda/v5/application/crypto/SigningService.java
@@ -49,10 +49,7 @@ public interface SigningService {
      * primary identity, or previously generated via the freshKey method. If the {@link PublicKey} is actually
      * a {@link CompositeKey}, the first leaf signing key hosted by the node is used.
      * @param signatureSpec The {@link SignatureSpec} to use when producing this signature.
-     * @param context The execution context of the signing operation as a map of strings. Currently accepts the following parameters
-     * <ul>
-     *   <li>category: the type of signing key requested for additional verification</li>
-     * </ul>
+     * @param context The execution context of the signing operation.
      *
      * @return A {@link DigitalSignature.WithKeyId} representing the signed data and the {@link PublicKey} that belongs to the
      * same {@link KeyPair} as the {@link PrivateKey} that signed the data.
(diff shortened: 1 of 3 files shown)


davidcurrie (dismissed):
LGTM

application/src/main/java/net/corda/v5/application/crypto/SigningService.java
```diff
@@ -41,6 +41,28 @@ public interface SigningService {
...
+     * </ul>
+     *
+     * @return A {@link DigitalSignature.WithKeyId} representing the signed data and the {@link PublicKey} that belongs to the
+     * same {@link KeyPair} as the {@link PrivateKey} that signed the data.
+     *
+     * @throws CordaRuntimeException If the input key is not a member of {@code keys}.
+     */
+    @Suspendable
+    @NotNull
+    DigitalSignature.WithKeyId sign(@NotNull byte[] bytes, @NotNull PublicKey publicKey, @NotNull SignatureSpec signatureSpec, @NotNull Map<String, String> context);
```
davidcurrie: Is there a particular reason why we want the extensibility of using a `Map` here? If there is, should the allowed keys also form part of the API?

malachyb: The idea was to do that because if we later add another context parameter to the operation we would need to double the overloads and it would be better to have a generalised parameter to minimise how many changes we need to make.

davidcurrie: Better still to pass in an object with the required fields. That way you don't have to keep adding new versions of the operation but retain a well-defined API.

malachyb: Alright, I'll look into doing that instead.

dickon: The context parameter was something that @alexey-kadyrov-r3 and @mnesbit put in essentially at the start of the Corda 5 design. It's intentionally more dynamic so that we can add extra content without API versioning. Whether or not that's the best idea, we're basically locked down now on this since the API is considered frozen for incompatible changes.

davidcurrie: So, does it already exist elsewhere? I was under the impression it was something being added under this PR.

application/src/main/java/net/corda/v5/application/crypto/SigningServiceSignContext.java
```diff
@@ -0,0 +1,18 @@
...
+ * Context attached to the sign operation of SigningService
+ */
+@CordaSerializable
+public final class SigningServiceSignContext {
+    private final String keyCategory;
+
+    /**
+     * @param keyCategory the category of the key to be signed with
+     */
+    public SigningServiceSignContext(String keyCategory) {
```
dickon, line 15: I suppose if ever want to add additional fields in the future we can add additional public constructors in the future without breaking compatibility. However, I wonder if it'd be simpler to specify a parameter directly in an overloaded method.

application/src/main/java/net/corda/v5/application/crypto/SigningService.java
dickon: Discussed at length on #corda-platform-crypto. To summarise, I had failed to notice that we'd taken out all these context map<string,string> parameters a few months ago, so it would now be unprecedented. It does still exist in Avro messages but that's not really relevant to the question of good design of the API. I was also referred to https://r3-cev.atlassian.net/wiki/spaces/CB/pages/4082761773/Public+API+coding+standards but I think that's out of date, trying to confirm that now on #corda-platform. As it is I think Malachy's latest iteration  may now be good, though I'm also attracted to using a primitive string type for category on sign rather than a dataclass-like additional class.

dickon (dismissed):
LGTM

application/src/main/java/net/corda/v5/application/crypto/SigningServiceSignContext.java
Sakpal: @dickon is the above suggestion a blocker/major or a minor one?

dickon: It's just a minor observation, not a blocker.

davidcurrie (dismissed):
👍

relyafi (dismissed):
I know there's been some back and forth on how to pass in the key category, personally I'm not sure if `SigningServiceSignContext` is adding much value, but this still looks fine to me in terms of it satisfying our notary use-case.

dickon (approved):
LGTM

dickon (changes requested):
See discussion on #corda-platform-crypto - we need to clarify requirements.

PR #1256, opened 2023-09-20
CORE-16860 counterparty info refactor

LWogan:
To request counterparty info when it has not already been exchanged, send a SessionCounterpartyInfoRQ.
Include SessionInit payload to start the remote flow. 
If data has already been exchanged SessionCounterpartyInfoRQ wont be sent.
Respond with SessionCounterpartyInfoRS session event payload. Blank fields as contextSessionProperties are set on SessionEvent 
SequenceNumber will be set to null for both payload types. Only SessionData and SessionClose payloads will have sequence numbers set on SessionEvent

### DIFF
--- a/data/avro-schema/src/main/resources/avro/net/corda/data/flow/event/FlowEvent.avsc
+++ b/data/avro-schema/src/main/resources/avro/net/corda/data/flow/event/FlowEvent.avsc
@@ -11,7 +11,6 @@
       "name": "payload",
       "type": [
         "net.corda.data.flow.event.StartFlow",
-        "net.corda.data.flow.event.Wakeup",
         "net.corda.data.flow.output.FlowStatus",
         "net.corda.data.flow.event.SessionEvent",
         "net.corda.data.flow.event.external.ExternalEventResponse"
--- a/data/avro-schema/src/main/resources/avro/net/corda/data/flow/event/SessionEvent.avsc
+++ b/data/avro-schema/src/main/resources/avro/net/corda/data/flow/event/SessionEvent.avsc
@@ -50,8 +50,8 @@
     {
       "name": "payload",
       "type": [
-        "net.corda.data.flow.event.session.SessionInit",
-        "net.corda.data.flow.event.session.SessionConfirm",
+        "net.corda.data.flow.event.session.SessionCounterpartyInfoRQ",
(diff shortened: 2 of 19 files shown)


data/avro-schema/src/main/resources/avro/net/corda/data/flow/event/SessionEvent.avsc
```diff
@@ -50,8 +50,8 @@
     {
       "name": "payload",
       "type": [
-        "net.corda.data.flow.event.session.SessionInit",
```
mbrkic-r3, line 53: Do we need to remove https://github.com/corda/corda-api/blob/CORE-16860/counterparty-info/data/avro-schema/src/main/resources/avro/net/corda/data/flow/event/session/SessionInit.avsc as well?

data/avro-schema/src/main/resources/avro/net/corda/data/flow/event/FlowEvent.avsc
```diff
@@ -11,7 +11,6 @@
       "name": "payload",
       "type": [
         "net.corda.data.flow.event.StartFlow",
-        "net.corda.data.flow.event.Wakeup",
```
owenstanford, line 14: It's a remote edge case, but removing types like this could be a problem when upgrading from 5.0 to 5.1, need to check with @simon-johnson-r3, also worth adding an avro upgrade test to validate the change (see existing tests for an example)

data/avro-schema/src/main/resources/avro/net/corda/data/flow/event/SessionEvent.avsc
owenstanford, line 53: same as above

data/avro-schema/src/main/resources/avro/net/corda/data/flow/event/FlowEvent.avsc
JamesHR3: I think upgrade guarantees are just that you won't need to recompile your application. In particular I don't think we're guaranteeing that in flight flows will work across an upgrade, so any events of these types present in the system will need to be cleared before the upgrade can go ahead.

conal-smith (approved):

JamesHR3 (approved):

PR #1257, opened 2023-09-21
CORE-17257 - Reason column length is too short in RegistrationRequest table

nikinagy:
Resiliency test run failed caused by exception messages found in db worker logs. The exception messages suggested that the previous length of 255 was too short for the reason column.

### DIFF
--- a/data/db-schema/src/main/resources/net/corda/db/schema/vnode-vault/migration/vnode-vault-creation-v5.1.xml
+++ b/data/db-schema/src/main/resources/net/corda/db/schema/vnode-vault/migration/vnode-vault-creation-v5.1.xml
@@ -12,5 +12,10 @@
                 <constraints nullable="false" primaryKey="true"/>
             </column>
         </createTable>
+
+        <modifyDataType
+                columnName="reason"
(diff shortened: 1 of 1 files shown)


data/db-schema/src/main/resources/net/corda/db/schema/vnode-vault/migration/vnode-vault-creation-v5.1.xml
```diff
@@ -12,5 +12,10 @@
                 <constraints nullable="false" primaryKey="true"/>
             </column>
         </createTable>
+
+        <modifyDataType
+                columnName="reason"
+                newDataType="varchar(1000)"
```
yift-r3: Maybe we should change it to` TEXT`?

(If we do keep it as varchar, we need to add a code to trim the reason in the runtime code)

nikinagy: I think you are right. Text is better suited for long messages and we possibly have lengthy reasons at times.

yift-r3 (approved):

PR #1259, opened 2023-09-22
CORE-15396 Add support for an expiry timeout on a token claim

owenstanford:
Add support for an expiry timeout on a token claim, this is to support automatic cleanup of orphaned claims.

### DIFF
--- a/data/avro-schema/src/main/resources/avro/net/corda/data/ledger/utxo/token/selection/data/TokenClaim.avsc
+++ b/data/avro-schema/src/main/resources/avro/net/corda/data/ledger/utxo/token/selection/data/TokenClaim.avsc
@@ -10,6 +10,12 @@
         "doc": "Unique identifier for the claim"
       },
       {
+        "name": "claimTimestamp",
+        "type": ["null", "long"],
+        "default": null,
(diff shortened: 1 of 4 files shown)


data/avro-schema/src/main/resources/avro/net/corda/data/ledger/utxo/token/selection/data/TokenClaim.avsc
```diff
@@ -9,6 +9,12 @@
         "type": "string",
         "doc": "Unique identifier for the claim"
       },
+      {
+        "name": "claimTimestamp",
+        "type": ["null", "long"],
+        "default": null,
```
filipesoliveira, line 15: What does it mean if a claim does not have a timestamp? Will a claim with no timestamp ever expire?

owenstanford: in the runtime-os change that uses this, if it finds a old 5.0 version state it defaults the timestamp to the current time.

filipesoliveira (approved):

josephzunigadaly (approved):

PR #1260, opened 2023-09-25
CORE-16615: introduce new api to find an unconsumed states

jennyang-r3:
## Overview
This PR introduces a new API that finds unconsumed states given the state type. We are introducing this function even though there is `findUnconsumedStatesByType` because the function uses in-memory filtering, whereas this new function uses db query index.


ledger/ledger-utxo/src/main/java/net/corda/v5/ledger/utxo/UtxoLedgerService.java
```diff
@@ -87,7 +87,8 @@ public interface UtxoLedgerService {
     UtxoFilteredTransactionBuilder filterSignedTransaction(@NotNull UtxoSignedTransaction transaction);
 
     /**
-     * Finds unconsumed states of the specified {@link ContractState} type in the vault.
+     * Finds unconsumed states and subclasses of the states of the specified {@link ContractState} type in the vault.
+     * Only use this if you really care about catching all child classes, it has poor performance.
```
vlajos:
```suggestion
     * Finds unconsumed states of the specified {@link ContractState} type and its subclasses in the vault.
     * Only use this if you really care about catching all child classes. {@link findUnconsumedStatesByExactType} performs better for exact-matching use cases.
```
Sounds a bit more positive to me.

vlajos (dismissed):

vlajos (dismissed):

relyafi (dismissed):
LGTM

data/avro-schema/src/main/resources/avro/net/corda/data/ledger/persistence/FindUnconsumedVisibleExactStates.avsc
```diff
@@ -0,0 +1,13 @@
+{
+  "type": "record",
+  "name": "FindUnconsumedStatesByExactType",
+  "doc": "Retrieve the unconsumed visible states of specific type. This only retrieves exact type excluding its states of subclass unlike {@link FindUnconsumedStatesByExactType}. One of several types of ledger persistence request {@link LedgerPersistenceRequest}",
```
blsemo: I think you are comparing with the old `FindUnconsumedStatesByType` record.
```suggestion
  "doc": "Retrieve the unconsumed visible states of specific type. This only retrieves exact type excluding its states of subclass unlike {@link FindUnconsumedStatesByType}. One of several types of ledger persistence request {@link LedgerPersistenceRequest}",
```

blsemo (commented):
I think there's an error in the AVRO documentation.

data/avro-schema/src/test/kotlin/net/corda/data/ledger/persistence/LedgerPersistenceRequestSchemaCompatibilityTest.kt
```diff
@@ -50,6 +50,7 @@ class LedgerPersistenceRequestSchemaCompatibilityTest {
                 "net.corda.data.ledger.persistence.PersistTransactionIfDoesNotExist",
                 "net.corda.data.ledger.persistence.FindTransaction",
                 "net.corda.data.ledger.persistence.FindUnconsumedStatesByType",
+                "net.corda.data.ledger.persistence.FindUnconsumedStatesByExactType",
```
lankydan: I don't think we want changes to this file.

@vlajos I think you said we won't support 5.0 to 5.1 upgrades, therefore we _could_ make these changes in preparation of 5.1 to 5.2 upgrades. But, solely for the changes made in this PR, we don't need these.

ledger/ledger-utxo/src/main/java/net/corda/v5/ledger/utxo/UtxoLedgerService.java
```diff
@@ -87,7 +87,8 @@ public interface UtxoLedgerService {
     UtxoFilteredTransactionBuilder filterSignedTransaction(@NotNull UtxoSignedTransaction transaction);
 
     /**
-     * Finds unconsumed states of the specified {@link ContractState} type in the vault.
+     * Finds unconsumed states and subclasses of the states of the specified {@link ContractState} type and its subclasses in the vault.
+     * Only use this if you really care about catching all child classes. {@link #findUnconsumedStatesByExactType(Class<T>)} performs better for exact-matching use cases.
```
lankydan:
```
* Finds unconsumed states and subclasses of the states of the specified {@link ContractState} type and its subclasses in the vault.
* Only use this if you really care about catching all child classes. {@link #findUnconsumedStatesByExactType(Class<T>)} performs better for exact-matching use 
```

to

```
* Finds unconsumed states that are concrete implementations or subclasses of {@code type}. 
* <p>
* Only use this method if subclasses of {@code type} must be returned.
* <p>
* Use {@link #findUnconsumedStatesByExactType(Class<T>)} to return exact instances of the input {@code type}. This method is more performant than {@link #findUnconsumedStatesByType(Class<T>)}.
* <p>
* Use {@link #query(String, Class)} for a more performant method of retrieving subclasses of a specified type.
```

What do you think?

data/avro-schema/src/test/kotlin/net/corda/data/ledger/persistence/LedgerPersistenceRequestSchemaCompatibilityTest.kt
vlajos: AFAIK only 5.0-5.1 flow communication is not a requirement at the moment, but still, I think 
you are right.
`oldSchemaJson` should have the 5.0 schema. The new API is not a 5.0 one.

lankydan (approved):

blsemo (approved):
LGTM

PR #1261, opened 2023-09-25
CORE-12411 - Add reason for user in decline registration command

dimosr:
This change adds a reason that is communicated back to the user for declined registrations. It also adjusts the schema of the message that communicates the status back to the member to incorporate this reason. 

Used by: https://github.com/corda/corda-runtime-os/pull/4696


yift-r3 (dismissed):
LGTM

charlieR3 (approved):

PR #1263, opened 2023-09-26
CORE-9031: Add forceUpgrade parameter to VirtualNodeUpgradeRequest

Tom-Fitzpatrick:
This adds an optional forceUpgrade field to the VirtualNodeUpgradeRequest schema which defaults to false. This supports the work being done in https://github.com/corda/corda-runtime-os/pull/4734.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -9,7 +9,7 @@ cordaProductVersion = 5.1.0
 # NOTE: update this each time this module contains a breaking change
 ## NOTE: currently this is a top level revision, so all API versions will line up, but this could be moved to
 ##   a per module property in which case module versions can change independently.
-cordaApiRevision = 26
+cordaApiRevision = 27
 
 # Main
 kotlinVersion = 1.8.21
--- a/gradle.properties
+++ b/gradle.properties
@@ -9,7 +9,7 @@ cordaProductVersion = 5.1.0
 # NOTE: update this each time this module contains a breaking change
 ## NOTE: currently this is a top level revision, so all API versions will line up, but this could be moved to
 ##   a per module property in which case module versions can change independently.
-cordaApiRevision = 27
+cordaApiRevision = 26
 
 # Main
 kotlinVersion = 1.8.21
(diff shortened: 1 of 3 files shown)


gradle.properties
```diff
@@ -9,7 +9,7 @@ cordaProductVersion = 5.1.0
 # NOTE: update this each time this module contains a breaking change
 ## NOTE: currently this is a top level revision, so all API versions will line up, but this could be moved to
 ##   a per module property in which case module versions can change independently.
-cordaApiRevision = 26
+cordaApiRevision = 27
```
vkolomeyko: Is it a breaking change though?

I wonder if Avro will do defaulting for us automatically if the recipient has a new schema and payload does not have the value for the field?

FAO: @simon-johnson-r3

Tom-Fitzpatrick: I tried without the version bump, to see if I could get a green build in runtime-os without it. But the corda-api PR build here fails first because it tests the PR with corda-runtime-os code. The generated java file for VirtualNodeUpgradeRequest doesn't provide a constructor without forceUpgrade, and we get the following error when trying to use it:

`
e: file:///var/jenkins/workspace/ion-compatibility_release_os_5.1/components/virtual-node/virtual-node-rest-service-impl/src/main/kotlin/net/corda/virtualnode/rest/impl/v1/VirtualNodeRestResourceImpl.kt:348:41 None of the following functions can be called with the arguments supplied: 
  public constructor VirtualNodeUpgradeRequest() defined in net.corda.data.virtualnode.VirtualNodeUpgradeRequest |  
  public constructor VirtualNodeUpgradeRequest(virtualNodeShortHash: String!, cpiFileChecksum: String!, actor: String!, forceUpgrade: Boolean!) defined in net.corda.data.virtualnode.VirtualNodeUpgradeRequest
`

I don't think there's a good way around this, so I think the version bump is necessary.

vkolomeyko: Yes, I agree - looks like we will not be able o workaround without version bump.

vkolomeyko (dismissed):
LGTM

I have also reviewed the `corda-runtime-os` PR and it looks fine. So, tomorrow, we can start integrating both PRs.

josephzunigadaly (dismissed):

josephzunigadaly (approved):

vkolomeyko (approved):
LGTM

PR #1265, opened 2023-09-26
CORE-17360: Add topics for mapper scheduled cleanup and update cleanup Avro records

JamesHR3:
**Problem statement**

As part of ongoing performance work, there is a requirement to move the storage of flow mapper states to a shared state store. This will break the current implementation of mapper state cleanup. Flow mapper states are currently cleaned up using a scheduler implemented in each flow mapper, which relies on states being partitioned across the set of mapper instances. This will no longer be happening.

This problem can be addressed using the new task scheduler implemented in the DB worker, but doing so will require a few new topics for the mapper to listen to.

**Solution**

Add two new topics. The first is the topic that scheduled tasks for the flow mapper will be written to. The second is a dedicated cleanup event topic that will be written to as part of processing the scheduled event.

In addition, the `ExecuteCleanup` event has been modified to take a list of IDs as a field. This will allow the processor of the scheduled task to batch up the cleanup requests being sent to the mapper workers.

### DIFF
--- a/data/avro-schema/src/main/resources/avro/net/corda/data/flow/event/mapper/ExecuteCleanup.avsc
+++ b/data/avro-schema/src/main/resources/avro/net/corda/data/flow/event/mapper/ExecuteCleanup.avsc
@@ -4,5 +4,13 @@
   "docs": "When this event is processed the flow mapper state should be set to null",
   "namespace": "net.corda.data.flow.event.mapper",
   "fields": [
+    {
+      "name": "ids",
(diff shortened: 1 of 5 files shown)


jujoramos (commented):
LGTM, will approve once tests have been fixed 👍 .

jujoramos (approved):

conal-smith (approved):

jujoramos (approved):
LGTM

PR #1270, opened 2023-09-28
CORE-16183 Ledger data model simplifications (part 1)

nkovacsx:
### Overview

This PR contains the `AGREED` suggestions from this page: https://r3-cev.atlassian.net/wiki/spaces/CB/pages/4578214028/UTXO+Ledger+Data+Model+Simplifications#Remove-created-field-from-utxo_transaction_component-table-Agreed.

These are:
- Merge utxo_transaction_status table into utxo_transaction (then delete the `utxo_transaction_status` table) and rename utxo_transaction_output table to utxo_visible_transaction_output
- Merge utxo_visible_transaction_state table into utxo_transaction_output (then delete the `utxo_visible_transaction_state` table)
- Removal of utxo_transaction_sources table 
- Removal of utxo_transaction_cpk and utxo_cpk tables 
- Remove created field from utxo_transaction_component table 

### Test plan

1. 5.0 -> 5.1 
The upgrade SQL file was generated using the Corda CLI tool with the latest plugin. The SQL generated then was executed in a live Postgres database. The existing transactions were moved according to the 5.1 schema accordingly.

2. 5.1 Postgres
Started up a 5.1 instance with a fresh Postgres instance and schema creation ran successfully.

### DIFF
--- a/data/db-schema/src/main/resources/net/corda/db/schema/vnode-vault/migration/ledger-utxo-creation-v5.1.xml
+++ b/data/db-schema/src/main/resources/net/corda/db/schema/vnode-vault/migration/ledger-utxo-creation-v5.1.xml
@@ -3,11 +3,6 @@
                    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
                    xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-4.3.xsd">
 
-    <property name="json.column.type" value="JSONB" dbms="postgresql"/>
-
-    <!-- Please note that HSQLDB schema is supported for integration test purposes only. -->
-    <property name="json.column.type" value="CLOB" dbms="hsqldb"/>
-
     <changeSet author="R3.Corda" id="ledger-utxo-creation-v5.1">
         <createIndex indexName="utxo_visible_transaction_state_idx_consumed"
                      tableName="utxo_visible_transaction_state">
--- a/data/db-schema/src/main/resources/net/corda/db/schema/vnode-vault/migration/ledger-utxo-creation-v5.1.xml
+++ b/data/db-schema/src/main/resources/net/corda/db/schema/vnode-vault/migration/ledger-utxo-creation-v5.1.xml
@@ -9,11 +9,6 @@
     <property name="json.column.type" value="CLOB" dbms="hsqldb"/>
 
     <changeSet author="R3.Corda" id="ledger-utxo-creation-v5.1">
-        <createIndex
-                indexName="utxo_visible_transaction_state_idx_consumed"
-                tableName="utxo_visible_transaction_state">
-            <column name="consumed"/>
-        </createIndex>
 
         <createIndex
                 indexName="utxo_transaction_output_idx_type"
@@ -89,6 +84,13 @@
 
         <!-- Rename utxo_transaction_output to utxo_visible_transaction_output -->
         <renameTable newTableName="utxo_visible_transaction_output" oldTableName="utxo_transaction_output"/>
+
+        <!-- Re-create the index on the consumed column in the new table -->
+        <createIndex
+                indexName="utxo_visible_transaction_output_idx_consumed"
+                tableName="utxo_visible_transaction_output">
+            <column name="consumed"/>
+        </createIndex>
 
         <!-- Drop the created column from the utxo_transaction_component table -->
         <dropColumn columnName="created" tableName="utxo_transaction_component"/>
--- a/data/db-schema/src/main/resources/net/corda/db/schema/vnode-vault/migration/ledger-utxo-creation-v5.1.xml
+++ b/data/db-schema/src/main/resources/net/corda/db/schema/vnode-vault/migration/ledger-utxo-creation-v5.1.xml
@@ -15,6 +15,94 @@
                 tableName="utxo_transaction_output">
             <column name="type"/>
         </createIndex>
+
+    <property name="json.column.type" value="JSONB" dbms="postgresql"/>
+
+    <!-- Please note that HSQLDB schema is supported for integration test purposes only. -->
+    <property name="json.column.type" value="CLOB" dbms="hsqldb"/>
+
+    <changeSet author="R3.Corda" id="ledger-utxo-creation-v5.2">
+
+        <!-- Add new status column to utxo_transaction -->
+        <addColumn tableName="utxo_transaction">
+            <column name="status" type="VARCHAR(255)">
+                <constraints nullable="true"/>
+            </column>
+            <column name="updated" type="TIMESTAMP">
+                <constraints nullable="true"/>
+            </column>
+        </addColumn>
+
+        <!-- Migrate data from the old status table -->
+        <update tableName="utxo_transaction">
+            <column name="status"
+                    valueComputed="(select status from utxo_transaction_status uts where uts.transaction_id=id)"/>
+            <column name="updated"
+                    valueComputed="(select updated from utxo_transaction_status uts where uts.transaction_id=id)"/>
+        </update>
+
+        <!-- Drop the old status table -->
+        <dropTable tableName="utxo_transaction_status"/>
+
+        <!-- After status data migration we can make the status/updated column non-nullable -->
+        <addNotNullConstraint tableName="utxo_transaction" columnName="status" columnDataType="VARCHAR(255)"/>
+        <addNotNullConstraint tableName="utxo_transaction" columnName="updated" columnDataType="TIMESTAMP"/>
+
+        <!-- Add new custom_representation and consumed columns to utxo_transaction_output -->
+        <addColumn tableName="utxo_transaction_output">
+            <column name="custom_representation" type="${json.column.type}">
+                <constraints nullable="true"/>
+            </column>
+            <column name="consumed" type="TIMESTAMP">
+                <constraints nullable="true"/>
+            </column>
+        </addColumn>
+
+        <!--
+        Migrate data from the old visible states table
+        NOTE: We can't use <update> here because both tables have the same column naming
+        (transaction_id, group_idx etc.) and the SQL statement generated will not work in
+        Postgres, and we have no way to alias the base table.
+        -->
+        <sql>
+            UPDATE utxo_transaction_output txo
+            SET
+            custom_representation = (
+            SELECT custom_representation from utxo_visible_transaction_state vts
+            WHERE vts.transaction_id=txo.transaction_id
+            AND vts.group_idx=txo.group_idx
+            AND vts.leaf_idx=txo.leaf_idx
+            ),
+            consumed = (
+            SELECT consumed from utxo_visible_transaction_state vts
+            WHERE vts.transaction_id=txo.transaction_id
+            AND vts.group_idx=txo.group_idx
+            AND vts.leaf_idx=txo.leaf_idx
+            );
+        </sql>
+
+        <!-- After visible states data migration we can make the custom_representation column non-nullable -->
+        <addNotNullConstraint tableName="utxo_transaction_output"
+                              columnName="custom_representation"
+                              columnDataType="${json.column.type}"/>
+
+        <!-- Drop the utxo_visible_transaction_state table after data migration -->
+        <dropTable tableName="utxo_visible_transaction_state"/>
+
+        <!-- Rename utxo_transaction_output to utxo_visible_transaction_output -->
+        <renameTable newTableName="utxo_visible_transaction_output" oldTableName="utxo_transaction_output"/>
+
+        <!-- Drop the created column from the utxo_transaction_component table -->
+        <dropColumn columnName="created" tableName="utxo_transaction_component"/>
+
+        <!-- Drop the utxo_transaction_sources table -->
+        <dropTable tableName="utxo_transaction_sources"/>
+
+        <!-- Drop the utxo_transaction_cpk table -->
+        <dropTable tableName="utxo_transaction_cpk"/>
+
+        <!-- Drop the utxo_cpk table -->
+        <dropTable tableName="utxo_cpk"/>
     </changeSet>
 
 </databaseChangeLog>
--- a/data/db-schema/src/main/resources/net/corda/db/schema/vnode-vault/migration/ledger-utxo-creation-v5.1.xml
+++ b/data/db-schema/src/main/resources/net/corda/db/schema/vnode-vault/migration/ledger-utxo-creation-v5.1.xml
@@ -56,7 +56,7 @@
         <dropTable tableName="utxo_transaction_status"/>
 
         <!-- After status data migration we can make the status/updated column non-nullable -->
-        <addNotNullConstraint tableName="utxo_transaction" columnName="status" columnDataType="VARCHAR(255)"/>
+        <addNotNullConstraint tableName="utxo_transaction" columnName="status" columnDataType="VARCHAR(16)"/>
         <addNotNullConstraint tableName="utxo_transaction" columnName="updated" columnDataType="TIMESTAMP"/>
 
         <!-- Add new custom_representation and consumed columns to utxo_transaction_output -->
--- a/data/db-schema/src/main/resources/net/corda/db/schema/vnode-vault/migration/ledger-utxo-creation-v5.1.xml
+++ b/data/db-schema/src/main/resources/net/corda/db/schema/vnode-vault/migration/ledger-utxo-creation-v5.1.xml
@@ -41,7 +41,7 @@
 
         <!-- Add new status column to utxo_transaction -->
         <addColumn tableName="utxo_transaction">
-            <column name="status" type="VARCHAR(255)">
+            <column name="status" type="VARCHAR(16)">
                 <constraints nullable="true"/>
             </column>
             <column name="updated" type="TIMESTAMP">
--- a/data/db-schema/src/main/resources/net/corda/db/schema/vnode-vault/migration/ledger-utxo-creation-v5.1.xml
+++ b/data/db-schema/src/main/resources/net/corda/db/schema/vnode-vault/migration/ledger-utxo-creation-v5.1.xml
@@ -84,16 +84,16 @@
             UPDATE utxo_transaction_output txo
             SET
             custom_representation = (
-            SELECT custom_representation from utxo_visible_transaction_state vts
-            WHERE vts.transaction_id=txo.transaction_id
-            AND vts.group_idx=txo.group_idx
-            AND vts.leaf_idx=txo.leaf_idx
+                SELECT custom_representation from utxo_visible_transaction_state vts
+                WHERE vts.transaction_id=txo.transaction_id
+                AND vts.group_idx=txo.group_idx
+                AND vts.leaf_idx=txo.leaf_idx
             ),
             consumed = (
-            SELECT consumed from utxo_visible_transaction_state vts
-            WHERE vts.transaction_id=txo.transaction_id
-            AND vts.group_idx=txo.group_idx
-            AND vts.leaf_idx=txo.leaf_idx
+                SELECT consumed from utxo_visible_transaction_state vts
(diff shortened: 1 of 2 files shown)


data/db-schema/src/main/resources/net/corda/db/schema/vnode-vault/migration/ledger-utxo-creation-v5.1.xml
```diff
@@ -15,6 +15,94 @@
...
+    <property name="json.column.type" value="JSONB" dbms="postgresql"/>
+
+    <!-- Please note that HSQLDB schema is supported for integration test purposes only. -->
+    <property name="json.column.type" value="CLOB" dbms="hsqldb"/>
+
+    <changeSet author="R3.Corda" id="ledger-utxo-creation-v5.2">
+
+        <!-- Add new status column to utxo_transaction -->
+        <addColumn tableName="utxo_transaction">
+            <column name="status" type="VARCHAR(255)">
```
vlajos: There were some discussions to change this to CHAR(1) or something similar since we put only 1 letter there.

data/db-schema/src/main/resources/net/corda/db/schema/vnode-vault/migration/ledger-utxo-creation-v5.1.xml
```diff
@@ -15,6 +15,94 @@
...
+                <constraints nullable="true"/>
+            </column>
+        </addColumn>
+
+        <!--
+        Migrate data from the old visible states table
+        NOTE: We can't use <update> here because both tables have the same column naming
+        (transaction_id, group_idx etc.) and the SQL statement generated will not work in
+        Postgres, and we have no way to alias the base table.
+        -->
```
vlajos, line 77: Maybe creating a temporary view with different field names?

data/db-schema/src/main/resources/net/corda/db/schema/vnode-vault/migration/ledger-utxo-creation-v5.1.xml
```diff
@@ -15,6 +15,94 @@
                 tableName="utxo_transaction_output">
             <column name="type"/>
         </createIndex>
+
+    <property name="json.column.type" value="JSONB" dbms="postgresql"/>
+
+    <!-- Please note that HSQLDB schema is supported for integration test purposes only. -->
+    <property name="json.column.type" value="CLOB" dbms="hsqldb"/>
+
+    <changeSet author="R3.Corda" id="ledger-utxo-creation-v5.2">
```
mbrkic-r3: Assuming this change is for ver 5.1, it should be part of the changeset above (otherwise in a different, 5.2 xml file)

data/db-schema/src/main/resources/net/corda/db/schema/vnode-vault/migration/ledger-utxo-creation-v5.1.xml
```diff
@@ -15,6 +15,94 @@
...
+                <constraints nullable="true"/>
+            </column>
+            <column name="consumed" type="TIMESTAMP">
+                <constraints nullable="true"/>
+            </column>
+        </addColumn>
+
+        <!--
+        Migrate data from the old visible states table
+        NOTE: We can't use <update> here because both tables have the same column naming
```
mbrkic-r3, line 74: How about this?
```
        <update tableName="utxo_transaction_output">
            <column name="custom_representation" valueComputed="
                (SELECT vts.custom_representation
                FROM utxo_visible_transaction_state vts
                WHERE vts.transaction_id = transaction_id
                AND vts.group_idx = group_idx
                AND vts.leaf_idx = leaf_idx)
            "/>
            <column name="consumed" valueComputed="
                (SELECT vts.consumed
                FROM utxo_visible_transaction_state vts
                WHERE vts.transaction_id = transaction_id
                AND vts.group_idx = group_idx
                AND vts.leaf_idx = leaf_idx)
            "/>
        </update>
```

data/db-schema/src/main/resources/net/corda/db/schema/vnode-vault/migration/ledger-utxo-creation-v5.1.xml
nkovacsx: Thanks, this was just a typo, I put these changes in a separate file first.

data/db-schema/src/main/resources/net/corda/db/schema/vnode-vault/migration/ledger-utxo-creation-v5.1.xml
nkovacsx: @mbrkic-r3 Made a suggestion, trying that now.

data/db-schema/src/main/resources/net/corda/db/schema/vnode-vault/migration/ledger-utxo-creation-v5.1.xml
nkovacsx: Yeah I thought about this too but as far as I know `VARCHAR` will be optimised if it's not the full length? Not sure.

data/db-schema/src/main/resources/net/corda/db/schema/vnode-vault/migration/ledger-utxo-creation-v5.1.xml
nkovacsx: I tried this but it seems like without somehow aliasing the base table it will not work.
It seems like the query generated by the upgrade plugin is the following:
```
SET SEARCH_PATH TO vnode_vault_37db0c9d2e17, "$user","public";
UPDATE vnode_vault_37db0c9d2e17.utxo_transaction_output SET 
	consumed = (
		SELECT vts.consumed FROM utxo_visible_transaction_state vts 
		WHERE vts.transaction_id = transaction_id                 
		AND vts.group_idx = group_idx                 
		AND vts.leaf_idx = leaf_idx
	), 
	custom_representation = (
		SELECT vts.custom_representation FROM utxo_visible_transaction_state vts 
		WHERE vts.transaction_id = transaction_id                 
		AND vts.group_idx = group_idx                 
		AND vts.leaf_idx = leaf_idx
	);
```
Executing this in the postgres DB I’m getting the following error:
```
SQL Error [21000]: ERROR: more than one row returned by a subquery used as an expression
```

I think that's because these lines:
`WHERE vts.transaction_id = transaction_id` are evaluated as `WHERE vts.transaction_id =vts.transaction_id` and returns every line.

nkovacsx: @vlajos ' suggestion was to create a temp view. It works like this:
```
<createView fullDefinition="false"
                    remarks="A view of utxo_visible_transaction_state with renamed fields"
                    replaceIfExists="false"
                    viewName="v_utxo_visible_transaction_state">
            SELECT
            transaction_id "vts_transaction_id",
            group_idx "vts_group_idx",
            leaf_idx "vts_leaf_idx",
            custom_representation "vts_custom_representation",
            created "vts_created",
            consumed "vts_consumed"
            FROM utxo_visible_transaction_state
        </createView>

        <update tableName="utxo_transaction_output">
            <column name="custom_representation" valueComputed="
                (SELECT vts_custom_representation
                FROM v_utxo_visible_transaction_state
                WHERE vts_transaction_id = transaction_id
                AND vts_group_idx = group_idx
                AND vts_leaf_idx = leaf_idx)
            "/>
            <column name="consumed" valueComputed="
                (SELECT vts_consumed
                FROM v_utxo_visible_transaction_state
                WHERE vts_transaction_id = transaction_id
                AND vts_group_idx = group_idx
                AND vts_leaf_idx = leaf_idx)
            "/>
        </update>

        <dropView viewName="v_utxo_visible_transaction_state"/>
```

But this seems a bit overkill for me but it might be a better solution than the raw SQL. Not sure what people think. @mbrkic-r3 @vlajos @relyafi @blsemo

relyafi: I'm no expert as I've not had to do this before, but from looking online it appears that raw SQL is the way to go? The above does seem quite complicated.

We'd presumably just need to ensure we're sticking to ANSI SQL.

nkovacsx: Yeah raw SQL is the best way I could find too. I just gave the view a try but not sure if it's worth it at all.

data/db-schema/src/main/resources/net/corda/db/schema/vnode-vault/migration/ledger-utxo-creation-v5.1.xml
nkovacsx: Discussed this with @vlajos offline and he is happy with the raw SQL approach as of now.

data/db-schema/src/main/resources/net/corda/db/schema/vnode-vault/migration/ledger-utxo-creation-v5.1.xml
mbrkic-r3: I agree, raw ANSI SQL seems a better choice. In this case it might be more readable to use a simpler query that doesn't use sub-selects.

data/db-schema/src/main/resources/net/corda/db/schema/vnode-vault/migration/ledger-utxo-creation-v5.1.xml
```diff
@@ -3,18 +3,104 @@
...
-       <createIndex
-           indexName="utxo_visible_transaction_state_idx_consumed"
-           tableName="utxo_visible_transaction_state">
-           <column name="consumed"/>
-       </createIndex>
+        <createIndex
+                indexName="utxo_visible_transaction_state_idx_consumed"
+                tableName="utxo_visible_transaction_state">
+            <column name="consumed"/>
+        </createIndex>
```
vlajos: This table gets dropped later.
We can remove this index from here.
Although we may need an equivalent to the data's new place.

nkovacsx: Good point, thanks.

vlajos (dismissed):

data/db-schema/src/main/resources/net/corda/db/schema/vnode-vault/migration/ledger-utxo-creation-v5.1.xml
relyafi: I believe `VARCHAR` will only use the space needed based on the actual length of the string, so this probably doesn't need to be exactly right. but ideally it would be in the right ballpark. I'd imagine this will only be 1-2 characters in practice even taking future statuses into account, so something like 16/32 would be more reasonable IMO.

(FWIW, I believe `CHAR` _can_ be slightly more efficient to index on etc. since it's always a fixed length, but I don't think that is worth worrying about right now).

data/db-schema/src/main/resources/net/corda/db/schema/vnode-vault/migration/ledger-utxo-creation-v5.1.xml
```diff
@@ -3,18 +3,106 @@
                    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
                    xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-4.3.xsd">
 
+    <property name="json.column.type" value="JSONB" dbms="postgresql"/>
+
+    <!-- Please note that HSQLDB schema is supported for integration test purposes only. -->
+    <property name="json.column.type" value="CLOB" dbms="hsqldb"/>
```
relyafi: Are these not already available via `ledger-utxo-visible-transaction-state-creation-v1.0.xml`?

data/db-schema/src/main/resources/net/corda/db/schema/vnode-vault/migration/ledger-utxo-creation-v5.1.xml
```diff
@@ -3,18 +3,106 @@
...
+
     <changeSet author="R3.Corda" id="ledger-utxo-creation-v5.1">
```
relyafi: According to the [Liquibase Best Practice](https://www.liquibase.org/get-started/best-practices) guide, we should generally do a single change per changeset. I believe that is what we generally do elsewhere (and certainly what we did for the notary in C4).

data/db-schema/src/main/resources/net/corda/db/schema/vnode-vault/migration/ledger-utxo-creation-v5.1.xml
```diff
@@ -3,18 +3,106 @@
...
+        Postgres, and we have no way to alias the base table.
+        -->
+        <sql>
+            UPDATE utxo_transaction_output txo
+            SET
+            custom_representation = (
+            SELECT custom_representation from utxo_visible_transaction_state vts
+            WHERE vts.transaction_id=txo.transaction_id
+            AND vts.group_idx=txo.group_idx
+            AND vts.leaf_idx=txo.leaf_idx
```
relyafi: Might be a bit more readable to indent this one level (same for `consumed`)

relyafi (commented):
Some minor comments but broadly looks good, final approval should be deferred until corresponding `corda-runtime-os` changes are also ready for review and approved.

I think it might be worth clarifying the test scenarios in the test notes, what I'd like to see is:

- Able to deploy a "fresh" Postgres based Corda deployment
- Able to run up a "fresh" simulator deployment (or the combined worker) with HSQLDB. I'm assuming no upgrade concerns with HSQLDB as we'd always use a fresh deployment given its in-memory nature
- Able to deploy a Postgres based Corda deployment running 5.0, generate some transactions, and then upgrade to 5.1. Ideally we'd have an integration test for this in runtime-os (though not sure if our test framework is mature enough to do this yet)

data/db-schema/src/main/resources/net/corda/db/schema/vnode-vault/migration/ledger-utxo-creation-v5.1.xml
nkovacsx: I just copied other changesets to be honest. 😅 `vnode-vault-creation` has tons of changes in one big changeset. I thought that's the practice we are following.

data/db-schema/src/main/resources/net/corda/db/schema/vnode-vault/migration/ledger-utxo-creation-v5.1.xml
nkovacsx: Good point, thanks!

data/db-schema/src/main/resources/net/corda/db/schema/vnode-vault/migration/ledger-utxo-creation-v5.1.xml
nkovacsx: I remember that one change per changeset used to be the practice for C4 but it seems like we moved away from that (not sure if intentionally)?

relyafi: The rationale is that each changeset represents an atomic step that can be rolled back independently if possible, the more complexity / steps in a changeset, the more scope there is for things to go wrong.

If you remember this was particularly important when we were working with CockroachDB because that cannot run changsets in transactions IIRC - so it was super important to keep them simple to minimise the chance of failure (which would then potentially need a manual fix)

data/db-schema/src/main/resources/net/corda/db/schema/vnode-vault/migration/ledger-utxo-creation-v5.1.xml
```diff
@@ -3,18 +3,106 @@
                    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
```
vlajos, line 3: (It is not related to this change.)
But the `creation` in the file name is not really accurate.
Maybe `migration`?
(But that can be done separately if needed at all.)

data/db-schema/src/main/resources/net/corda/db/schema/vnode-vault/migration/ledger-utxo-creation-v5.1.xml
lankydan: Are we changing this part based on @relyafi's comment or not? It doesn't look like it has changed at least, I don't have strong opinions myself, so happy to go either way.

lankydan (commented):
Changes generally look fine to me, left one comment on another comment though.

One other thing

> These are:
>
> Merge utxo_transaction_status table into utxo_transaction (then delete the utxo_transaction_status table)
> Merge utxo_visible_transaction_state table into utxo_transaction_output (then delete the utxo_visible_transaction_state table)
> Removal of utxo_transaction_sources table
> Removal of utxo_transaction_cpk and utxo_cpk tables
> Remove created field from utxo_transaction_component table
> Rename utxo_transaction_output table to utxo_visible_transaction_output

Can we either specify these as steps or only say what the outcome is.

Mainly seeing "Merge utxo_visible_transaction_state table into utxo_transaction_output" and then "utxo_transaction_output table to utxo_visible_transaction_output" confused me when reading through the change.

If it was "Merge utxo_visible_transaction_state table into utxo_transaction_output and rename it to utxo_visible_transaction_output" it might be clearer. This is nit-picky though.

data/db-schema/src/main/resources/net/corda/db/schema/vnode-vault/migration/ledger-utxo-creation-v5.1.xml
nkovacsx: I think the idea is that we merge this as-is then rework how changesets are defined in the whole codebase (other areas of the db as well) in a separate PR.

data/db-schema/src/main/resources/net/corda/db/schema/vnode-vault/migration/ledger-utxo-creation-v5.1.xml
nkovacsx: I haven't tried it yet tbh, will give it a go

nkovacsx: just did a start-up and it seems like the schema is created without these properties in this XML so will remove. I suppose they are already present from the other changeset as you said.

vlajos (dismissed):

lankydan (dismissed):

relyafi (dismissed):

data/db-schema/src/main/resources/net/corda/db/schema/vnode-vault/migration/ledger-utxo-creation-v5.1.xml
```diff
@@ -33,6 +33,94 @@
...
+                    valueComputed="(select status from utxo_transaction_status uts where uts.transaction_id=id)"/>
+            <column name="updated"
+                    valueComputed="(select updated from utxo_transaction_status uts where uts.transaction_id=id)"/>
+        </update>
+
+        <!-- Drop the old status table -->
+        <dropTable tableName="utxo_transaction_status"/>
+
+        <!-- After status data migration we can make the status/updated column non-nullable -->
+        <addNotNullConstraint tableName="utxo_transaction" columnName="status" columnDataType="VARCHAR(255)"/>
```
blsemo: Doesn't this also change the type from `VARCHAR(16)` to `VARCHAR(255)`?

nkovacsx: Updated thanks!

vlajos (approved):

blsemo (approved):
LGTM

PR #1271, opened 2023-09-29
CORE-17386 - Create new topic for flow processor scheduled tasks.

driessamyn:
- Create new topic.
- Also create constants for name of the task.
- Fix unit test logic for which the assert was the wrong way around.

Runtime-os PR: https://github.com/corda/corda-runtime-os/pull/4740

### DIFF
--- a/data/topic-schema/src/main/java/net/corda/schema/Schemas.java
+++ b/data/topic-schema/src/main/java/net/corda/schema/Schemas.java
@@ -267,7 +267,9 @@ public final class Schemas {
     public static final class ScheduledTask {
         private ScheduledTask() {}
 
+        // TODO - duplicated to ease migration from first to second. First can be removed once integrated with runtime-os
         public static final String SCHEDULED_TASK_DB_PROCESSOR = "scheduled.task.db.processor";
+        public static final String SCHEDULED_TASK_TOPIC_DB_PROCESSOR = "scheduled.task.db.processor";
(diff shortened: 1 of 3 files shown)


data/topic-schema/src/main/java/net/corda/schema/Schemas.java
```diff
@@ -269,5 +269,8 @@ private ScheduledTask() {}
 
         public static final String SCHEDULED_TASK_DB_PROCESSOR = "scheduled.task.db.processor";
         public static final String SCHEDULED_TASK_MAPPER_PROCESSOR = "scheduled.task.mapper.processor";
+        public static final String SCHEDULED_TASK_TOPIC_FLOW_PROCESSOR = "scheduled.task.flow.processor";
```
kyriathar, line 275:
```suggestion
        public static final String SCHEDULED_TASK_FLOW_PROCESSOR_TOPIC = "scheduled.task.flow.processor";
```

I think we should do the same for the 2 above constants for consistency with rest of the cases (I missed that with `SCHEDULED_TASK_DB_PROCESSOR`).

driessamyn: Sure. I can add it in this PR if you want?

JamesHR3 (approved):

jujoramos (approved):
LGTM

kyriathar (approved):

PR #1272, opened 2023-10-02
CORE-17387 - Flow timeout events.

driessamyn:
Topic and Avro schema for flow timeout events.

### DIFF
--- a/data/avro-schema/src/main/resources/avro/net/corda/data/flow/FlowTimeout.avsc
+++ b/data/avro-schema/src/main/resources/avro/net/corda/data/flow/FlowTimeout.avsc
@@ -1,0 +1,20 @@
+{
+  "type": "record",
+  "name": "FlowTimeout",
+  "namespace": "net.corda.data.flow",
+  "doc": "Represents of a flow timeout instruction.",
+  "fields": [
+      {
+        "name": "checkpointStateKey",
+        "type": "string",
+        "doc": "Key for the state record that is storing the checkpoint."
+      },
+      {
+        "name": "timeoutDateTime",
+        "type": {
+            "type": "long",
+            "logicalType": "timestamp-millis"
+          }
+      }
+    ]
+}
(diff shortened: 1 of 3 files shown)


data/avro-schema/src/main/resources/avro/net/corda/data/flow/FlowTimeout.avsc
```diff
@@ -0,0 +1,20 @@
+{
+  "type": "record",
+  "name": "FlowTimeout",
+  "namespace": "net.corda.data.flow",
+  "doc": "Represents of a flow timeout instruction.",
+  "fields": [
+      {
+        "name": "checkpointStateKey",
```
driessamyn, line 8: As I understand, this is the flow ID, but I thought maybe it's worth being specific in case this changes/evolves.

data/avro-schema/src/main/resources/avro/net/corda/data/flow/FlowTimeout.avsc
```diff
@@ -0,0 +1,20 @@
...
+        "type": "string",
+        "doc": "Key for the state record that is storing the checkpoint."
+      },
+      {
+        "name": "timeoutDateTime",
```
driessamyn, line 13: This is probably not strictly necessary but may be useful when troubleshooting or monitoring the system health?

data/avro-schema/src/main/resources/avro/net/corda/data/flow/FlowTimeout.avsc
JamesHR3: Think there's no harm in having it.

data/avro-schema/src/main/resources/avro/net/corda/data/flow/FlowTimeout.avsc
JamesHR3: We're not great in general at using Avro records for this sort of thing, but it does give us a chance to log at the point where we receive the event.

JamesHR3 (approved):

mbrkic-r3 (approved):
LGTM

PR #1273, opened 2023-10-02
CORE-17359: Rename topic for flow mapper scheduled task and add a scheduled task name

JamesHR3:
### Problem statement

The flow mapper requires a topic and a scheduled task name to perform cleanup of the mapper states once the cleanup time window expires. The previous name of the topic constant was not clear, and the task name did not previously exist.

### Solution

Clean up the topic name and add a scheduled task name.

### DIFF
--- a/data/topic-schema/src/main/java/net/corda/schema/Schemas.java
+++ b/data/topic-schema/src/main/java/net/corda/schema/Schemas.java
@@ -267,13 +267,12 @@ public final class Schemas {
     public static final class ScheduledTask {
         private ScheduledTask() {}
 
-        // TODO - duplicated to ease migration from first to second. First can be removed once integrated with runtime-os
-        public static final String SCHEDULED_TASK_DB_PROCESSOR = "scheduled.task.db.processor";
         public static final String SCHEDULED_TASK_TOPIC_DB_PROCESSOR = "scheduled.task.db.processor";
         public static final String SCHEDULED_TASK_NAME_DB_PROCESSOR = "deduplication-table-clean-up-task";
-        public static final String SCHEDULED_TASK_MAPPER_PROCESSOR = "scheduled.task.mapper.processor";
+        public static final String SCHEDULED_TASK_TOPIC_MAPPER_PROCESSOR = "scheduled.task.mapper.processor";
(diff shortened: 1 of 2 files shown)


jujoramos (dismissed):

data/topic-schema/src/main/java/net/corda/schema/Schemas.java
```diff
@@ -268,13 +268,12 @@ private VirtualNode() {
...
-        // TODO - duplicated to ease migration from first to second. First can be removed once integrated with runtime-os
-        public static final String SCHEDULED_TASK_DB_PROCESSOR = "scheduled.task.db.processor";
         public static final String SCHEDULED_TASK_TOPIC_DB_PROCESSOR = "scheduled.task.db.processor";
         public static final String SCHEDULED_TASK_NAME_DB_PROCESSOR = "deduplication-table-clean-up-task";
-        public static final String SCHEDULED_TASK_MAPPER_PROCESSOR = "scheduled.task.mapper.processor";
+        public static final String SCHEDULED_TASK_TOPIC_MAPPER_PROCESSOR = "scheduled.task.mapper.processor";
+        public static final String SCHEDULED_TASK_NAME_MAPPER_CLEANUP = "flow-mapper-state-cleanup";
         public static final String SCHEDULED_TASK_TOPIC_FLOW_PROCESSOR = "scheduled.task.flow.processor";
-        public static final String SCHEDULED_TASK_NAME_SESSION_TIMEOUT = "flow-session-timout";
+        public static final String SCHEDULED_TASK_NAME_SESSION_TIMEOUT = "flow-session-timeout";
```
jujoramos, line 276: Any reason for using `-` instead of `.` as the separator within the topic name for these topics?, wouldn't it be better to keep consistency with the others?. Not a blocking concern, just out of curiosity.

jujoramos (approved):
LGTM

data/topic-schema/src/main/java/net/corda/schema/Schemas.java
JamesHR3: They're not topics, they're names of scheduled tasks. I think mixing these up is perhaps a little questionable, but not to the extent where I'd want to change it here.

jujoramos: 👍

PR #1274, opened 2023-10-03
ES-1398 Pipeline integration with corda-api scanner PR gates

jacob-scott3141:
Adds internal gradle plugin dependency and YAML files to track API changes with @tomstark99 's API Scanner

Shared pipeline changes add the following comments on Corda-API PR's:

- On Failed Scan:
Screenshot 2023-10-04 150333
- On Successful Scan:
Screenshot 2023-10-04 172634

# PR Checklist:

- [x] Have you run the unit, integration and smoke tests as described [here](https://docs.corda.net/head/testing.html)?
- [ ] ~~If you added public APIs, did you write the JavaDocs?~~
- [ ] ~~If the changes are of interest to application developers, have you added them to the [changelog](https://docs.corda.net/head/changelog.html) (`/docs/source/changelog.rst`), and potentially the [release notes](https://docs.corda.net/head/release-notes.html) (`/docs/source/release-notes.rst`)?~~
- [ ] ~~If you are contributing for the first time, please read the [contributor agreement](https://docs.corda.net/head/contributing.html) now and add a comment to this pull request stating that your PR is in accordance with the [Developer's Certificate of Origin](https://docs.corda.net/head/contributing.html#merging-the-changes-back-into-corda).~~

Thanks for your code, it's appreciated! :)


.ci/JenkinsApiCompatibility
```diff
@@ -1,5 +1,5 @@
 // Check corda-api compatibility with downstream consumers which implement CordApps
-@Library('corda-shared-build-pipeline-steps@5.1') _
+@Library('corda-shared-build-pipeline-steps@jacob/ES-1398') _
```
tomstark99: I think we'll have to wait for this pipeline change to be merged into 5.1 (in corda-shared-build-pipeline-steps) and not change this here

Jenkinsfile
```diff
@@ -1,4 +1,4 @@
-@Library('corda-shared-build-pipeline-steps@5.1') _
+@Library('corda-shared-build-pipeline-steps@jacob/ES-1398') _
```
tomstark99: see comment above

.ci/JenkinsApiCompatibility
jacob-scott3141: Thanks, once the pipeline PR is merged I'll change this and mark as ready for review

application/src/api/corda-application-5.1.0.yaml
driessamyn, line 1: I'm not sure `src` is the right directory for this. 
Would it be better to place this file in the project root? (e.g. in this example in `application/` ?

driessamyn, line 1: With the version added to the file, I'm assuming that we would generate a new one for each increment and keep the existing?
In that case, would it be better to use the 2 digits only (i.e. in this case `5.1` rather than `5.1.0`)? We would never accept an API change in a patch release.

application/src/api/corda-application-5.1.0.yaml
```diff
@@ -0,0 +1,1898 @@
+net.corda.v5.application.crypto.CompositeKeyGenerator:
```
driessamyn: Nitpick - is it possible to add a comment in line 1 that says something like?
```
-- Auto-generated by gradle cementAPI task
-- Do NOT edit manually
```

buildSrc/src/main/groovy/corda.api-scanner.gradle
```diff
@@ -0,0 +1,4 @@
+// Apply api-scanner
+if (System.getenv('CORDA_ARTIFACTORY_USERNAME') != null || project.hasProperty('cordaArtifactoryUsername')) {
+    project.pluginManager.apply('net.corda.internal.gradle.plugins.api-scanner')
```
driessamyn, line 3: Question maybe for @ronanbrowne .

I'm guessing this means Open Source users won't be able to run this, which is probably fine, but will this result in an error if they try to do a normal "build"?

data/config-schema/build.gradle
```diff
@@ -2,6 +2,7 @@ plugins {
     id 'corda-api.common-library'
     id 'corda.common-publishing'
     id 'corda.java-only'
+    id 'corda.api-scanner'
```
driessamyn: I don't think this project needs to be included.
The JSON schemas themselves are what is public API.

data/avro-schema/build.gradle
```diff
@@ -7,6 +7,7 @@ plugins {
     id 'corda.common-publishing'
     id 'corda.java-only'
     id "com.github.davidmc24.gradle.plugin.avro-base"
+    id 'corda.api-scanner'
```
driessamyn: I don't think this project needs to be includes as the AVRO schemas themselves are the public API.

data/db-schema/build.gradle
```diff
@@ -2,6 +2,7 @@ plugins {
     id 'corda-api.common-library'
     id 'corda.common-publishing'
     id 'corda.java-only'
+    id 'corda.api-scanner'
```
driessamyn: Similar to config and avro, I don't think this project needs to be included.

data/rbac-schema/build.gradle
```diff
@@ -2,6 +2,7 @@ plugins {
     id 'corda-api.common-library'
     id 'corda.common-publishing'
     id 'corda.java-only'
+    id 'corda.api-scanner'
```
driessamyn: As above, I don't think this module needs to be included.

data/topic-schema/build.gradle
```diff
@@ -2,6 +2,7 @@ plugins {
     id 'corda-api.common-library'
     id 'corda.common-publishing'
     id 'corda.java-only'
+    id 'corda.api-scanner'
```
driessamyn: As above, I don't think this module needs to be included.

driessamyn (commented):
Mostly minor comments, and one more nit: would it be possible to include the `apiDiff` in the `verification` group and a dependency of the `check` task?

buildSrc/src/main/groovy/corda.api-scanner.gradle
jacob-scott3141: Not having credentials won't prevent normal builds, only the api export, diff, and cement

ronanbrowne: @driessamyn a normal build `./gradlew build` or any other task wont trigger the above, only issue is if a non-r3 user tries to use any of the tasks provided by the plugin directly, which as this is a internal tool, my presumption is we don't expect them to.

So the code here ensures we can compile for OS users

serialization/src/api/corda-serialization-5.1.0.yaml
```diff
@@ -0,0 +1,123 @@
+net.corda.v5.serialization.MissingSerializerException:
```
simon-johnson-r3: Some of the things in here are not part of the public api, so I assume the whole module isn't. Is that correct @driessamyn ? If so we can remove the scanning from here.

For example `SingletonSerializeAsToken`.

application/src/api/corda-application-5.1.0.yaml
simon-johnson-r3: The point of these files is that we don't make breaking changes between say 5.1 and 5.2. If we create a new file for 5.2, how do we compare it with 5.1 to check we haven't made breaking changes?

What I'm really asking is how are these versioned files used in the process of automated api break checking?

jacob-scott3141: The version from the `cordaProductVersion` in gradle.properties is what gets appended to the file name, but I don't think the scan relies on the version when looking at differences

jacob-scott3141: The plugin checks the difference between current API and the previous version that has been cemented. So, we cement the API in this PR and then any future PRs will be compared to the state of the API now, and if there are breaking changes, there will be a GitHub comment detailing the breakage. If these changes are intentional, then the user will run `./gradlew cementApi`, which will change the content of the yaml files, triggering a review from corda5 team leads.

jacob-scott3141: The CODEOWNERS file is looking for all yaml within a specific directory - what about `<sub-project>/scans/*.yaml` ?

simon-johnson-r3: Yes, this was really a more general question about process (with the intent of ensuring the process we want matches the tool). So for example when 5.2 is branched, we'll get a 5.2 yaml file. And every api change we'll review in the PR the changes to the 5.2 yaml file and someone will approve.

What role does the 5.1 yaml play in the process? Do we want for example an "end of release" review that all the changes in the 5.2 file are ok when compared with the 5.1 file? That might be quite involved, although perhaps we take that hit because it is quite important to check.

If we don't want to do that check at any point then the 5.1 file serves no purpose and we might as well just write all versions to the same yaml.

jacob-scott3141: If you bump the corda version, it will create a new yaml file and leave the old version untouched, so if you still had the 5.1 yaml files present on a 5.2 corda api, you could temporarily revert the `cordaProductVersion` to 5.1.0 and run an `apiDiff` to see all the breaking changes from 5.1 -> 5.2 

I guess one question would be do we want to keep the 5.1 yaml's sitting around or do we need to move them over manually if / when we do an "end of release" review. If we go with the former then I think having a `scans` directory in each subproject would be wise.

serialization/src/api/corda-serialization-5.1.0.yaml
driessamyn: It's a good question. Maybe @josephzunigadaly or @kyriathar may be able to confirm for sure.
On the other hand, I think it's ok to leave it as is, but maybe we need a ticket to move everything that doesn't need to be public API away from this repo instead?

josephzunigadaly: Early in Corda 5 development, we went through this module and moved anything that was not public API. What remains here was chosen to be public API at that time.

wzur-r3 (approved):
LGTM for Gradle configuration files

ronanbrowne (approved):
LGTM

PR #1275, opened 2023-10-04
CORE-17471 - Apply indexing to frequently used columns of the db table `utxo_transaction_output`

filipesoliveira:
The token selection feature uses db queries frequently. Indexing the columns that are used by those queries will improve performance. This commit indexes the columns `token_type`, `token_issuer_hash`, `token_symbol`, and `token_notary_x500_name` from table `utxo_transaction_output`.

DB query details for the token balance execution:
image

DB query details for the token selection execution:
image

### DIFF
--- a/data/db-schema/src/main/resources/net/corda/db/schema/vnode-vault/migration/ledger-utxo-creation-v5.1.xml
+++ b/data/db-schema/src/main/resources/net/corda/db/schema/vnode-vault/migration/ledger-utxo-creation-v5.1.xml
@@ -4,39 +4,33 @@
                    xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-4.3.xsd">
 
     <changeSet author="R3.Corda" id="ledger-utxo-creation-v5.1">
-        <createIndex
-                indexName="utxo_visible_transaction_state_idx_consumed"
-                tableName="utxo_visible_transaction_state">
+        <createIndex indexName="utxo_visible_transaction_state_idx_consumed"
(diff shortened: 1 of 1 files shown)


data/db-schema/src/main/resources/net/corda/db/schema/vnode-vault/migration/ledger-utxo-creation-v5.1.xml
```diff
@@ -4,17 +4,41 @@
...
-       <createIndex
-           indexName="utxo_visible_transaction_state_idx_consumed"
-           tableName="utxo_visible_transaction_state">
-           <column name="consumed"/>
-       </createIndex>
+        <createIndex
+                indexName="utxo_visible_transaction_state_idx_consumed"
+                tableName="utxo_visible_transaction_state">
+            <column name="consumed"/>
+        </createIndex>
```
```suggestion
        <createIndex indexName="utxo_visible_transaction_state_idx_consumed"
                     tableName="utxo_visible_transaction_state">
            <column name="consumed"/>
        </createIndex>
```

I can see why you've changed the formatting, but the above seems consistent with the Liquibase styling convention (and what is used in other C5 Liquibase files). I'd suggest applying this to all of the entries in this file.

vlajos:
https://github.com/corda/corda-api/pull/1270 has some related updates.
That one is renaming tables used by this/etc.

relyafi (commented):
One minor comment regarding style. I think we need to review how we are defining changesets generally, but that is a broader question and I don't think we should tackle that as part of this PR.

The PR however does need a description and some test notes, to provide background to the change. I'm assuming we should be able to demonstrate the new indexes being used by running the SQL queries that utilise these indexes with `EXPLAIN ANALYSE`.

relyafi (approved):
LGTM

PR #1276, opened 2023-10-04
CORE-17476 Adding Corda worker RPC endpoints to API schema

ben-millar:
In a related corda-runtime-os PR ([HERE](https://github.com/corda/corda-runtime-os/pull/4777)), we're adding a `ClusterIP` load balancer service to each Corda worker which will distribute HTTP calls to various instances of each worker type. The endpoints for these services are then passed into the `FlowWorker` startup args in key:value pairs. This PR adds these keys to the API so that we can query the endpoint map passed into the `FlowWorker`.

**PR Building with these changes:** https://ci02.dev.r3.com/job/Corda5/job/corda-runtime-os/job/PR-4823/8/

### DIFF
--- a/data/topic-schema/src/main/java/net/corda/schema/Schemas.java
+++ b/data/topic-schema/src/main/java/net/corda/schema/Schemas.java
@@ -287,9 +287,9 @@ public final class Schemas {
      */
     public static final class Endpoint {
         private Endpoint() {}
-        public static final String CRYPTO_WORKER_REST_ENDPOINT = "worker.rest.endpoint.crypto";
-        public static final String VERIFICATION_WORKER_REST_ENDPOINT = "worker.rest.endpoint.verification";
-        public static final String UNIQUENESS_WORKER_REST_ENDPOINT = "worker.rest.endpoint.uniqueness";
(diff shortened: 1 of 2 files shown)


data/topic-schema/src/main/java/net/corda/schema/Schemas.java
```diff
@@ -277,4 +277,15 @@ private ScheduledTask() {}
...
+
+    /**
+     * Corda worker RPC endpoints
+     */
+    public static final class Endpoint {
+        private Endpoint() {}
+        public static final String CRYPTO_WORKER_RPC_ENDPOINT = "worker.rpc.endpoint.crypto";
+        public static final String VERIFICATION_WORKER_RPC_ENDPOINT = "worker.rpc.endpoint.verification";
+        public static final String UNIQUENESS_WORKER_RPC_ENDPOINT = "worker.rpc.endpoint.uniqueness";
+        public static final String PERSISTENCE_WORKER_RPC_ENDPOINT = "worker.rpc.endpoint.persistence";
```
jujoramos: Is this _really_ RPC?, would it make sense to call it HTTP directly instead?.

jujoramos (dismissed):
LGTM

jujoramos (approved):
LGTM

PR #1286, opened 2023-10-10
CORE-17431 utxo_transaction_metadata table and related migrations

vlajos:
CORE-17431 Create utxo_transaction_metadata for storing metadata since it is relatively large but does not change frequently. Also, two new fields for accessing signed group parameters and CPI file checksum easier.

Tests:
* Running the 5.1 migration script on top of a 5.0 combined worker with some vnodes and transactions. After running the migration, the new table and fields appeared with the expected content and the metadata components were deleted from the components table.
* Running utxo smoke tests on the 5.1 combined worker executed the migrations on a clean database.

Runtime-os: https://github.com/corda/corda-runtime-os/pull/4845


blsemo (approved):
LGTM

PR #1288, opened 2023-10-10
CORE-17430 Modify the utxo_transaction_sources table by removing unused columns and renaming

nkovacsx:
### Overview

The `created` and `is_ref_input` columns are unnecessary to have in the `utxo_transaction_sources` table.
The created column will be the same as in utxo_transaction_component table and it’s a 1:1 mapping.
Whether a state is an input or a reference can be found out by using the groupIdx column rather than having an extra is_ref_input column.

Also rename the ref_transaction_id and ref_leaf_idx columns as those are quite confusing.

### Test plan
- Run the combined worker with Postgres (5.0) then build the VNode CLI plugin (5.1), run the migration that generates a SQL file. Run the SQL file in the 5.0 DB. Check that the fields are present and data was migrated properly.

- Run a combined worker with Postgres (5.1) and make sure no errors during startup / e2e tests. Check the DB that transactions are actually stored and the new fields are populated.


data/db-schema/src/main/resources/net/corda/db/schema/vnode-vault/migration/ledger-utxo-creation-v5.1.xml
```diff
@@ -113,47 +113,29 @@
...
+
+        <!-- Rename the columns in utxo_transaction_sources to avoid confusion -->
+        <renameColumn columnDataType="VARCHAR(160)"
+                      newColumnName="referenced_state_transaction_id"
+                      oldColumnName="ref_transaction_id"
+                      tableName="utxo_transaction_sources"/>
+
+        <renameColumn columnDataType="INT"
+                      newColumnName="referenced_state_index"
+                      oldColumnName="ref_leaf_idx"
```
```suggestion
                      newColumnName="source_state_transaction_id"
                      oldColumnName="ref_transaction_id"
                      tableName="utxo_transaction_sources"/>

        <renameColumn columnDataType="INT"
                      newColumnName="source_state_idx"
                      oldColumnName="ref_leaf_idx"
```

Feels better imo as it matches the use of "sources" in the table name. Also keep `idx` over `index` for consistency with other fields.

vlajos (dismissed):

relyafi (approved):

vlajos (approved):

blsemo (approved):
LGTM

PR #1289, opened 2023-10-10
CORE-17605 - Deprecated `TokenClaim.useAndRelease`

filipesoliveira:
Claimed tokens are now managed differently which makes the method irrelevant. From release 5.1, nothing will happen if the method is called.


vkolomeyko (approved):
LGTM

owenstanford (approved):
LGTM

PR #1290, opened 2023-10-10
CORE-16626 Replace string UUIDs with UUID objects

ac101m:
# Changes
- Change type of groupId field in `IntropIdentityInfo` from string to UUID object.
- Change type of groupId parameter of `callFacade()` in `FlowMessaging` from string to UUID object.
- Bump API version to 24.


dickon:
Isn't this a breaking change to the API? I don't see this aligns with https://r3-cev.atlassian.net/wiki/spaces/CB/pages/4601151665/Evolution+of+the+CorDapp+API

szymonsztuka:
> Isn't this a breaking change to the API? I don't see this aligns with https://r3-cev.atlassian.net/wiki/spaces/CB/pages/4601151665/Evolution+of+the+CorDapp+API

@dickon No API is broken as this is the feature branch never released.

szymonsztuka (approved):

dickon:
> @dickon No API is broken as this is the feature branch never released.
Ah yes, my bad, I'd missed the target of this was not a release branch, and you're right that this declaration never went into a frozen branch, please disregard my comment.

PR #1291, opened 2023-10-11
CORE-17623 - Change ACL mappings to add tokenSelection worker.

driessamyn:
Add `tokenSelection` to the topic ACL mappings. 

NOTE: the `db` user should be removed from these, but to avoid API version bumps, this will be done once they are not used anymore.

runtime changes: https://github.com/corda/corda-runtime-os/pull/4835


owenstanford (approved):
LGTM

PR #1292, opened 2023-10-11
CORE-15552 Add `findUnconsumedStatesByExactType` paging API

lankydan:
Add `findUnconsumedStatesByExactType` to `UtxoLedgerService `, allowing developers to retrieve states by type from the vault in pages. This dissuades developers from retrieving every state from the vault in one go and returning them to a flow.


ledger/ledger-utxo/src/main/java/net/corda/v5/ledger/utxo/UtxoLedgerService.java
```diff
@@ -115,6 +117,46 @@ public interface UtxoLedgerService {
     @Suspendable
     <T extends ContractState> List<StateAndRef<T>> findUnconsumedStatesByExactType(@NotNull Class<T> type);
 
+    /**
+     * Finds unconsumed states of the specified {@link ContractState} type in the vault.
```
relyafi:
```suggestion
     * Finds unconsumed states of the specified {@link ContractState} type in the vault. This version supports paging, limiting the number of results returned in a single query call by setting the `limit` argument.
```

I'd also update the doc for the existing call to recommend using this in preference if a (potentially unbounded) number of states may be returned, for performance reasons.

ledger/ledger-utxo/src/main/java/net/corda/v5/ledger/utxo/UtxoLedgerService.java
```diff
@@ -115,6 +117,46 @@ public interface UtxoLedgerService {
...
+     * @param limit The size of each page.
+     * @param createdTimestampLimit The timestamp limit the underlying query enforces.
+     * @return Returns a {@link ResultSet} of {@link StateAndRef} of unconsumed states of the specified type.
+     */
+    @NotNull
+    @Suspendable
+    <T extends ContractState> ResultSet<StateAndRef<T>> findUnconsumedStatesByExactType(
+            @NotNull Class<T> type,
+            @NotNull Integer limit,
+            @NotNull Instant createdTimestampLimit
```
relyafi, line 149: Strictly speaking, this is a change beyond what is claimed by this PR. If we want to add this, then the PR title / description should reflect this as well.

Do we actually have a use-case for this? If the intent is just to limit it to what exists at the start of the query, then this could just be done internally. Or we could actually just allow it to get the latest results until it runs out of results to return, and not set any limit.

relyafi (commented):
I think the PR needs a bit more background adding to the description, and a link to the runtime-os PR.

ledger/ledger-utxo/src/main/java/net/corda/v5/ledger/utxo/UtxoLedgerService.java
lankydan: It keeps its functionality in line with the vault named query functionality. I think it makes sense to restrict it in the same way we do there, especially since the underlying mechanism is the same between them. We can do it internally if we wanted to, but either way it needs setting whether publicly or internally.

lankydan:
A question in regards to this PR.

Should we remove the `findUnconsumedStatesByExactType` that does not support paging? The API was added in 5.1 (the current release) so it hasn't been released yet. Instinctively I think that is the right decision, especially since I considered deprecating it.

ledger/ledger-utxo/src/main/java/net/corda/v5/ledger/utxo/UtxoLedgerService.java
relyafi: Ok, then the question is whether we want to force users to have to provide it, or we provide another overload / a default argument that just sets it to now?

relyafi:
> A question in regards to this PR.
> 
> Should we remove the `findUnconsumedStatesByExactType` that does not support paging? The API was added in 5.1 (the current release) so it hasn't been released yet. Instinctively I think that is the right decision, especially since I considered deprecating it.

That sounds reasonable if it's not been released yet. We can also ask the question whether we should support paging on `findUnconsumedStatesByType`, but I think that is deprecated now anyway, in which case there is no point?

lankydan:
> > A question in regards to this PR.
> > Should we remove the `findUnconsumedStatesByExactType` that does not support paging? The API was added in 5.1 (the current release) so it hasn't been released yet. Instinctively I think that is the right decision, especially since I considered deprecating it.
> 
> That sounds reasonable if it's not been released yet. We can also ask the question whether we should support paging on `findUnconsumedStatesByType`, but I think that is deprecated now anyway, in which case there is no point?

Agreed, I think there is no point. In fact, I suggest we `@Deprecate` it if we haven't already.

vlajos (dismissed):

ledger/ledger-utxo/src/main/java/net/corda/v5/ledger/utxo/UtxoLedgerService.java
blsemo: I don't like the fact users have to provide a timestamp on this API - there should at least be an override that defaults to `now()`.

vlajos (dismissed):

vlajos (dismissed):

vlajos (dismissed):

vlajos (approved):

rick-r3 (approved):

PR #1295, opened 2023-10-12
CORE-16323: Add Index on State Version Column

jujoramos:
Add index on the version field to improve optimistic locking checks.

### DIFF
--- a/data/db-schema/src/main/resources/net/corda/db/schema/statemanager/migration/state-manager-migration-v5.1.xml
+++ b/data/db-schema/src/main/resources/net/corda/db/schema/statemanager/migration/state-manager-migration-v5.1.xml
@@ -30,6 +30,9 @@
                 <constraints nullable="false"/>
             </column>
         </createTable>
+        <createIndex tableName="state" indexName="state_version_idx">
+            <column name="version"/>
+        </createIndex>
         <addPrimaryKey columnNames="key" constraintName="state_key" tableName="state"/>
     </changeSet>
 </databaseChangeLog>


JamesHR3 (approved):

data/db-schema/src/main/resources/net/corda/db/schema/statemanager/migration/state-manager-migration-v5.1.xml
```diff
@@ -30,6 +30,9 @@
                 <constraints nullable="false"/>
             </column>
         </createTable>
+        <createIndex tableName="state" indexName="state_version_idx">
+            <column name="version"/>
+        </createIndex>
```
sprintengine, line 35: Won't the index need to include (key, version) - particularly for these queries?
```
DELETE FROM $STATE_MANAGER_TABLE s WHERE s.key = :$KEY_PARAMETER_NAME AND s.version = :$VERSION_PARAMETER_NAME

UPDATE $STATE_MANAGER_TABLE SET
            key = :$KEY_PARAMETER_NAME, value = :$VALUE_PARAMETER_NAME, version = version + 1, metadata = CAST(:$METADATA_PARAMETER_NAME as JSONB), modified_time = CURRENT_TIMESTAMP AT TIME ZONE 'UTC'
            WHERE key = :$KEY_PARAMETER_NAME AND version = :$VERSION_PARAMETER_NAME
```

jujoramos: Think your'e right... will update the changelog and open another `PR`.

PR #1296, opened 2023-10-12
CORE-16323: Improve State Manager Index

jujoramos:
Update recently introduced index so it uses both the "key" and
"version" columns as both are used by the State Manager during update
and delete queries.


conal-smith (approved):
lgtm

PR #1303, opened 2023-10-17
ES-1485 PR and Branch housekeeping automation (still dry-run)

jacob-scott3141:
- Adds a cron-job that will trigger this workflow at midnight, every Monday - Friday
- Still in dry-run mode
- Increased the limit of affected branches per execution from 10 -> 30


ronanbrowne (approved):
LGTM

PR #1304, opened 2023-10-17
CORE-17012 dlq topic to be used by multi source mediator for producer  serialization errors

LWogan:
It is not possible for a multi source mediator to determine which topic a record came from which triggered an output record that failed to serialize and therefore should be DLQ'd.

Adding a common topic for all mediators to send a DLQ record for to record serialization failures. 

Open to debate whether we should even bother DLQing this records. However as we DLQ these in the S+E this seems the only way to preserve that logic


LWogan:
On second thought, im going to close this. It doesnt make sense to me that we DLQ a producer record that fails to send. We already log it. That should be enough.

PR #1311, opened 2023-10-23
CORE-17941 Add backchain verifying property to notary info

nkovacsx:
This is a SPIKE Work on a SPIKE branch. 
Refer to https://r3-cev.atlassian.net/browse/CORE-17807


gradle.properties
```diff
@@ -9,7 +9,7 @@ cordaProductVersion = 5.1.0
 # NOTE: update this each time this module contains a breaking change
 ## NOTE: currently this is a top level revision, so all API versions will line up, but this could be moved to
 ##   a per module property in which case module versions can change independently.
-cordaApiRevision = 37
+cordaApiRevision = 38
```
blsemo: Can we either use a special Corda version, or an API revision way out of the range of the current 5.1 for this spike?
It will have to be updated to the then current 5.2 version anyway when/if merging to release.

PR #1313, opened 2023-10-23
CORE-17502 Added token selection worker HTTP endpoint definition.

owenstanford:
Changes to support HTTP RPC in Token Selection:
Adds the definition for the token selection workers HTTP RPC endpoint.

### DIFF
--- a/data/config-schema/src/main/java/net/corda/schema/configuration/BootConfig.java
+++ b/data/config-schema/src/main/java/net/corda/schema/configuration/BootConfig.java
@@ -47,4 +47,5 @@ public final class BootConfig {
     public static final String VERIFICATION_WORKER_REST_ENDPOINT = BOOT_WORKER_SERVICE + ".endpoints.verification";
     public static final String UNIQUENESS_WORKER_REST_ENDPOINT = BOOT_WORKER_SERVICE + ".endpoints.uniqueness";
     public static final String PERSISTENCE_WORKER_REST_ENDPOINT = BOOT_WORKER_SERVICE + ".endpoints.persistence";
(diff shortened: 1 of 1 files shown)


vkolomeyko (approved):
LGTM

data/config-schema/src/main/java/net/corda/schema/configuration/BootConfig.java
```diff
@@ -47,4 +47,5 @@ private BootConfig() {
     public static final String VERIFICATION_WORKER_REST_ENDPOINT = BOOT_WORKER_SERVICE + ".endpoints.verification";
     public static final String UNIQUENESS_WORKER_REST_ENDPOINT = BOOT_WORKER_SERVICE + ".endpoints.uniqueness";
     public static final String PERSISTENCE_WORKER_REST_ENDPOINT = BOOT_WORKER_SERVICE + ".endpoints.persistence";
+    public static final String TOKEN_SELECTION_WORKER_REST_ENDPOINT = BOOT_WORKER_SERVICE + ".endpoints.tokenSelection";
```
ben-millar, line 50:
```suggestion
    public static final String TOKEN_SELECTION_WORKER_REST_ENDPOINT = BOOT_WORKER_SERVICE + ".endpoints.token-selection";
```
I think helm will generate the service endpoint in kebab case https://github.com/corda/corda-runtime-os/blob/release/os/5.1/charts/corda-lib/templates/_helpers.tpl#L653C87-L653C114

ben-millar: Sorry, no, my mistake: this is for the value only. `tokenSelection` is fine here provided it matches what you specify in`workers.yaml`

ben-millar (approved):

PR #1327, opened 2023-11-03
CORE-18180 Correct `StateRef.toString`

lankydan:
The index in `StateRef.toString` was being formatted which added commas to the output. This created an invalid string representation of a `StateRef`, causing errors if we ever called `StateRef.parse` on the output of `StateRef.toString`.

Rely on normal string concatenation instead of `MessageFormat.format` to resolve this.

### DIFF
--- a/ledger/ledger-utxo/src/main/java/net/corda/v5/ledger/utxo/StateRef.java
+++ b/ledger/ledger-utxo/src/main/java/net/corda/v5/ledger/utxo/StateRef.java
@@ -6,7 +6,6 @@ import net.corda.v5.crypto.SecureHash;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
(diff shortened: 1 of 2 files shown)


ledger/ledger-utxo/src/main/java/net/corda/v5/ledger/utxo/StateRef.java
```diff
@@ -126,6 +126,6 @@ public int hashCode() {
      */
     @Override
     public String toString() {
-        return MessageFormat.format("{0}{1}{2}", transactionId, DELIMITER, index);
+        return transactionId + ":" + index;
```
vkolomeyko:
```suggestion
        return transactionId + DELIMITER + index;
```

jennyang-r3 (approved):
LGTM

aditisdesai (approved):
LGTM 👍

PR #1353, opened 2023-11-21
CORE-17821 Removing unused crypto topics

ben-millar:
Now that communication with the `CryptoWorker` is done via RPC rather than over Kafka, we no longer need to define the relevant topics in our API. This PR removes those which are no longer used.

Associated `corda-runtime-os` changes: https://github.com/corda/corda-runtime-os/pull/5150/


anien:
Do you need version bump? Is there a corresponding runtime-os change?

ben-millar:
> Do you need version bump? Is there a corresponding runtime-os change?

@anien My PR gate seemed to be failing on the "execute dependent jobs" step without incrementing the API version. runtime-os changes are here: https://github.com/corda/corda-runtime-os/pull/5150/

dickon (approved):
LGTM

PR #1354, opened 2023-11-21
CORE-12165: merge in 5.2 changes and bump version

dickon:


anien (approved):
Approving on the fact this is simple merge plus a version bump.

PR #1356, opened 2023-11-21
CORE-18168: Add Kafka payload and config to support changing user password

Tom-Fitzpatrick:
Creates
- ChangeUserPasswordOtherRequest.avsc
- ChangeUserPasswordSelfRequest.avsc

These will be used to allow users to change their own password, and for an admin to change a particular user's password.

This also adds new config settings related to password expiry, specifically, default values for how long a password should last before expiring depending on whether a user or an admin changed the password.


data/config-schema/src/main/resources/net/corda/schema/configuration/rbac.1.0/corda.rbac.json
```diff
@@ -0,0 +1,36 @@
...
+        "maximum": 30,
+        "default": 7
+      },
+      "passwordExpiryWarningWindow": {
+        "description": "The time (days) before a password expires in which we begin to offer warnings about upcoming expiry.",
+        "type": "integer",
+        "minimum": 1,
+        "default": 30,
+        "maximum": 365
+      }
```
driessamyn: We might need some input from @jamesbr3 with regards to best practice, but I thought that these days the best practice was not to enforce password expiry but instead focus on complexity (which we may want to make a rule that can be changed by configuration).

Do we even need an expiry policy? Shouldn't Corda be un-opinionated about this and instead offer the ability to report on password age so that integrators can integrate with their own policy.

Either way, I think don't think forcing a maximum is correct as this would prevent a customer setting their own policy.

vkolomeyko: Yes, I agree we can remove `maximum` value from all the parameters. This would allow setting any big value, effectively disabling password expiry, if this is what cluster administrator wants.

In our current RBAC data model, we do have provisions for password expiry and also we do have customer requests for it.

Tom-Fitzpatrick: Thanks guys, I removed the maximum values from these configs.

vkolomeyko (approved):
LGTM

PR #1357, opened 2023-11-21
CORE-18495 Made topic field in ExternalEvent.avsc optional; defaults to null

ben-millar:
Now that some external events are send over HTTP RPC rather than the Kafka message bus, we should make the topic field in the ExternalEvent AVRO object optional so it need not be set for HTTP-based communications.

Runtime OS PR with these changes building successfully: https://github.com/corda/corda-runtime-os/pull/5152 🟢


mbrkic-r3 (approved):
LGTM

PR #1358, opened 2023-11-21
CORE-12165: resolve conflict on API version

dickon:


anien (approved):
We hope this fixed github thinking we are bringing more commits into 5.2 than we actually are.

PR #1359, opened 2023-11-22
CORE-16125: add Avro messages and topics for crypto key rotation

anien:
Add initial versions of the Avro messages for crypto key rotation, and Kafka topics for them.


driessamyn (approved):
LGTM

PR #1362, opened 2023-11-23
CORE-18492 notary registration & add backchain required flag on Notar…

jennyang-r3:
This PR includes changes such as:
- new field schema for notary service registration that allows skip backchain or not.
- added a `backchainRequired` flag to `NotaryInfo` to be able to use in the implementation of a contract verifying notary which will be a follow-up ticket of this.

### DIFF
--- a/membership/src/main/java/net/corda/v5/membership/NotaryInfo.java
+++ b/membership/src/main/java/net/corda/v5/membership/NotaryInfo.java
@@ -18,12 +18,14 @@ import java.util.Collection;
  * String protocol = notaryInfo.getProtocol();
  * Collection<Integer> protocolVersions = notaryInfo.getProtocolVersions();
  * PublicKey publicKey = notaryInfo.getPublicKey();
+ * Boolean backchainRequired = notaryInfo.getBackchainRequired();
  * }</pre></li>
  * <li>Kotlin:<pre>{@code
  * val name = notaryInfo.name
  * val protocol = notaryInfo.protocol
  * val protocolVersions = notaryInfo.protocolVersions
  * val publicKey = notaryInfo.publicKey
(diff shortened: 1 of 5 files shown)


membership/src/main/java/net/corda/v5/membership/NotaryInfo.java
```diff
@@ -48,4 +50,9 @@ public interface NotaryInfo {
      * @return The public key of the notary service, which will be a composite key of all notary virtual nodes keys.
      */
     @NotNull PublicKey getPublicKey();
+
+    /**
+     * @return boolean of whether it requires backchain verification.
+     */
+    @NotNull Boolean getBackchainRequired();
```
lankydan: Should this be `isBackchainRequired` to stick with normal java naming conventions for booleans?

membership/src/main/java/net/corda/v5/membership/NotaryInfo.java
```diff
@@ -48,4 +50,9 @@ public interface NotaryInfo {
      * @return The public key of the notary service, which will be a composite key of all notary virtual nodes keys.
      */
     @NotNull PublicKey getPublicKey();
+
+    /**
+     * @return boolean of whether it requires backchain verification.
+     */
+    @NotNull Boolean getIsBackchainRequired();
```
blsemo: I understood @lankydan 's comment as whether this method should be `@NotNull Boolean isBackchainRequired();` - doesn't turn into a property in kotlin, but would be the idiomatic Java way of doing it.

lankydan: ➕

lankydan (approved):

blsemo (approved):
LGTM

PR #1366, opened 2023-11-23
CORE-18168: Insert new rbac config section with config-creation

Tom-Fitzpatrick:
This inserts the new rbac config section with a new config-creation-v5.2.xml in db-schema migration.

This also creates a new ConfigKey for RBAC_CONFIG

There is an accompanying change to update the support bundle scripts here: https://github.com/corda/corda-runtime-os/pull/5178


vkolomeyko (approved):
LGTM

PR #1371, opened 2023-11-30
CORE-17818: Api change to remove uniqueness checker duplication

emilybowe:
This change is required as part of this PR https://github.com/corda/corda-runtime-os/pull/5184 to remove uniqueness checker duplication. This change will remove an unused uniqueness checker topic from the api.

### DIFF
--- a/data/topic-schema/src/main/java/net/corda/schema/Schemas.java
+++ b/data/topic-schema/src/main/java/net/corda/schema/Schemas.java
@@ -225,16 +225,6 @@ public final class Schemas {
     }
 
     /**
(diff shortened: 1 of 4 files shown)


mbrkic-r3:
Should we remove [this](https://github.com/corda/corda-api/blob/emily/core-17818/api-change-to-remove-uniqueness-checker-duplication/data/topic-schema/src/main/resources/net/corda/schema/Flow.yaml#L12) as well?

mbrkic-r3 (approved):
LGTM

PR #1372, opened 2023-11-30
CORE-17816 contract verifying notary APIs

jennyang-r3:


blsemo (dismissed):
LGTM

blsemo (dismissed):
LGTM

ledger/ledger-utxo/src/main/java/net/corda/v5/ledger/utxo/UtxoLedgerService.java
```diff
@@ -37,6 +37,14 @@ public interface UtxoLedgerService {
     @Suspendable
     UtxoTransactionBuilder createTransactionBuilder();
 
+    /**
+     * Verifies {@link UtxoLedgerTransaction} in the notary server when using contract-verifying-notary.
```
nkovacsx: I think this is a bit too specific for an API doc. We shouldn't really say where this will be called, it might be used in other places as well in the future and then this documentation will not make sense then.

jennyang-r3: That's true but I think we still should list what notary can be called and what not to. If this is going to be public API.

ledger/ledger-utxo/src/main/java/net/corda/v5/ledger/utxo/NotarySignatureVerificationService.java
```diff
@@ -0,0 +1,43 @@
...
+import java.util.List;
+import java.util.Map;
+
+@DoNotImplement
+public interface NotarySignatureVerificationService {
+
+    /**
+     * Verifies given transaction by id is signed by the notary signatures.
+     *
+     * @param transactionId of signatures
```
blsemo: That line does not make a lot of sense - do you mean something like
```suggestion
     * @param transactionId that is exepcted to be signed over
```

ledger/ledger-utxo/src/main/java/net/corda/v5/ledger/utxo/UtxoLedgerService.java
blsemo: I agree with Nandor here - we're using it for the notary implementation here, but this can (and probably will) have other use cases, so this comment it not quite right.

ledger/ledger-utxo/src/main/java/net/corda/v5/ledger/utxo/NotarySignatureVerificationService.java
```diff
@@ -0,0 +1,43 @@
...
+    );
+
+    /**
+     * Get notary public key by key id of signature.
+     *
+     * @param keyId of notary signature
+     * @param notaryKey that is expected to be signed with
+     * @param keyIdToNotaryKeys a map of keyId to notaryKeys
+     */
+    PublicKey getNotaryPublicKeyByKeyId(
```
blsemo: Do we need this function on the public API?

blsemo (dismissed):
LGTM

ledger/ledger-utxo/src/main/java/net/corda/v5/ledger/utxo/transaction/filtered/UtxoFilteredTransactionBuilder.java
```diff
@@ -178,6 +179,19 @@ public interface UtxoFilteredTransactionBuilder {
     @Suspendable
     UtxoFilteredTransactionBuilder withOutputStates(@NotNull Predicate<ContractState> predicate);
 
+    /**
+     * Includes an audit proof of the output state refs component group from a {@link UtxoSignedTransaction} in the
+     * current {@link UtxoFilteredTransaction}.
+     *
+     * @param indexes index of transaction to include its output to {@link UtxoFilteredTransaction}
```
vlajos: Is this correct like this?

jennyang-r3: could you elaborate on your question?

vlajos: Is it not `indexes of the output states to be included` or something like that?
(But I might have misunderstood something.)

jennyang-r3: Ah, yeah it should be plural

blsemo (approved):
LGTM

PR #1375, opened 2023-12-01
CORE-18813 - enable Dependabot

driessamyn:
- Use gradle catalogue using a versions toml file to define library versions.
- Enable Dependabot so it can raise PRs to update dependencies

image

E.g. https://github.com/corda/corda-api/pull/1383/files

NOTE: this doesn't complete the work as there are some gaps, but this evolved from the POC and may be a good first step.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -32,13 +32,6 @@ org.gradle.jvmargs=-Dfile.encoding=UTF-8 -XX:MaxMetaspaceSize=2g
 internalPluginVersion = 1.+
 artifactoryPluginVersion = 4.28.2
 
-# Testing
-junitVersion = 5.10.0
-
-# OSGi
-bndVersion = 6.4.0
-bndlibVersion = 6.4.1
-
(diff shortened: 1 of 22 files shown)


gradle.properties
```diff
@@ -30,48 +30,19 @@ publicArtifactURL = https://download.corda.net/maven
...
-commonsCompressVersion = 1.24.0
-bouncycastleVersion = 1.73
-grgitPluginVersion = 5.2.0
-taskTreePluginVersion = 2.1.1
-javaxPersistenceApiVersion = 2.2
-jacksonVersion = 2.15.2
 
 # Testing
-assertjVersion = 3.24.+
 junitVersion = 5.10.0
```
driessamyn: Some versions still left in here because the version catalog doesn't seem to work for the cordapp configuration plugin. This may need some looking into.

ronanbrowne (dismissed):
LGTM

ronanbrowne (approved):

PR #1390, opened 2023-12-07
CORE-18271: add LeveledHash type and method

dickon:


blsemo:
Do we actually need this on the public API?

PR #1394, opened 2023-12-12
CORE-18320 Support for a flow specific timeout on Flow Message API

mbrkic-r3:
Support for a flow specific timeout on Flow Message API. Client can specify a flow session specific timeout on Flow Message API. If set, it will be passed to the initiated party to also use it. If the client does not provide a timeout value then the value from Corda Configuration is used.

A new class `FlowSessionConfiguration` is introduced to store session configuration. Methods that used configuration parameter `requireClose` will be deprecated with another PR.

Avro schema for `SessionState` was modified, field `requireClose` was removed as this value will be stored in `sessionProperties`.


data/avro-schema/src/main/resources/avro/net/corda/data/flow/state/session/SessionState.avsc
```diff
@@ -34,6 +34,15 @@
       "type": "boolean",
       "doc": "True if the user has set requireClose to be true when calling initiate flow. False otherwise."
     },
+    {
+      "name": "sessionTimeout",
```
LWogan: I don't think you need to add a new avro field. You can just set it in the sessionProperties which are already exchanged on SessionEvents and then stored in the SessionState

LWogan (commented):
The method overloading is not ideal like you already said. Could we potentially add a field flowSessionConfig/sessionProperties Map<String, String> or something like that to the initiateFlow API? I think it's too late for requireClose but there may be more things we want to configure per session in the future.

LWogan:
The compatibility scan failed due to the API change. Is this just a warning or something that needs to be addressed?

mbrkic-r3:
> The compatibility scan failed due to the API change. Is this just a warning or something that needs to be addressed?

https://r3holdco.slack.com/archives/C039J1A3ZQB/p1702390578322899

> Looks like an issue with the API Scanner, for now - since the change doesn't look like it is breaking - I would run ./gradlew cementApi and push the changes

data/avro-schema/src/main/resources/avro/net/corda/data/flow/state/session/SessionState.avsc
mbrkic-r3: I was following changes related to `requireClose` as example. Is there some reason that filed was created as new avro field instead of setting it in sessionProperties?

mbrkic-r3:
> The method overloading is not ideal like you already said. Could we potentially add a field flowSessionConfig/sessionProperties Map<String, String> or something like that to the initiateFlow API? I think it's too late for requireClose but there may be more things we want to configure per session in the future.

Yes, we might add that to make it future-proof, in case we'll need to add more properties. It might be better to use `SessionConfigBuilder` as that would enforce setting properties that actually exist and also type safety. We can add `requireClose` there and deprecate methods that use it as separate parameter.

LWogan (dismissed):
LGTM

application/src/main/java/net/corda/v5/application/messaging/FlowMessaging.java
```diff
@@ -106,6 +106,25 @@ public interface FlowMessaging {
     @NotNull
     FlowSession initiateFlow(@NotNull MemberX500Name x500Name, boolean requireClose);
```
LWogan, line 107: I think based on the slack thread, this can be marked @Deprecated as well the other one with requireClose in it

mbrkic-r3: Marked as deprecated

application/src/main/java/net/corda/v5/application/messaging/FlowSessionConfiguration.java
```diff
@@ -0,0 +1,62 @@
+package net.corda.v5.application.messaging;
+
+import java.time.Duration;
+
+/**
+ * Flow session configuration. Instances should be created using {@link Builder}.
+ */
+public class FlowSessionConfiguration {
```
simon-johnson-r3: Should this and the builder be `final`?

mbrkic-r3: This class will be container for session configuration properties, so when we'll need another property (similar to `requireClose` or `timeout`) in the future we'll add fields to this class. Since we don't plan to extend it, I'll change classes to be final.

simon-johnson-r3 (dismissed):

thiagoviana (approved):

PR #1395, opened 2023-12-12
CORE-18550: add payload reply from gateway

yift-r3:
See runtime pull request in https://github.com/corda/corda-runtime-os/pull/5250


data/avro-schema/src/main/resources/avro/net/corda/data/p2p/gateway/GatewayResponse.avsc
```diff
@@ -8,6 +8,13 @@
       "name": "id",
       "type": "string",
       "doc": "The id of the request corresponding to this response."
+    },
+    {
+      "name": "payload",
+      "type": [
+        "null",
+        "net.corda.data.p2p.crypto.AuthenticatedDataMessage"
```
dimosr, line 17: Should we also have `net.corda.data.p2p.crypto.AuthenticatedEncryptedDataMessage` here?

dimosr (approved):

PR #1400, opened 2023-12-13
CORE-18635 Avro schemas for session encryption operations

YashNabar:
APIs for an implementation of encryption and decryption.

Runtime pull request in https://github.com/corda/corda-runtime-os/pull/5258

### DIFF
--- a/data/avro-schema/src/main/resources/avro/net/corda/data/crypto/wire/ops/encryption/request/DecryptRpcCommand.avsc
+++ b/data/avro-schema/src/main/resources/avro/net/corda/data/crypto/wire/ops/encryption/request/DecryptRpcCommand.avsc
@@ -1,0 +1,28 @@
+{
+  "type": "record",
+  "name": "DecryptRpcCommand",
+  "namespace": "net.corda.data.crypto.wire.ops.encryption.request",
+  "doc": "Request to decrypt the given byte array",
+  "fields": [
+    {
+      "name": "category",
+      "type": "string",
+      "doc": "The category of HSM e.g. ENCRYPTION_SECRET, TLS, etc."
+    },
+    {
+      "name": "alias",
+      "doc": "The symmetric key alias.",
+      "type": ["null", "string"]
+    },
+    {
+      "name": "cipherBytes",
+      "type": "bytes",
+      "doc": "The data to decrypt."
+    },
+    {
+      "name": "context",
+      "type": "net.corda.data.KeyValuePairList",
+      "doc": "Optional key/value operation specific context"
+    }
+  ]
+}
--- a/data/avro-schema/src/main/resources/avro/net/corda/data/crypto/wire/ops/encryption/request/DecryptRpcCommand.avsc
+++ b/data/avro-schema/src/main/resources/avro/net/corda/data/crypto/wire/ops/encryption/request/DecryptRpcCommand.avsc
@@ -12,7 +12,7 @@
     {
       "name": "alias",
       "doc": "The symmetric key alias.",
-      "type": ["null", "string"]
+      "type": "string"
     },
     {
       "name": "cipherBytes",
--- a/data/avro-schema/src/main/resources/avro/net/corda/data/crypto/wire/ops/encryption/request/DecryptRpcCommand.avsc
+++ b/data/avro-schema/src/main/resources/avro/net/corda/data/crypto/wire/ops/encryption/request/DecryptRpcCommand.avsc
@@ -18,11 +18,6 @@
       "name": "cipherBytes",
       "type": "bytes",
       "doc": "The data to decrypt."
-    },
-    {
-      "name": "context",
-      "type": "net.corda.data.KeyValuePairList",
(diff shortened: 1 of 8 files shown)


data/avro-schema/src/main/resources/avro/net/corda/data/crypto/wire/ops/encryption/response/EncryptionOpsError.avsc
```diff
@@ -0,0 +1,12 @@
+{
+  "type": "record",
+  "name": "EncryptionOpsError",
+  "doc": "Error result while performing encryption or decryption operation.",
+  "namespace": "net.corda.data.crypto.wire.ops.encryption.response",
```
anien:
```suggestion
  "namespace": "net.corda.data.crypto.wire.ops.encryption.response",
  "doc": "Error result while performing encryption or decryption operation.",
```

data/avro-schema/src/main/resources/avro/net/corda/data/crypto/wire/ops/encryption/request/EncryptRpcCommand.avsc
```diff
@@ -0,0 +1,28 @@
...
+  "fields": [
+    {
+      "name": "category",
+      "type": "string",
+      "doc": "The category of HSM e.g. ENCRYPTION_SECRET, TLS, etc."
+    },
+    {
+      "name": "alias",
+      "doc": "The symmetric key alias.",
+      "type": ["null", "string"]
```
anien:
```suggestion
      "type": ["null", "string"]
      "doc": "The symmetric key alias.",
```

data/avro-schema/src/main/resources/avro/net/corda/data/crypto/wire/ops/encryption/request/DecryptRpcCommand.avsc
```diff
@@ -0,0 +1,28 @@
...
+  "fields": [
+    {
+      "name": "category",
+      "type": "string",
+      "doc": "The category of HSM e.g. ENCRYPTION_SECRET, TLS, etc."
+    },
+    {
+      "name": "alias",
+      "doc": "The symmetric key alias.",
+      "type": ["null", "string"]
```
anien:
```suggestion
      "type": ["null", "string"]
      "doc": "The symmetric key alias.",
```

anien (commented):
Just a minor suggestion, as I found it easier to read if the order is consistent.

dimosr (dismissed):

anien (dismissed):
LGTM.

data/avro-schema/src/main/resources/avro/net/corda/data/crypto/wire/ops/encryption/request/DecryptRpcCommand.avsc
```diff
@@ -0,0 +1,28 @@
...
+      "type": ["null", "string"],
+      "doc": "The symmetric key alias."
+    },
+    {
+      "name": "cipherBytes",
+      "type": "bytes",
+      "doc": "The data to decrypt."
+    },
+    {
+      "name": "context",
```
kyriathar: I think adding this `context` was was decided at design level. 

However, I have seen before a case where we have been using it while we shouldn't and we had to go in there and revert all of that, to use some other field instead back then (IRRC we used to pass the signature spec in context. This piece of work can be found [here](https://r3-cev.atlassian.net/browse/CORE-11521)). 

I would say if you are not planning on using it remove it as it will cause at the very least confusion on the API, like what is good for. It can always be added later if found it's needed which I doubt It will be needed.

data/avro-schema/src/main/resources/avro/net/corda/data/crypto/wire/ops/encryption/request/DecryptRpcCommand.avsc
```diff
@@ -0,0 +1,28 @@
+{
+  "type": "record",
+  "name": "DecryptRpcCommand",
+  "namespace": "net.corda.data.crypto.wire.ops.encryption.request",
+  "doc": "Request to decrypt the given byte array",
+  "fields": [
+    {
+      "name": "category",
```
kyriathar, line 8: nit: so this means we could be passing in any category. So for example `LEDGER` (ledger keys are meant to be used for signing) to decrypt secrets. Should we be restricting this field with an enum of valid categories?

data/avro-schema/src/main/resources/avro/net/corda/data/crypto/wire/ops/encryption/request/DecryptRpcCommand.avsc
yift-r3: I prefer to remove it. 

@dickon WDYT?

kyriathar (approved):
OK (crypto)

dimosr (approved):

PR #1402, opened 2023-12-13
CORE-18913 adjust the corda api configs for the mediator

LWogan:
Move flow config value to subscription config as it is a mediator property not a flow one.
Add a new configuration for minimum pool group size.

https://github.com/corda/corda-runtime-os/pull/5257


ben-millar (approved):

mbrkic-r3 (approved):
LGTM

PR #1403, opened 2023-12-13
CORE-18961: Up flow sandbox cache size to 20

vkolomeyko:
Having a current limit of 10 may too low to allow E2E tests to pass given that we have a number of vNodes running flows concurrently during E2E tests execution. With current limit of 10, sandbox context for the currently running flow may be closed, resulting bundle de-activation and `ClassNotFoundException` on the initiated flow execution.

Tried with: https://github.com/corda/corda-runtime-os/pull/5260

Proven to have a positive effect on E2E run: https://ci02.dev.r3.com/job/Corda5/job/corda5-e2e-tests/job/PR-369/13/

### DIFF
--- a/data/config-schema/src/main/resources/net/corda/schema/configuration/sandbox/1.0/corda.sandbox.json
+++ b/data/config-schema/src/main/resources/net/corda/schema/configuration/sandbox/1.0/corda.sandbox.json
@@ -22,7 +22,7 @@
               "description": "The maximum number of cached flow sandboxes.",
               "type": "integer",
               "minimum": 0,
-              "default": 10
+              "default": 20
             }
           }
         }


data/config-schema/src/main/resources/net/corda/schema/configuration/sandbox/1.0/corda.sandbox.json
```diff
@@ -22,7 +22,7 @@
               "description": "The maximum number of cached flow sandboxes.",
               "type": "integer",
               "minimum": 0,
-              "default": 10
+              "default": 20
```
driessamyn, line 25: I think we need to be careful with this.
While I'm sure increasing this is a good idea (and maybe it should be much much higher (100+) in a production system, increasing this to get our tests to pass is masking a problem rather than fixing it.
The same issues will be hit at a higher load.

CC @simon-johnson-r3

vkolomeyko: Yes, I agree. The underlying problem is that we do not handle flow execution very well when sandbox cache eviction happens. This PR was just an attempt to improve the current situation with E2E tests.
An alternative could be to make E2E tests run less in parallel by using `@Isolated` or `SAME_THREAD` annotations.

vkolomeyko:
Closing without merging.
E2E tests stability will be addressed by running less tests in parallel at the same time.

PR #1417, opened 2023-12-19
CORE-16870 Support nullable parameters in queries

nkovacsx:
### Overview

Right now nullable parameters are not supported in named queries, after this change we'll support setting null parameters. This feature was requested by various teams before.


blsemo (dismissed):
LGTM

simon-johnson-r3 (approved):

PR #1424, opened 2023-12-21
CORE-18553: Add response type from link manager to the gateway

yift-r3:
Introduce a type to wrap the responses from the link manager to the gateway.

Runtime pull request is in https://github.com/corda/corda-runtime-os/pull/5322

### DIFF
--- a/data/avro-schema/src/main/resources/avro/net/corda/data/p2p/gateway/LinkManagerResponse.avsc
+++ b/data/avro-schema/src/main/resources/avro/net/corda/data/p2p/gateway/LinkManagerResponse.avsc
@@ -1,0 +1,16 @@
+{
+  "type": "record",
+  "name": "LinkManagerResponse",
+  "namespace": "net.corda.data.p2p.gateway",
+  "doc": "A response from the p2p link manager to the p2p gateway component",
+  "fields": [
+    {
+      "name": "payload",
+      "type": [
+        "null",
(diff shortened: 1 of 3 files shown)


data/avro-schema/src/main/resources/avro/net/corda/data/p2p/gateway/LinkManagerResponse.avsc
dimosr, line 1: given this will be effectively part of the interface/contract of the link manager's endpoints, I think it's better to have it under a folder call `linkmanager` (or smth along those lines), instead of `gateway`. Or alternatively under the general folder `p2p`, although that has started becoming quite big now, so if we can start separating things a bit, I think it will help.

data/avro-schema/src/main/resources/avro/net/corda/data/p2p/gateway/LinkManagerResponse.avsc
```diff
@@ -0,0 +1,16 @@
+{
+  "type": "record",
+  "name": "LinkManagerResponse",
+  "namespace": "net.corda.data.p2p.gateway",
+  "doc": "A response from the p2p link manager to the p2p gateway component",
```
dimosr: similar spirit, might be worth making it a bit more generic (even though it's only used by the gateway currently) e.g.
```
"doc": "A response from the p2p link manager when being sent a LinkInMessage via the HTTP endpoint",
```

dimosr (approved):

PR #1429, opened 2024-01-02
CORE-18533: Remove Checkpoint Retry

jujoramos:
- Remove "RetryState" avro schema.
- Move "FlowStates" enumeration to its own avro schema definition file.
- Remove "RETRYING" and set "FAILED" as the default value (for schema
  evolution, unmatched values do not throw an error but are resolved 
  to the default instead).
- Remove "retryState" and "maxFlowSleepDuration" fields from the
  "PipelineState" avro schema.
- Remove "CheckpointSchemaCompatibilityTest" (we only support evolving
  types one minor version at a time).
- Add schema compatibility tests for "FlowStates" and "PipelineState".

Runtime Change: https://github.com/corda/corda-runtime-os/pull/5374


simon-johnson-r3 (dismissed):
LGTM

data/config-schema/src/main/resources/net/corda/schema/configuration/flow/1.0/corda.flow.json
```diff
@@ -33,7 +33,7 @@
           "default": 16000
         },
         "maxFlowSleepDuration": {
-          "description": "The maximum delay in milliseconds before Corda schedules a periodic wake-up.",
+          "description": "Deprecated. The maximum delay in milliseconds before Corda schedules a periodic wake-up.",
```
LWogan, line 36: Is this the official way we are to deprecate configs? Just curious as i know some other ones which need deprecating.

LWogan (approved):
LGTM

data/config-schema/src/main/resources/net/corda/schema/configuration/flow/1.0/corda.flow.json
jujoramos: I'm not 100% sure to be honest... @simon-johnson-r3, thoughts?

thiagoviana (approved):

PR #1430, opened 2024-01-02
CORE-18532: Explicit Crypto Retry Settings

jujoramos:
Instead of setting the maximum amount of retries within the code and
ignore the values configured by the user, set the minimum and maximum
allowed values in the configuration schema directly.


kyriathar (approved):

dickon (approved):
LGTM. Documents what we already enforce rather than silently ignoring extra retries.

PR #1432, opened 2024-01-03
CORE-18984 additional mediator config for configuring poll timeout

LWogan:
The value for the mediator poll timeout is currently hard coded. This pr adds a config to make it configurable so it can be configured separately to other patterns. 

https://github.com/corda/corda-runtime-os/pull/5360


mbrkic-r3 (dismissed):
LGTM

LWogan:
Updated minimum validation to be below the default

mbrkic-r3 (approved):
Re-approved

PR #1433, opened 2024-01-03
CORE-2840: Remove no longer needed package exports

vkolomeyko:
The packages are no longer necessary and should not be exported.
This is a non-breaking change.


yift-r3 (approved):

ronanbrowne (approved):
LGTM

PR #1439, opened 2024-01-08
API changes for merkle proof poc

nkovacsx:


ledger/ledger-utxo/src/main/java/net/corda/v5/ledger/utxo/transaction/filtered/UtxoFilteredTransaction.java
```diff
@@ -124,4 +126,7 @@ public interface UtxoFilteredTransaction extends TransactionWithMetadata {
      * @throws CordaRuntimeException if the current {@link UtxoFilteredTransaction} fails to verify correctly.
      */
     void verify();
+
+    @NotNull
+    Map<Integer, MerkleProof> getMerkleProofs();
```
dickon, line 131: Would it be better to make this an operation which takes a set of integers as an argument and returns exactly the `MerkleProof` for each requested integer? If callers ever don't need certain component groups we could be wasting a noticeable effort in many cases.

data/avro-schema/src/main/resources/avro/net/corda/data/ledger/persistence/PersistMerkleProofIfDoesNotExist.avsc
dickon, line 1: I think `PersistMerkelProof` is cleaner, the conflict case can be covered in the doc field or left to the implementation.

PR #1440, opened 2024-01-08
CORE-18507: Add UnamangedKeyStatus that will serve both key status and key rotation

anien:
* Adds avro schema that serves both key status and key rotation status requests. 
* This schema is stored in the db.

### DIFF
--- a/data/avro-schema/src/main/resources/avro/net/corda/data/crypto/wire/ops/key.status/UnmanagedKeyStatus.avsc
+++ b/data/avro-schema/src/main/resources/avro/net/corda/data/crypto/wire/ops/key.status/UnmanagedKeyStatus.avsc
@@ -2,7 +2,7 @@
   "type": "record",
   "name": "UnmanagedKeyStatus",
   "namespace": "net.corda.data.crypto.wire.ops.key.status",
-  "doc": "Total number of keys for tenantId that needs rotating away from rootKeyAlias",
+  "doc": "Defines the key status and key rotation status data stored in the database.",
   "fields": [
     {
       "name": "rootKeyAlias",
@@ -22,7 +22,7 @@
     {
       "name": "rotatedKeys",
       "type": ["null", "int"],
-      "doc": "Number of keys for tenantId that has been rotated."
+      "doc": "Number of keys for tenantId that has been rotated. Null in case of key status."
     }
   ]
 }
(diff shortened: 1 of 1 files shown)


data/avro-schema/src/main/resources/avro/net/corda/data/crypto/wire/ops/key.status/UnmanagedKeyStatus.avsc
```diff
@@ -0,0 +1,28 @@
...
+  "fields": [
+    {
+      "name": "rootKeyAlias",
+      "type": "string",
+      "doc": "Alias of an unmanaged key that we are rotating away from."
+    },
+    {
+      "name": "tenantId",
+      "type": "string",
+      "doc": "GUID of the tenantId or clusterDB."
```
simon-johnson-r3: Is this really a GUID? I thought it was either the short hash of the vnode or a string from a list defined in Corda? Also do we need this if it's the key?

data/avro-schema/src/main/resources/avro/net/corda/data/crypto/wire/ops/key.status/UnmanagedKeyStatus.avsc
```diff
@@ -0,0 +1,28 @@
+{
+  "type": "record",
+  "name": "UnmanagedKeyStatus",
+  "namespace": "net.corda.data.crypto.wire.ops.key.status",
+  "doc": "Defines the key status and key rotation status data stored in the database.",
```
simon-johnson-r3: I wouldn't use the word database here as there are future plans to sit the statemanager on other things.

data/avro-schema/src/main/resources/avro/net/corda/data/crypto/wire/ops/key.status/UnmanagedKeyStatus.avsc
anien: You are right. It's not a GUID. 
I put it here to have the relevant data in one place, but you are right, `tenantId` will be a key, so we don't necessarily need it here. I will remove it and we can always put it back if we find the need for it.

data/avro-schema/src/main/resources/avro/net/corda/data/crypto/wire/ops/key.status/UnmanagedKeyStatus.avsc
anien: Good point. Removed.

simon-johnson-r3 (approved):

PR #1441, opened 2024-01-09
CORE-18691 Store Merkle proofs and fetch them (DB/Avro changes)

nkovacsx:
### Overview

We now want to store Merkle proofs in the database as a separate table. 
Added a 5.2 ledger changelog which creates this new table.

Also created requests for both persisting and fetching the merkle proofs from these tables.

`UtxoFilteredTransaction` API had to be changed so we can retrieve the Merkle proofs for each component group.


ledger/ledger-utxo/src/main/java/net/corda/v5/ledger/utxo/transaction/filtered/UtxoFilteredTransaction.java
```diff
@@ -124,4 +126,12 @@ public interface UtxoFilteredTransaction extends TransactionWithMetadata {
...
      */
     void verify();
+
+    /**
+     * Retrieve the Merkle proof for each component group.
+     *
+     * @return A map containing component group IDs as keys and Merkle proofs as values.
+     */
+    @NotNull
+    Map<Integer, MerkleProof> getMerkleProofs();
```
dickon: How about having this take a set of ints as argument and only returning `MerkleProof` for the specificed component groups?

data/avro-schema/src/main/resources/avro/net/corda/data/ledger/persistence/PersistMerkleProofIfDoesNotExist.avsc
dickon, line 1: Would `PersistMerkleProof` be better? The handling of conflicts could be an implementation detail of the server.

nkovacsx: I only followed the pattern how other requests are named:
- PersistSignedGroupParametersIfDoNotExist
- PersistTransactionIfDoesNotExist
etc.

ledger/ledger-utxo/src/main/java/net/corda/v5/ledger/utxo/transaction/filtered/UtxoFilteredTransaction.java
nkovacsx: Yeah that might be a better approach.

nkovacsx: Moving this to the internal API instead, I don't think this needs to be public after giving it some thoughts.

dickon (approved):
LGTM

blsemo (approved):
Ok as a stepping stone  - just need to make sure we remove the intermediate APIs again.

PR #1450, opened 2024-01-22
CORE-18631: Make link out message nullable

yift-r3:
The session state link out message can be null. This pull request should allow it in the Avro scheme.

I think we don't need to update the version; it will make a record nullable, so everything that works now should work with this change. The actual usage of this class is protected with a feature flag and will be protected for a while...


nikinagy (approved):
LGTM

dimosr (approved):

PR #1455, opened 2024-01-23
CORE-19368 Managed key rotation

simon-johnson-r3:
Changes (breaking) to 5.2 introduced key rotation messages to accommodate managed key rotation. Changes are
- Made the unmanaged key rotation fields optional (nullable)
- Removed "old generation" because you always and only rotate from the latest generation
- Added keyUuid which is the input for managed key rotation

runtime-os changes to accommodate these changes here: https://github.com/corda/corda-runtime-os/pull/5489

### DIFF
--- a/data/avro-schema/src/main/resources/avro/net/corda/data/crypto/wire/ops/key/rotation/IndividualKeyRotationRequest.avsc
+++ b/data/avro-schema/src/main/resources/avro/net/corda/data/crypto/wire/ops/key/rotation/IndividualKeyRotationRequest.avsc
@@ -17,22 +17,22 @@
     {
       "name": "oldParentKeyAlias",
       "type": ["null", "string"],
-      "doc": "For unmanaged key rotation. The key alias whose protected content will be re-wrapped with a new key."
+      "doc": "Mandatory for unmanaged key rotation only, always null for managed key rotation. The key alias whose protected content will be re-wrapped with a new key."
     },
     {
       "name": "newParentKeyAlias",
       "type": ["null", "string"],
-      "doc": "For unmanaged key rotation. The new wrapping key which oldParentKeyAlias' content will be re-wrapped with."
+      "doc": "Mandatory for unmanaged key rotation only, always null for managed key rotation. The new wrapping key which oldParentKeyAlias' content will be re-wrapped with."
     },
     {
       "name": "targetKeyAlias",
       "type": ["null", "string"],
-      "doc": "For unmanaged key rotation. Specifies the wrapped key to rotate."
+      "doc": "Mandatory for unmanaged key rotation only, always null for managed key rotation. Specifies the wrapped key to rotate."
     },
         {
       "name": "keyUuid",
       "type": ["null", "string"],
-      "doc": "For managed key rotation. Specifies the managed wrapping key id to rotate."
+      "doc": "Mandatory for managed key rotation only, always null for unmanaged key rotation. Specifies the managed wrapping key id to rotate."
     },
     {
       "name": "keyType",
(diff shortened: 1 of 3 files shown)


data/avro-schema/src/main/resources/avro/net/corda/data/crypto/wire/ops/key/rotation/IndividualKeyRotationRequest.avsc
```diff
@@ -16,18 +16,23 @@
...
-      "doc": "The key alias whose protected content will be re-wrapped with a new key."
+      "type": ["null", "string"],
+      "doc": "For unmanaged key rotation. The key alias whose protected content will be re-wrapped with a new key."
     },
     {
       "name": "newParentKeyAlias",
-      "type": "string",
-      "doc": "The new wrapping key which oldParentKeyAlias' content will be re-wrapped with."
+      "type": ["null", "string"],
+      "doc": "For unmanaged key rotation. The new wrapping key which oldParentKeyAlias' content will be re-wrapped with."
```
dickon: What hapepns if newParentKeyAlias is null? I guess we use the crypto service default for unmanaged keys, and make up a new wrapping key for managed but let's say so.

data/avro-schema/src/main/resources/avro/net/corda/data/crypto/wire/ops/key/rotation/IndividualKeyRotationRequest.avsc
```diff
@@ -16,18 +16,23 @@
     },
     {
       "name": "oldParentKeyAlias",
-      "type": "string",
-      "doc": "The key alias whose protected content will be re-wrapped with a new key."
+      "type": ["null", "string"],
+      "doc": "For unmanaged key rotation. The key alias whose protected content will be re-wrapped with a new key."
```
dickon: null means rewrap all matching wrapping keys?

data/avro-schema/src/main/resources/avro/net/corda/data/crypto/wire/ops/key/rotation/IndividualKeyRotationRequest.avsc
simon-johnson-r3: If it's null for unmanaged keys that's an error and it wont rotate. For managed keys you never need a new alias, because you are always rotating to a key with the same alias and a higher generation.

Note that there is no extra explicit validation yet, that will come when we implement the managed key handling. For now it will throw and fail, the error is likely not the best, but the code hasn't actually changed here as Java strings could always be null, so there's no extra path in the code which could result in a new error so to speak.

data/avro-schema/src/main/resources/avro/net/corda/data/crypto/wire/ops/key/rotation/IndividualKeyRotationRequest.avsc
simon-johnson-r3: Yes, and for managed key rotation that is always the case. For unmanaged, not setting this would generate an error.

data/avro-schema/src/main/resources/avro/net/corda/data/crypto/wire/ops/key/rotation/IndividualKeyRotationRequest.avsc
```diff
@@ -16,18 +16,23 @@
...
       "name": "targetKeyAlias",
-      "type": "string",
-      "doc": "Specifies the wrapped key to rotate."
+      "type": ["null", "string"],
+      "doc": "For unmanaged key rotation. Specifies the wrapped key to rotate."
+    },
+        {
+      "name": "keyUuid",
+      "type": ["null", "string"],
+      "doc": "For managed key rotation. Specifies the managed wrapping key id to rotate."
```
dickon: I wish this wasn't in the `corda-api` repo but there's a number of such Avro messages already. I think we just have to be clear that the Avro messages are much more likely to change than say the Flow API or the REST API.

dickon (dismissed):
Approve, noting omissions in the doc fields that could be added later if that's going to slow us down now. And our structural problem with needing all our Avro messages go alongside committed public API, which is not something to fix in this PR.

anien (dismissed):
Maybe we could discuss splitting unmanaged and managed avro schemas as they are now completely separate. But that's not for this PR.

dickon (approved):
LGTM

PR #1454, opened 2024-01-22
CORE-18698 Add new `is_filtered` column to store filtered transactions, remove Merkle proof public API

nkovacsx:
- Add a new `is_filtered` column to the `utxo_transaction` table to flag filtered transactions
- Remove the previously added Merkle Proof public API
- Draft PR for the public API that I used for testing: https://github.com/corda/corda-api/pull/1456 (won't be merged for now)

PR #1456, opened 2024-01-23
CORE-18698 Store and fetch filtered transaction API

nkovacsx:
This PR is only a draft as it contains a public API that we don't need for now. 
The real mergable PR is at: https://github.com/corda/corda-api/pull/1454


data/avro-schema/src/main/resources/avro/net/corda/data/ledger/persistence/PersistFilteredTransactionsIfDoNotExist.avsc
```diff
@@ -0,0 +1,13 @@
+{
+  "type": "record",
+  "name": "PersistFilteredTransactionsIfDoNotExist",
```
dickon, line 3: simply PersistFilteredTransaction is a better name.

PR #1457, opened 2024-01-23
CORE-19399 Update the membership registration state topic to be compacted

charlieR3:
The membership registration state topic for the state and event message pattern usage is the only state topic used in this pattern which is not compacted. This PR makes this topic compacted.
The impact of this not being compacted was that when a new consumer came online and all states were read into memory, every state ever written is read rather than just the latest state. This resulted in rebalances taking longer than necessary.


dimosr (approved):
LGTM (pending some questions around upgrade tooling that I've asked offline)

charlieR3:
> LGTM (pending some questions around upgrade tooling that I've asked offline)

Added comment to the original ticket addressing these questions.

PR #1462, opened 2024-01-24
CORE-19020 extend Checkpoint schema to allow outputs from FE to be stored.

LWogan:
Store outputs from the flow engine in the Checkpoint for replaying


data/avro-schema/src/main/resources/avro/net/corda/data/messaging/mediator/MediatorReplayOutputEventType.avsc
```diff
@@ -0,0 +1,22 @@
...
+      "name": "value",
+      "type": "bytes",
+      "doc": "Serialized bytes of the object."
+    },
+    {
+      "name": "type",
+      "type": {
+        "type": "enum",
+        "name": "SerializationType",
+        "symbols": ["STRING", "BYTEARRAY", "AVRO"]
```
mbrkic-r3: Not directly related to this PR, but what was the reason to handle `String` and `ByteArray` differently than other types when serializing to Avro? (We wouldn't need this class if all serialized objects contain type info)

data/avro-schema/src/main/resources/avro/net/corda/data/flow/state/checkpoint/SavedOutput.avsc
```diff
@@ -0,0 +1,28 @@
+{
+  "type": "record",
+  "name": "SavedOutput",
+  "namespace": "net.corda.data.flow.state.checkpoint",
+  "doc": "Record the details of output events to be replayed if necessary.",
+  "fields": [
+    {
+      "name": "topic",
+      "type": "string",
+      "doc": "The topic the event should be sent to."
```
LWogan, line 10: This is needed for external messaging which sets the topic in the flow engine, which is then read from the MessageRouter.

For other payload types we'd rely on the MessageRouter to direct the record.

data/avro-schema/src/main/resources/avro/net/corda/data/flow/state/checkpoint/SavedOutputs.avsc
```diff
@@ -0,0 +1,21 @@
+{
+  "type": "record",
+  "name": "SavedOutputs",
+  "namespace": "net.corda.data.flow.state.checkpoint",
+  "doc": "Record the output events for a given input event, to be replayed by the flow engine if necessary.",
+  "fields": [
+    {
+      "name": "inputEventHash",
+      "type": "string",
```
mbrkic-r3, line 9: Is there a reason not to use bytes?

LWogan: It will be passed to the flow engine as a string from the mediator

mbrkic-r3: Maybe it's better not to base it on current implementation of  mediator that works with String only because it's stored with Record that allows only String headers. It might be better to represent it with bytes (it will also take less space).

Also, we shouldn't base hashes on UUIDs as it's a bit confusing and we might get in trouble if its implementation changes in the future. Instead we should use a proper hashing algorithm e.g. MD5.

LWogan: For the purposes of this pr, i think it will have to remain a string. as changing the mediator message processor to accept things other than Record<K, V> is not in scope.

I will refactor the runtime to remove UUIDs

mbrkic-r3 (dismissed):
LGTM

thiagoviana (approved):

PR #1464, opened 2024-01-25
CORE-19403 - Ensure one of the strings `ext` or `EXT` are part of the Kafka topic used by the external messaging service

filipesoliveira:
This change change will prevent people from accidently sending messages to topics that should be used exclusively by the workers.


vkolomeyko (approved):
LGTM

PR #1466, opened 2024-01-26
CORE-18960 Managed key status

simon-johnson-r3:
Added managed key status avro type. This is the type which is stored in state manager for key status and rotation status. It is the equivalent of the existing unmanaged key status but for managed keys.


malachyb (approved):
LGTM

Sakpal (approved):

PR #1471, opened 2024-01-26
CORE-18342 API changes to add order by clause to custom queries

blsemo:
Adding API to vault named queries to add an order by clause (and unconsumed states only(


ledger/ledger-utxo/src/main/java/net/corda/v5/ledger/utxo/query/registration/VaultNamedQueryBuilder.java
```diff
@@ -26,6 +26,49 @@ public interface VaultNamedQueryBuilder extends VaultNamedQueryBuilderBase {
     @NotNull
     VaultNamedQueryBuilder whereJson(@NotNull String query);
 
+
+    /**
+     * Sets an ORDER BY expression. The columnExpression needs to evaluate to a value in the vault database, i.e.
+     * either a column name or a JSON expression. The flags string allows to set order by flags, e.g. "ASC",
+     * these will be passed through to the database as is, so be careful with using database vendor specific
+     * expressions.
```
rick-r3: > expressions 

should be "flags" I think.

rick-r3 (dismissed):

PR #1477, opened 2024-01-31
CORE-19490 Adding topics, task names and config values for schedule FlowStatus cleanup processor

ben-millar:
This PR is part of the **CORE-19435** epic, and supports us in adding a scheduled job to clean-up stale flow statuses from the StateManager which backs the `getStatus()` REST endpoint.

Added:
1. The Kafka topic `scheduled.task.flow.status.processor` to which we will post scheduled FlowStatusCleanup jobs.
2. The Kafka topic `rest.flow.status.cleanup` where batches of stale FlowKeys to be cleaned up will be posted, as determined during the FlowStatusCleanup job.
3. An avro object, `FlowStatusRecord`, which contains the 'fingerprint' (key, version) of a FlowStatus in state storage.
4. An avro object, `ExecuteFlowStatusCleanup`, which will be posted to the `rest.flow.status.cleanup` topic and will contain the **FlowStatusRecords** to be deleted.
5. A new REST config, `flowStatusCleanupTime`, which determines how long a FlowStatus in a terminated state can remain idle (no updates) before it's deleted. Default: 7 days.
6. The task name for the FlowStatusCleanup job, `flow-status-cleanup`.
7. Permission config for the newly created Kafka topics.

### DIFF
--- a/data/avro-schema/src/main/resources/avro/net/corda/data/rest/ExecuteFlowStatusCleanup.avsc
+++ b/data/avro-schema/src/main/resources/avro/net/corda/data/rest/ExecuteFlowStatusCleanup.avsc
@@ -1,0 +1,16 @@
+{
+  "type": "record",
+  "name": "ExecuteFlowStatusCleanup",
+  "docs": "When this event is processed the listed FlowKeys should be deleted from the flow status lookup service store",
+  "namespace": "net.corda.data.rest",
+  "fields": [
+    {
+      "name": "keys",
+      "docs": "A list of FlowKeys for flow states that should be deleted",
+      "type": {
+        "type": "array",
+        "items": "string"
+      }
+    }
+  ]
+}
--- a/data/avro-schema/src/main/resources/avro/net/corda/data/rest/ExecuteFlowStatusCleanup.avsc
+++ b/data/avro-schema/src/main/resources/avro/net/corda/data/rest/ExecuteFlowStatusCleanup.avsc
@@ -1,15 +1,33 @@
 {
   "type": "record",
-  "name": "ExecuteFlowStatusCleanup",
-  "docs": "When this event is processed the listed FlowKeys should be deleted from the flow status lookup service store",
+  "name": "FlowStatusRecord",
+  "docs": "This record represents the fingerprint of a specific FlowStatus in state storage, holding its key and its version",
   "namespace": "net.corda.data.rest",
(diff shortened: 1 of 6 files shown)


data/avro-schema/src/main/resources/avro/net/corda/data/rest/ExecuteFlowStatusCleanup.avsc
```diff
@@ -0,0 +1,16 @@
+{
+  "type": "record",
+  "name": "ExecuteFlowStatusCleanup",
+  "docs": "When this event is processed the listed FlowKeys should be deleted from the flow status lookup service store",
+  "namespace": "net.corda.data.rest",
+  "fields": [
```
mbrkic-r3, line 24: Do we need state version here as well?

data/config-schema/src/main/resources/net/corda/schema/configuration/rest/1.0/corda.rest.json
```diff
@@ -62,6 +62,13 @@
           "default": 30000
         }
       }
+    },
+    "flowStatusCleanupTime": {
```
josephzunigadaly: I feel like milliseconds is too fine a unit to use here.

Assuming we keep milliseconds, should we add 'Ms' to the end of the name, like `idleTimeoutMs` above?

ben-millar: I agree that milliseconds probably allows a lot more granularity than we need, but I think I'll keep it now for a couple of reasons:
1. All other config values in this section use milliseconds, so it's predictable;
2. We suggest in the description that it must be set higher than another config, which is defined in ms, so I don't want to make users do the maths.

We can get this config up to about 24 days using milliseconds; if there comes a point that we need a longer retention period, I think we should change it.

data/avro-schema/src/main/resources/avro/net/corda/data/rest/ExecuteFlowStatusCleanup.avsc
ben-millar: Yeah, good point; I'll add this too.

josephzunigadaly (approved):

mbrkic-r3 (approved):
LGTM

PR #1479, opened 2024-01-31
CORE-19519 persist transaction signatures request

vlajos:
Introduce a dedicated PersistTransactionSignatures call to avoid persisting the whole transaction again when only signatures have been added.

Runtime: https://github.com/corda/corda-runtime-os/pull/5579


gradle.properties
```diff
@@ -5,7 +5,7 @@ cordaProductVersion = 5.2.0
 # NOTE: update this each time this module contains a breaking change
 ## NOTE: currently this is a top level revision, so all API versions will line up, but this could be moved to
 ##   a per module property in which case module versions can change independently.
-cordaApiRevision = 33
+cordaApiRevision = 34
```
lankydan: Does this need an API change? I think it should be backwards compatible.

lankydan (approved):
Left a comment but looks ok.

PR #1481, opened 2024-02-01
CORE-18929 make objects passed into external event executor @CordaSerializable

LWogan:
see https://github.com/corda/corda-runtime-os/pull/5562

All objects passed into the external event request executor must now be @CordaSerializable


gradle.properties
```diff
@@ -5,7 +5,7 @@ cordaProductVersion = 5.2.0
 # NOTE: update this each time this module contains a breaking change
 ## NOTE: currently this is a top level revision, so all API versions will line up, but this could be moved to
 ##   a per module property in which case module versions can change independently.
-cordaApiRevision = 34
+cordaApiRevision = 999
```
mbrkic-r3: Is this mistake?

LWogan: its easier when merging down from release when in dev

mbrkic-r3 (dismissed):
LGTM

mbrkic-r3 (dismissed):
re-approving

thiagoviana (approved):

PR #1489, opened 2024-02-02
CORE-17933 Fetch filtered transactions and signatures avro

jennyang-r3:
Added avro schemas to persistence to
- find filtered transactions and signatures
- persist filtered transactions and signatures (for the following ticket to add loading/storing job)


ledger/ledger-utxo/src/main/java/net/corda/v5/ledger/utxo/UtxoLedgerService.java
```diff
@@ -87,6 +88,16 @@ public interface UtxoLedgerService {
...
 
+    /**
+     * Finds matching {@link UtxoFilteredTransaction} and the notary signatures for a given signed transaction in the vault.
+     *
+     * @param signedTransaction A new transaction of dependencies to find.
+     * @return Returns the map of past transactions, containing only notary signatures in the objects.
+     */
+    @NotNull
+    @Suspendable
+    Map<SecureHash, Map<UtxoFilteredTransaction, List<DigitalSignatureAndMetadata>>> findFilteredTransactionsAndSignatures(@NotNull UtxoSignedTransaction signedTransaction);
```
lankydan, line 99: Does this need to be a public api? Or do we only want it for our own usage?

I say this because I worry that the name is a bit confusing (although the doc makes it clear what it does), `findFilteredTransactionsAndSignaturesForTransaction` would be cleared but also aware that it becomes extremely long.

I also think we could remove the `AndSignatures` from the name to be `findFilteredTransactions` / `findFilteredTransactionsForTransaction` / `findFilteredTransactionsFor`. I think the `AndSignatures` conflates the api and puts too much empahsis of the signatures part.

Don;t feel super strong about the naming but at least worth considering if it hasn't been thought through already.

jennyang-r3: > Does this need to be a public api? Or do we only want it for our own usage?

This is used to replace code that gathers `FilteredTransactionAndSignatures` in finality flow and contract verifying notary client. I made it public to be able to use this in notary as well. Not sure if I should have left notary code as it is and the API in finality flow only.

rick-r3 (approved):

PR #1492, opened 2024-02-06
CORE-19586 improve config descriptions for flow cleanup

LWogan:
Description was copied from the session cleanup config, which is not accurate for this field


ben-millar (approved):
LGTM

PR #1493, opened 2024-02-06
CORE-19018 async checkpoint cleanup

LWogan:
https://github.com/corda/corda-runtime-os/pull/5648

Currently checkpoints are deleted in the flow engine when a flow completes successfully, when an error occurs, or if it is killed.

The flow idempotence and determinism work saves all outputs from the flow engine for a flow to the checkpoint. This will allow for a replay mechanism to be built in the future. It will also allow for the worker to replay outputs when the consumer is reset or a producer sends duplicate messages to the bus that are consumed by the flow engine.

In order to ensure the last outputs are sent to the bus, we must ensure the checkpoint lives for a period longer than it currently does. At the moment the flow engine could delete the checkpoint and write this to the state manager, but then fail to send the last outputs to the message bus, perhaps due to kafka or worker errors.

This could result in flows that have completed successfully, being marked as failed, or stuck in running. It may also leave sessions that hang. Delegating deletion to a process after a configurable period of time solves this.


mbrkic-r3 (approved):
LGTM

PR #1495, opened 2024-02-07
CORE-19401: Add a processor timeout config for the mediator

JamesHR3:
### Problem description

The event mediator may need to time out event processing. Exactly how long the timeout should be is heavily dependent on the workflows being executed, but waiting too long in general will result in the consumer group being timed out.

### Solution

Make this configurable, so the user has a chance to tune this according to the needs of the application.

Note that a new config item has been added rather than reusing the state and event one. This is because the default value needs to be much higher, and bumping the old default value may have unintended side effects elsewhere in the system.


data/config-schema/src/main/resources/net/corda/schema/configuration/messaging/1.0/subscription.json
```diff
@@ -72,6 +72,13 @@
           "minimum": 10,
           "maximum": 2147483647,
           "default": 50
+        },
+        "processorTimeout": {
+          "description": "The length of time the mediator will wait for a single group of events to process. This value should be set to less than the consumer max.poll.interval.ms config value, which by default is 5 minutes.",
```
jujoramos: Can we add the unit to the description?.

jujoramos (dismissed):
LGTM, just one minor comment that will make our lives easier.

jujoramos (approved):

PR #1506, opened 2024-02-12
CORE-18730 Increase session timeout default value

owenstanford:
This change increases the P2P Link Manager session timeout, this is to avoid unnecessary session recycling for slow responders.


williamvigorr3 (approved):
LGTM

PR #1513, opened 2024-02-16
CORE-18471 Remove Java API checks for crypto topic names

bpaunescu:
After careful consideration, I ended up removing the SchemaTets.java file. 

Reasons: 

- The source of truth for topic names (and config) is in Crypto.yaml; should never be in a test

- The test simply asserted that some Java constants (how topic names were accessed programmatically) have non-empty, non-null Strings

- The test did not cover all existing crypto topics

- There are no other such tests for any other topic names in Corda

- There is a better test (SchemaTests.kt) which verifies that whatever topic names exist in the .yaml files also exist in the Schemas.java as public, static fields.


anien (approved):
LGTM.

bpaunescu:
Raising PR against 5.3

PR #1516, opened 2024-02-19
CORE-19714 remove scheduled job to delete checkpoints asynchronously

LWogan:
Remove jobs no longer being used along with associated topics and configs

https://github.com/corda/corda-runtime-os/pull/5720


mbrkic-r3:
Did we have any config values for this that should be removed as well?

LWogan:
> Did we have any config values for this that should be removed as well?

Yes. Updated to remove

mbrkic-r3 (approved):
LGTM

driessamyn (approved):

PR #1522, opened 2024-02-20
CORE-18539: Increase "maxIdleTime" Flow Property

jujoramos:
Increase the default maximum time a flow can be in 'RUNNING' state
without receiving any updates before Corda marks the flow as 'FAILED'
and deletes associated state from 5 to 30 minutes (same default
timeout set for flow sessions).


driessamyn (approved):
Ok.

PR #1524, opened 2024-02-21
CORE-19323: Clean up things that are not needed due to new link manager sessions manager

yift-r3:
Clean up parts that the link manager no longer needs after moving to the stateful session manager. 

See runtime pull request in https://github.com/corda/corda-runtime-os/pull/5758


nikinagy (commented):
Out of curiosity: are these schemas that were unused in 5.2? And now we clean them up? Or is it possible we might have unprocessed messages in 5.3 after upgrade from these types?
It would be nice to run an upgrade test pipeline just in case (which is unfortunately broken atm).

owenstanford (approved):

PR #1538, opened 2024-02-29
CORE-19839 Add APIs for PublicKey encoding and decoding

bpaunescu:
Runtime OS PR is [here](https://github.com/corda/corda-runtime-os/pull/5798)

### DIFF
--- a/application/src/main/java/net/corda/v5/application/crypto/SigningService.java
+++ b/application/src/main/java/net/corda/v5/application/crypto/SigningService.java
@@ -57,4 +57,44 @@ public interface SigningService {
     @Suspendable
     @NotNull
     Map<PublicKey, PublicKey> findMySigningKeys(@NotNull Set<PublicKey> keys);
+    
+    /**
+     * Deserializes a {@link PublicKey} from a {@code byte} array.
+     *
+     * @param encodedKey The public key represented asa a {@code byte} array.
+     * @return An instance of {@link PublicKey} constructed from the encoded key.
+     */
+    @Suspendable
+    @NotNull
+    PublicKey decodePublicKey(@NotNull byte[] encodedKey);
+
+    /**
+     * Deserializes a {@link PublicKey} from a {@link String}.
+     *
+     * @param encodedKey The public key represented asa a hex encoded {@link String}.
+     * @return An instance of {@link PublicKey} constructed from the encoded key.
+     */
+    @Suspendable
+    @NotNull
+    PublicKey decodePublicKey(@NotNull String encodedKey);
+
+    /**
+     * Serializes a {@link PublicKey} into a byte array.
+     *
+     * @param publicKey The {@link PublicKey} to be encoded.
+     * @return A {@code byte} array representation of the public key.
+     */
+    @Suspendable
+    @NotNull
+    byte[] encodeAsByteArray(@NotNull PublicKey publicKey);
+
+    /**
+     * Serializes a {@link PublicKey} into a hex encoded String.
+     *
+     * @param publicKey The {@link PublicKey} to be encoded.
+     * @return A hex encoded {@link String} representation of the public key.
+     */
+    @Suspendable
+    @NotNull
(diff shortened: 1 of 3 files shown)


application/src/main/java/net/corda/v5/application/crypto/SigningService.java
```diff
@@ -57,4 +57,44 @@ public interface SigningService {
     @Suspendable
     @NotNull
     Map<PublicKey, PublicKey> findMySigningKeys(@NotNull Set<PublicKey> keys);
+    
+    /**
+     * Deserializes a {@link PublicKey} from a {@code byte} array.
+     *
+     * @param encodedKey The public key represented asa a {@code byte} array.
```
simon-johnson-r3: typo -> asa

application/src/main/java/net/corda/v5/application/crypto/SigningService.java
```diff
@@ -57,4 +57,44 @@ public interface SigningService {
...
+     * @return An instance of {@link PublicKey} constructed from the encoded key.
+     */
+    @Suspendable
+    @NotNull
+    PublicKey decodePublicKey(@NotNull byte[] encodedKey);
+
+    /**
+     * Deserializes a {@link PublicKey} from a {@link String}.
+     *
+     * @param encodedKey The public key represented asa a hex encoded {@link String}.
```
simon-johnson-r3: we refer to this as PEM encoded in the KeyEncodingService which I think is a bit more explanatory?

application/src/main/java/net/corda/v5/application/crypto/SigningService.java
```diff
@@ -57,4 +57,44 @@ public interface SigningService {
...
+     *
+     * @param publicKey The {@link PublicKey} to be encoded.
+     * @return A {@code byte} array representation of the public key.
+     */
+    @Suspendable
+    @NotNull
+    byte[] encodeAsByteArray(@NotNull PublicKey publicKey);
+
+    /**
+     * Serializes a {@link PublicKey} into a hex encoded String.
```
simon-johnson-r3: same here

simon-johnson-r3 (changes requested):

application/src/main/java/net/corda/v5/application/crypto/SigningService.java
```diff
@@ -57,4 +57,44 @@ public interface SigningService {
...
     @NotNull
     Map<PublicKey, PublicKey> findMySigningKeys(@NotNull Set<PublicKey> keys);
+    
+    /**
+     * Deserializes a {@link PublicKey} from a {@code byte} array.
+     *
+     * @param encodedKey The public key represented asa a {@code byte} array.
+     * @return An instance of {@link PublicKey} constructed from the encoded key.
+     */
+    @Suspendable
```
simon-johnson-r3: In two minds whether these should be Suspendable, they will never actually suspend, but marking them as such opens up to future changes - however I see other functions in this module are not all marked as suspendable so maybe we should remove that?

application/src/main/java/net/corda/v5/application/crypto/SigningService.java
bpaunescu: Fixed.

application/src/main/java/net/corda/v5/application/crypto/SigningService.java
bpaunescu: Fixed.

application/src/main/java/net/corda/v5/application/crypto/SigningService.java
bpaunescu: Fixed.

application/src/main/java/net/corda/v5/application/crypto/SigningService.java
bpaunescu: Removed suspendable annotation.

simon-johnson-r3 (approved):

PR #1543, opened 2024-03-05
CORE-19996 Ledger repair on a schedule

lankydan:
## Problem

A transaction can be notarized successfully, leading to the appropriate records being stored in the notary's database, but transaction fails to be stored as `VERIFIED` in the finalizing members' databases.

This inconsistency breaks the ledger from the perspective of the members that failed to store the transaction, as any states consumed by the transaction will still show as unconsumed even though, from the perspective of the notary and the members the did store it correctly, the states are consumed. This incorrect view causes attempted double spends which the notary then shuts down, meaning there are no further issues caused by failing to store the transaction. However, depending on the CorDapp logic of the member that failed to store the transaction, they might not be able to progress any further until the inconsistency is resolved.

## Solution

To solve this problem, the `corda-runtime-os` change has introduced:

- Added "read" logic to the notary and uniqueness code that checks if a transaction has already been notarized.
- A flow that runs on a schedule calls the notarization "read" and stores a transaction as `VERIFIED` if it has already been notarized.

The changes in this PR support the runtime changes.


blsemo (dismissed):
LGTM

rick-r3 (dismissed):

ledger/notary-plugin/src/main/java/net/corda/v5/ledger/notary/plugin/core/NotaryExceptionUnknown.java
```diff
@@ -6,8 +6,8 @@
 import org.jetbrains.annotations.Nullable;
 
 /**
- * The base class for notary exceptions that we cannot guarantee were fatal and might return with a different result
- * when retried. These type of exceptions will not invalidate the transaction immediately.
+ * The base class for notary exceptions that were fatal and might return with a different result when retried. These
```
blsemo: "that were fatal" - isn't that wrong as in they are not necessarily fatal to the tx? I find this change confusing.

blsemo (dismissed):
LGTM

rick-r3 (approved):

PR #1548, opened 2024-03-07
CORE-18986 - Add topic and task name for scheduled stale sessions clean up

nikinagy:
We need to implement the mechanism that will periodically check (e.g. every 7 days) whether there are sessions that have expired, but not cleaned up yet. This will use the task scheduler to publish the task events and a durable subscription to process them.

As part of this PR, I added the required topic and task name for the scheduled task's subscription.


williamvigorr3 (approved):
LGTM

PR #1559, opened 2024-03-13
CORE-9933 Key Alias constraint moved to higher version of LB file

bpaunescu:
Looks like the correct way to update db schema files is to create a new file and update it instead of changing the existing one because it breaks upgrade tests.


data/db-schema/src/main/resources/net/corda/db/schema/crypto/migration/crypto-creation-v5.3.xml
```diff
@@ -0,0 +1,18 @@
...
+                   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-4.3.xsd">
+    <!--
+     For now, the schema must be identical between here and the 
+     virtual node schema (in ../../vnode-crypto/migration/crypto-creation-v1.0.xml)
+     since the crypto processor uses the same code irrespective of where the table resides.
+     
+     We could just include the same changeset files in both cases, but that might confuse readers
+     and liquibase, so instead we have to keep the content in sync.  
+     -->
```
kyriathar, line 12: I am not sure we need this comment as it refers to internal things (crypto processor).

data/db-schema/src/main/resources/net/corda/db/schema/vnode-crypto/migration/vnode-crypto-creation-v5.3.xml
```diff
@@ -0,0 +1,18 @@
...
+                   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-4.3.xsd">
+    <!--
+    For now, the crypto_signing_key table schema must be identical between here and the 
+    cluster schema (in ../../crypto/migration/crypto-creation-v1.0.xml)
+    since the crypto processor uses the same code irrespective of where the table resides.
+    
+    We could just include the same changeset files in both cases, but that might confuse readers
+    and liquibase, so instead we have to keep the content in sync.  
+    -->
```
kyriathar, line 12: Same as above

kyriathar (approved):
LGTM

PR #1563, opened 2024-03-18
CORE-20131: Make Strategy CordaSerializable

josephzunigadaly:


filipesoliveira (approved):
The test `TokenSelectionTests` is green.

ledger/ledger-utxo/src/main/java/net/corda/v5/ledger/utxo/token/selection/TokenClaimCriteria.java
```diff
@@ -94,6 +95,7 @@ public TokenClaimCriteria(
      * @param targetAmount   The minimum value for the sum of {@link ClaimedToken#getAmount()} for the selected tokens.
      * @param strategy       The token selection strategy to use.
      */
+    @ConstructorForDeserialization
     public TokenClaimCriteria(
```
vkolomeyko, line 99: Should we add a test to assert that `TokenClaimCriteria` can be serialized?
I am happy if we do it as a separate PR after this one is merged.

vkolomeyko (approved):

ledger/ledger-utxo/src/main/java/net/corda/v5/ledger/utxo/token/selection/TokenClaimCriteria.java
josephzunigadaly: I'll raise a different PR for that one.

PR #1564, opened 2024-03-18
CORE-20044 Updated the docs in the persistence service

simon-johnson-r3:
The persistence service will do automatic retries as long as some criteria are met in the definitions and use of the entities being persisted. This change makes that clear to CorDapp developers in the docs.

There are no functional changes to the api.


rick-r3 (approved):

ben-millar (approved):
Very explicit, nice work

PR #1566, opened 2024-03-19
CORE-19290 - Set the `tokenCacheExpiryPeriodMilliseconds` to be an integer instead of a long

filipesoliveira:
The type `long` prevents the `combined worker` from starting up. This change changes the type of `tokenCacheExpiryPeriodMilliseconds ` from long to integer.


vkolomeyko (approved):
LGTM

PR #1581, opened 2024-03-26
CORE-20210 Flow worker must be added as a producer to flow mapper start due to ledger repair tool changes

LWogan:
Without this change the ledger repair tool will not work on a kafka deployment


ronanbrowne (approved):
LGTM

PR #1585, opened 2024-03-27
CORE-19290 - Added the field `strategy` to `TokenClaimQuery`

filipesoliveira:


vkolomeyko (dismissed):
LGTM

data/avro-schema/src/main/resources/avro/net/corda/data/ledger/utxo/token/selection/data/StrategyAvro.avsc
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "enum",
+  "name": "StrategyAvro",
```
josephzunigadaly: We don't normally add "Avro" to the name here. We name the Avro class like we normally would, then import it with a different name for the source files that need both Avro and Corda types.

Here is one example:
```kotlin
import net.corda.ledger.utxo.verification.TransactionVerificationStatus as TransactionVerificationStatusAvro
```

Not saying that's the best approach, but we should aim for consistency. Otherwise we will have multiple approaches in our code and things will get confusing.

gradle.properties
```diff
@@ -5,7 +5,7 @@ cordaProductVersion = 5.3.0
 # NOTE: update this each time this module contains a breaking change
 ## NOTE: currently this is a top level revision, so all API versions will line up, but this could be moved to
 ##   a per module property in which case module versions can change independently.
-cordaApiRevision = 8
+cordaApiRevision = 9
```
josephzunigadaly: We only need to update this if we have a breaking change. I'd expect this change to be compatible with the existing runtime-os code.

vkolomeyko (approved):
LGTM

josephzunigadaly (approved):

PR #1598, opened 2024-04-08
CORE-19289 - Updated the table `utxo_visible_transaction_output`

filipesoliveira:
Added the column `token_priority` to the database table `utxo_visible_transaction_output`.
Created an index that uses the `token_priority` and `transaction_id` columns.

Note: 
1) Liquibase does not support the creation of an index where NULL values are last unless we write the SQL code ourselves. (See reference: https://docs.liquibase.com/change-types/create-index.html)
2) Postgres reference for creating an index with the `NULLS LAST` option: https://www.postgresql.org/docs/current/indexes-ordering.html

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -5,7 +5,7 @@ cordaProductVersion = 5.3.0
 # NOTE: update this each time this module contains a breaking change
 ## NOTE: currently this is a top level revision, so all API versions will line up, but this could be moved to
 ##   a per module property in which case module versions can change independently.
-cordaApiRevision = 9
+cordaApiRevision = 10
 
 # Main
 kotlin.stdlib.default.dependency = false
--- a/gradle.properties
+++ b/gradle.properties
@@ -5,7 +5,7 @@ cordaProductVersion = 5.3.0
 # NOTE: update this each time this module contains a breaking change
 ## NOTE: currently this is a top level revision, so all API versions will line up, but this could be moved to
 ##   a per module property in which case module versions can change independently.
-cordaApiRevision = 10
+cordaApiRevision = 9
 
 # Main
 kotlin.stdlib.default.dependency = false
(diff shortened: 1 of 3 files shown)


gradle.properties
```diff
@@ -5,7 +5,7 @@ cordaProductVersion = 5.3.0
 # NOTE: update this each time this module contains a breaking change
 ## NOTE: currently this is a top level revision, so all API versions will line up, but this could be moved to
 ##   a per module property in which case module versions can change independently.
-cordaApiRevision = 9
```
josephzunigadaly: We should not need to change the API revision for this change. It is not a breaking change.

filipesoliveira: Thanks @josephzunigadaly. I was playing it safe to avoid disrupting other people's work.

josephzunigadaly: Changing this disrupts people that do need to change `cordaApiRevision`. Please don't change this unless it is a breaking change.

vkolomeyko: I think I would agree with @josephzunigadaly on this instance - let's not bump-up `cordaApiRevision` when we do not need to.

It will also make our life easier when integrating this change, as we will not have to apply for `corda-api` lock.

josephzunigadaly (approved):

vkolomeyko (approved):
LGTM

filipesoliveira:
Added the column `token_priority` to the database table `utxo_visible_transaction_output`.
Created an index that uses the `token_priority` and `transaction_id` columns.

Note: 
1) Liquibase does not support the creation of an index where NULL values are last unless we write the SQL code ourselves. (See reference: https://docs.liquibase.com/change-types/create-index.html)
2) Postgres reference for creating an index with the `NULLS LAST` option: https://www.postgresql.org/docs/current/indexes-ordering.html

PR #1602, opened 2024-04-10
CORE-19221 Add MessageAck to AppMessage schema

YashNabar:
Adds `MessageAck` to `AppMessage` schema. Extends `AuthenticatedMessageAck` to include the key from the associated `AuthenticatedMessageAndKey` to enable routing of the ack to the correct link manager.


data/avro-schema/src/main/resources/avro/net/corda/data/p2p/app/AppMessage.avsc
```diff
@@ -9,7 +9,8 @@
         "type": [
           "net.corda.data.p2p.app.AuthenticatedMessage",
           "net.corda.data.p2p.app.OutboundUnauthenticatedMessage",
-          "net.corda.data.p2p.app.InboundUnauthenticatedMessage"
+          "net.corda.data.p2p.app.InboundUnauthenticatedMessage",
```
yift-r3, line 12: Let's change the API version

PR #1607, opened 2024-04-16
CORE-20093: Add default master wrapping key alias to individual key rotation request message

anien:
To improve the sync issues between crypto workers when updating the configuration before key rotation, we use only one place to retrieve the default master wrapping key alias. To pass it to other workers, we use kafka message.

Runtime-os PR: [#6029](https://github.com/corda/corda-runtime-os/pull/6029)


bpaunescu (approved):
LGTM

PR #1614, opened 2024-04-18
CORE-20347: Add link ack in

yift-r3:
See details in the runtime pull request https://github.com/corda/corda-runtime-os/pull/6061

### DIFF
--- a/data/config-schema/src/main/resources/net/corda/schema/configuration/p2p.linkManager/1.0/corda.p2p.linkManager.json
+++ b/data/config-schema/src/main/resources/net/corda/schema/configuration/p2p.linkManager/1.0/corda.p2p.linkManager.json
@@ -171,12 +171,6 @@
       "type": "object",
       "default": {},
       "properties": {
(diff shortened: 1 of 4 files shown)


data/config-schema/src/main/resources/net/corda/schema/configuration/p2p.linkManager/1.0/corda.p2p.linkManager.json
```diff
@@ -171,12 +171,6 @@
       "type": "object",
       "default": {},
       "properties": {
```
simon-johnson-r3, line 173: You can't remove config properties, see https://r3-cev.atlassian.net/wiki/spaces/CB/pages/4564091182/Making+a+Config+Change#Not-Permitted-Changes

The reason being 5.3 Corda will read 5.2 config on startup after an upgrade, it will see this property from the existing config and flag it's an invalid property not in its schema and fail.

simon-johnson-r3 (changes requested):
Config property removed.

PR #1620, opened 2024-04-24
CORE-6717 Add schemas for hosted identity reconciliation

YashNabar:
Introduces database and Avro schemas to enable persistence and reconciliation of locally-hosted identities.

- Adds cluster-level tables `hosted_identity` and `hosted_identity_session_key_info`
- Extends `HostedIdentityEntry.avsc` to include `version` (compatibility test added)
- Adds config for hosted identity reconciliation

### DIFF
--- a/data/avro-schema/src/main/resources/avro/net/corda/data/p2p/HostedIdentityEntry.avsc
+++ b/data/avro-schema/src/main/resources/avro/net/corda/data/p2p/HostedIdentityEntry.avsc
@@ -33,6 +33,10 @@
          "type": "array",
          "items": "HostedIdentitySessionKeyAndCert"
        }
-     }
+     },
+    {
+      "name": "version",
+      "type": "int"
+    }
   ]
 }
--- a/data/avro-schema/src/main/resources/avro/net/corda/data/p2p/HostedIdentityEntry.avsc
+++ b/data/avro-schema/src/main/resources/avro/net/corda/data/p2p/HostedIdentityEntry.avsc
@@ -35,6 +35,7 @@
        }
      },
     {
+      "doc": "Version of the hosted identity",
       "name": "version",
       "type": ["null", "int"],
       "default": null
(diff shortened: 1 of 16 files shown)


data/avro-schema/src/main/resources/avro/net/corda/data/p2p/HostedIdentityEntry.avsc
```diff
@@ -33,6 +33,11 @@
          "type": "array",
          "items": "HostedIdentitySessionKeyAndCert"
        }
-     }
+     },
+    {
+      "name": "version",
+      "type": ["null", "int"],
```
nikinagy, line 40: add docs?

nikinagy (dismissed):
LGTM

owenstanford (dismissed):

driessamyn (dismissed):
This cannot go in 5.2 as patch releases must be able to be applied in place (without an upgrade process for DBs etc).

nikinagy:
> This cannot go in 5.2 as patch releases must be able to be applied in place (without an upgrade process for DBs etc).

@driessamyn The main problem is that we were not persisting data in the DB, so in case of kafka data loss we lost everything.
The hosted identities topic was never backed up by any sort of DB entity. Hence we had no DB reconciliation in place.
I don't see how we could fix the SBI issue for clients without making any changes to the DB. Especially, since the main issue is that we are missing persistence for this data.

data/config-schema/src/main/resources/net/corda/schema/configuration/reconciliation/1.0/corda.reconciliation.json
```diff
@@ -61,6 +61,13 @@
       "minimum": 5000,
       "maximum": 2147483647,
       "default": 120000
+    },
```
simon-johnson-r3, line 64: Just a heads up, after upgrade it's possible the config read by a worker will not yet have this property with a default propagated to it. The code must be able to handle this property is missing in 5.2.1. See https://r3-cev.atlassian.net/wiki/spaces/CB/pages/4564091182/Making+a+Config+Change for more details.

simon-johnson-r3 (approved):

conal-smith (approved):
LGTM

PR #1631, opened 2024-05-01
CORE-19840 Make `pub_key_hash` part of PK for signature table

lankydan:
Remove `signature_idx` column and use `pub_key_hash` in the primary key instead, which simplifies the insert and `ON CONFLICT` logic, while ensuring that we don't have issues when storing a filtered transaction followed by a signed transaction for the same transaction id.


blsemo (approved):
LGTM

PR #1636, opened 2024-05-07
ES-1946: Update CODEOWNERS for 5.2

ronanbrowne:
- Further updates add Conal at the request of Vini 
- Ensure David is also present at API layer


vkolomeyko (approved):
LGTM

PR #1647, opened 2024-05-20
CORE-19908: Fix crypto schema to include default values for retrying

malachyb:
The schema for the retrying section of the crypto config was malformed and didn't set defaults correctly. This PR fixes the schema and allows it to set defaults better. Additionally, the defaults in the caching section seemed like they had been swapped around, as the default caching timeout was 10,000 minutes and the number of cached items was only 60. This PR also swaps those values around.


anien (approved):
LGTM, I can see now config with defaults containing all the sections.

PR #1659, opened 2024-06-12
CORE-20522 Improve reconciler

williamvigorr3:


data/avro-schema/src/main/resources/avro/net/corda/data/crypto/wire/ops/reconciliation/request/LookupKeyById.avsc
```diff
@@ -0,0 +1,20 @@
+{
+  "type": "record",
+  "name": "LookUpKeyById",
```
kyriathar: I think we already have this type under name `ByIdsRpcQuery`? Do we need to introduce this new type, or maybe the old one covers this case as well?

data/avro-schema/src/main/resources/avro/net/corda/data/crypto/wire/ops/reconciliation/response/LookupKeyByIdResponse.avsc
```diff
@@ -0,0 +1,16 @@
+{
+  "type": "record",
+  "name": "LookupKeyByIdResponse",
+  "namespace": "net.corda.data.crypto.wire.ops.reconciliation.response",
```
kyriathar: I think this looks fine. 

Only thing, not really sure if we should be binding it with reconciliation (i.e. namespace). It could be consumed by other things in the future.

Perhaps some `sync` in the namespace would make sense since the form of it implies sync since it carries the error as well? (I think when it comes to types over Kafka, i.e. async we handle errors under the hood and then the user doesn't need to handle it themselves?)

williamvigorr3: I changed `reconciliation` to `sync`. In the future we can add other operations.

data/avro-schema/src/main/resources/avro/net/corda/data/crypto/wire/ops/reconciliation/request/LookupKeyById.avsc
williamvigorr3: This includes the tenant. I can change it if you want to be more generic so we can add more operations latter.

PR #1675, opened 2024-07-18
CORE-20790: Add version to ChangeGroupParentIdRequest

Tom-Fitzpatrick:
Used to prevent concurrency issues when updating the parent id of a group.


vkolomeyko (approved):
LGTM

PR #1677, opened 2024-07-24
CORE-20543: Update ChangeUserParentGroupIdRequest to use loginName

BenYip123:
- Updated the ChangeUserParentGroupIdRequest to be loginName instead of userId, as all user operations take loginName as input.


data/avro-schema/src/main/resources/avro/net/corda/data/permissions/management/user/ChangeUserParentGroupIdRequest.avsc
```diff
@@ -4,7 +4,7 @@
   "namespace": "net.corda.data.permissions.management.user",
   "fields": [
     {
-      "name": "userId",
+      "name": "loginName",
```
vkolomeyko, line 7: In theory, this is a breaking change - so we might be better off incrementing version of `corda-api` in this PR.

vkolomeyko (approved):
LGTM

Tom-Fitzpatrick (approved):

PR #1696, opened 2024-09-30
R3SOL-400  Explicitly add the JSON column type to v5.1 utxo creation

nkovacsx:
### Overiew

Right now when running the `ledger-utxo-creation-v5.1` changeset the property `json.column.type` comes from another changeset XML when they are imported together.

However, this makes it impossible to execute `ledger-utxo-creation-v5.1` as a standalone changeset because this property will become undefined. Adding this property to the changeset.


driessamyn (dismissed):
LGTM

driessamyn (approved):

PR #1707, opened 2024-10-22
CORE-20867 add a flow retry topic to handle retries of transient RPC calls

LWogan:
Add config for retrying external event transient errors

Add new FlowEvent payload type to capture the details required to generates an external event retry request for transient errors.

https://github.com/corda/corda-runtime-os/pull/6374

### DIFF
--- a/data/topic-schema/src/main/resources/net/corda/schema/Flow.yaml
+++ b/data/topic-schema/src/main/resources/net/corda/schema/Flow.yaml
@@ -89,4 +89,12 @@ topics:
     producers:
       - flow
     config:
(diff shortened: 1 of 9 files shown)


data/topic-schema/src/main/resources/net/corda/schema/Flow.yaml
```diff
@@ -89,4 +89,12 @@ topics:
     producers:
       - flow
     config:
+  FlowRetry:
```
LWogan: i dont think i need this. flow.event was never removed. i can use that

PR #1710, opened 2024-11-11
CORE-20867 add a flow retry topic to handle retries of transient RPC calls 

LWogan:
Add new FlowEvent payload type to capture the details required to generates an external event retry request for transient errors.


data/config-schema/src/main/java/net/corda/schema/configuration/MessagingConfig.java
```diff
@@ -25,6 +25,7 @@ private Bus() {
         public static final String KAFKA_BOOTSTRAP_SERVERS = KAFKA_PROPERTIES_COMMON + ".bootstrap.servers";
         public static final String KAFKA_PROPERTIES_CONSUMER = KAFKA_PROPERTIES + ".consumer";
         public static final String KAFKA_CONSUMER_MAX_POLL_INTERVAL = KAFKA_PROPERTIES_CONSUMER + ".max.poll.interval.ms";
+        public static final String KAFKA_CONSUMER_MAX_POLL_RECORDS = KAFKA_PROPERTIES_CONSUMER + ".max.poll.records";
```
sprintengine, line 28: is there an accompanying json schema change that sets a default for this, so configs held in memory on workers on 5.2.1 upgrading to 5.2.2 have a value?

sprintengine: DIscussed with Lorcan and it should be safe we don't set anything into the messaging config sitting on workers, there is no issue parsing the JSON into an object for the GET config API either.

conal-smith (approved):
LGTM

PR #1715, opened 2024-11-18
CORE-20817: waiver for snyk issues and upgrade to latest version of dokka

LWogan:
Corda5 Shippable artifacts do not make use of dokka-core, which is where this dependency originates, this is used at compile / build timeonly for Kdoc generation and not shipped in any of our releasable artifacts.


ronanbrowne (approved):
LGTM

